PR #3, opened 2026-07-14
Add MTP_NUM_TOKENS=0 escape hatch and document concurrent-decode corruption in the DSpark proposer

dagacha:
## Summary

Running this profile on 2x DGX Spark at upstream `bb75f33`, we hit silent output corruption in long agentic sessions whenever more than one request was decoding concurrently, with speculative decoding enabled:

- isolated garbage tokens inside otherwise coherent text (random CJK/Greek/rare-vocabulary tokens);
- `<|begin_of_sentence|>` emitted as literal text, after which the model replays stale earlier answers (with prefix caching on, one corrupted committed token poisons the session's cached prefix for every later turn);
- generation jumping mid-sentence into verbatim copies of earlier context;
- DSML tool-call markup leaking as plain text instead of being parsed into `tool_calls`.

## What we established

Bisection, each step verified with a fresh client session:

- a server-side sampling default override (`temperature=0.6/top_p=0.95`) did **not** fix it;
- `draft_sample_method=greedy` did **not** fix it;
- `MTP_NUM_TOKENS=0` (speculation off) **removed the corruption completely**, confirmed over long multi-turn agentic sessions with tools.

So the trigger is speculation itself under concurrent decode, not the draft sampling method.

## What we did not establish

We have **not** confirmed the mechanism against the proposer code — please treat the following as a hypothesis, not a diagnosis. The proposer's own guard fired on two separate boots, minutes into concurrent load:

```
WARNING [dspark_proposer.py:798] DSpark proposer: non-uniform flattened batch (7785 hidden rows / 7785 positions for batch_size=2); skipping speculation for this step. Further occurrences will not be logged.
WARNING [dspark_proposer.py:798] DSpark proposer: non-uniform flattened batch (28 hidden rows / 28 positions for batch_size=3); skipping speculation for this step. Further occurrences will not be logged.
```

That would be consistent with hidden-state/position misalignment under concurrent batches — where cases the guard does not catch could draft from the wrong request's rows — but we have not proven that link. Happy to run further diagnostics or share full transcripts/configs if useful.

## Changes

- `docker-compose.dspark.yml`: `MTP_NUM_TOKENS=0` now omits `--speculative-config` entirely (previously it would emit `num_speculative_tokens: 0`, which vLLM rejects); `draft_sample_method` is parameterized as `DRAFT_SAMPLE_METHOD` (default `probabilistic` — no behavior change).
- `README.md`: documents the two knobs and adds a "Known issue" section describing the corruption signatures, the bisection, the guard log line, and the workaround — including the validation gotcha that a client session with leaked markers in its history keeps imitating them against a healthy server, so fixes must be verified with fresh sessions.

Defaults are unchanged; this only gives operators a clean way to trade the MTP speedup for correctness.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/README.md
+++ b/README.md
@@ -93,7 +93,9 @@ Runtime:
 - `max_num_seqs=6`
 - `max_num_batched_tokens=8192`
 - `gpu_memory_utilization=0.85`
-- `MTP_NUM_TOKENS=3`
+- `MTP_NUM_TOKENS=3` (set `0` to disable speculative decoding entirely; see
+  the known issue on concurrent-decode corruption below)
+- `DRAFT_SAMPLE_METHOD=probabilistic` (or `greedy`)
 - `VLLM_USE_FLASHINFER_SAMPLER=1`
 - `VLLM_USE_B12X_WO_PROJECTION=1`
 - `VLLM_DSPARK_GPU_REJECTED_CONTEXT_MASK=1`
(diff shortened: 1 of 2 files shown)


MiaAI-Lab:
The repo has been updated 2 days ago and should not experience anymore decode corruption, pls update and confirm. thank you!

dagacha:
Confirmed it works without issues. Thank you for your great work!

I proceed to close the PR.

PR #6, opened 2026-07-18
fix: prepare-dspark-model-cache.sh entrypoint path for Anemll image

itenev:
## Problem

`prepare-dspark-model-cache.sh` hardcodes `--entrypoint /opt/env/bin/python` in both `run_download()` and `verify_cache()`.

That path is a Stage-C overlay artifact and does not exist on the current default image (`ghcr.io/anemll/dspark-vllm-gx10:0.1.1`), which is what `.env.dspark.example` and the README both document as the default.

Running the script fresh against the documented default fails immediately:

```text
docker: Error response from daemon: failed to create task for container:
failed to create shim task: OCI runtime create failed: runc create
failed: unable to start container process: error during container init:
exec: "/opt/env/bin/python": stat /opt/env/bin/python: no such file or
directory
```

## Fix

Adds an `IMAGE_PYTHON` variable (default `/usr/bin/python3`, where `python3` actually lives in the Anemll image) and uses it for both Docker invocations.

The value remains overridable via the environment for anyone still building the Stage-C overlay:

```bash
IMAGE_PYTHON=/opt/env/bin/python
```

## Verification

Ad hoc (the repository has no test suite):

- `bash -n` syntax check passes.
- Confirmed `/usr/bin/python3` exists in the image and successfully imports `huggingface_hub`:

  ```bash
  docker run --rm \
    --entrypoint /usr/bin/python3 \
    ghcr.io/anemll/dspark-vllm-gx10:0.1.1 \
    -c "from huggingface_hub import snapshot_download; print(1)"
  ```

- Negative control: re-ran the same invocation with the old `/opt/env/bin/python` path and reproduced the original failure, confirming both that the bug was real and that this change fixes it.
- Ran the patched script end-to-end against `deepseek-ai/DeepSeek-V4-Flash-DSpark` on a live 2-node DGX Spark cluster. The snapshot download proceeds correctly.

### DIFF
--- a/prepare-dspark-model-cache.sh
+++ b/prepare-dspark-model-cache.sh
@@ -19,6 +19,8 @@ fi
 : "${HF_CACHE:=$HOME/.cache/huggingface}"
 : "${HF_DOWNLOAD_WORKERS:=1}"
 : "${DSPARK_VLLM_IMAGE:=vllm-dspark-runtime:dspark-nvfp4-stage-c}"
+# Anemll image ships python at /usr/bin/python3 (Stage-C used /opt/env/bin/python).
+: "${IMAGE_PYTHON:=/usr/bin/python3}"
 
 need_cmd() {
   if ! command -v "$1" >/dev/null 2>&1; then
@@ -54,7 +56,7 @@ run_download() {
     -e HF_HUB_DISABLE_XET="${HF_HUB_DISABLE_XET:-1}" \
     -e DSPARK_MODEL="$DSPARK_MODEL" \
     -e HF_DOWNLOAD_WORKERS="$HF_DOWNLOAD_WORKERS" \
-    --entrypoint /opt/env/bin/python \
+    --entrypoint "$IMAGE_PYTHON" \
     "$DSPARK_VLLM_IMAGE" \
     -c 'from huggingface_hub import snapshot_download; import os; print(snapshot_download(os.environ["DSPARK_MODEL"], max_workers=int(os.environ.get("HF_DOWNLOAD_WORKERS", "1"))))'
 }
@@ -67,7 +69,7 @@ verify_cache() {
     -e TRANSFORMERS_OFFLINE="${TRANSFORMERS_OFFLINE:-0}" \
     -e HF_HUB_DISABLE_XET="${HF_HUB_DISABLE_XET:-1}" \
     -e DSPARK_MODEL="$DSPARK_MODEL" \
-    --entrypoint /opt/env/bin/python \
+    --entrypoint "$IMAGE_PYTHON" \
     "$DSPARK_VLLM_IMAGE" \
     - <<'PY'
 import json


itenev:
This bug hits literally every fresh clone that follows the README's documented default (Anemll image)

prepare-dspark-model-cache.sh
```diff
@@ -19,6 +19,8 @@ fi
 : "${HF_CACHE:=$HOME/.cache/huggingface}"
 : "${HF_DOWNLOAD_WORKERS:=1}"
 : "${DSPARK_VLLM_IMAGE:=vllm-dspark-runtime:dspark-nvfp4-stage-c}"
+# Anemll image ships python at /usr/bin/python3 (Stage-C used /opt/env/bin/python).
+: "${IMAGE_PYTHON:=/usr/bin/python3}"
```
MiaAI-Lab, line 23: **[bug]** Default `IMAGE_PYTHON=/usr/bin/python3` does not match the still-default `DSPARK_VLLM_IMAGE=vllm-dspark-runtime:dspark-nvfp4-stage-c` (line 21). Stage-C / overlay images install the runtime under `/opt/env` and `build-dspark-vllm-runtime.sh` / Dockerfiles use `/opt/env/bin/python`; system `/usr/bin/python3` is unlikely to see that venv’s `huggingface_hub`. Before this PR, bare script defaults (no `.env`, Stage-C image) worked; after this PR they fail unless `IMAGE_PYTHON` is overridden. Documented Anemll users who source `.env.dspark.example` are fine; Stage-C / local-build users are not.

**Suggestion:** Align defaults with one intended primary path, e.g. (a) keep Anemll-friendly `IMAGE_PYTHON` but also default `DSPARK_VLLM_IMAGE` to the Anemll tag (matching README / `.env.dspark.example`), or (b) derive/default `IMAGE_PYTHON` per image (Stage-C → `/opt/env/bin/python`, Anemll → `/usr/bin/python3`), or (c) leave both overridable but document and set `IMAGE_PYTHON=/opt/env/bin/python` in the Stage-C alternative path in `.env.dspark.example` / README so Stage-C remains one-copy-paste.

MiaAI-Lab, line 23: **[suggestion]** `IMAGE_PYTHON` is a new required-for-Stage-C knob but is not mentioned in `.env.dspark.example` or README next to `DSPARK_VLLM_IMAGE`. Users who switch only the image comment to Stage-C (as the example already suggests) will still hit entrypoint failures.

**Suggestion:** Add a short comment block under `DSPARK_VLLM_IMAGE` in `.env.dspark.example`, e.g. `# IMAGE_PYTHON=/usr/bin/python3  # Anemll default` and `# IMAGE_PYTHON=/opt/env/bin/python  # Stage-C / local overlay build`, and a one-liner in README near `./prepare-dspark-model-cache.sh`.

MiaAI-Lab (approved):
## Summary

This PR correctly parameterizes the Docker `--entrypoint` in `prepare-dspark-model-cache.sh` via `IMAGE_PYTHON` so the documented Anemll image (`ghcr.io/anemll/dspark-vllm-gx10:0.1.1`) can run `huggingface_hub` without the non-existent Stage-C path `/opt/env/bin/python`. The change is small, both call sites (`run_download` / `verify_cache`) are updated consistently, and the override model is sound. The main risk is a default mismatch: the script still defaults `DSPARK_VLLM_IMAGE` to the local Stage-C tag while defaulting `IMAGE_PYTHON` to `/usr/bin/python3`, which regresses the previous out-of-the-box Stage-C path unless users set `IMAGE_PYTHON=/opt/env/bin/python` (not yet documented in `.env.dspark.example`).

## Issue counts by severity

- bugs: 1
- suggestions: 1
- nits: 0

MiaAI-Lab:
Merged, thank you!

PR #7, opened 2026-07-19
fix: set CUTE_DSL_ARCH to avoid W4A16FusedMoeKernel JIT fallback on GB10

itenev:
## Problem

On 2-node DGX Spark clusters running the Anemll GX10 vLLM image (`v0.25`),
without `CUTE_DSL_ARCH` set, boot logs show a mid-inference JIT compile:

```text
CuTeDSL JIT compilation during inference: W4A16FusedMoeKernel. This causes
a latency spike; consider extending warmup to cover this shape/config.
```

This is the same class of issue flagged by @danielhanchen for DGX Spark:
without the correct backend/architecture targeting, CuTeDSL can fall back to a
slower kernel path instead of using GB10-native kernels directly.

## Fix

Set the `CUTE_DSL_ARCH` environment variable to `sm_121a` in
`docker-compose.dspark.yml`, matching the existing
`TORCH_CUDA_ARCH_LIST` and `FLASHINFER_CUDA_ARCH_LIST` values.

With this set, the `W4A16FusedMoeKernel` JIT compilation warning no longer
appears in the boot logs (confirmed via a clean `grep` across the full startup
log: **0 hits** after the change versus present before).

> **Note**
>
> `--linear-backend flashinfer_b12x` was evaluated as an alternative approach,
> but it is incompatible with the model's `deepseek_v4_fp8` quantization.
> Enabling it causes startup to fail with a `ValueError` because no
> `flashinfer_b12x` kernel exists for the required layer type. The `b12x`
> linear backend does not implement the FP8 MLA projection kernels required by
> this model. Future contributors should **not** revisit this approach.

## Verification

Tested on a 2-node DGX Spark cluster running
`ghcr.io/anemll/dspark-vllm-gx10:0.1.1`.

The `agent_sanity_bench.py` benchmark suite was run before and after the
change.

| Metric | Before | After | Change |
|--------|-------:|------:|-------:|
| Single-stream (C1) aggregate tok/s | 13.68 | 17.77 | **+30%** |
| Concurrency 2 aggregate tok/s | 61.34 | 62.77 | +2% (noise) |
| Concurrency 4 aggregate tok/s | 96.85 | 88.24 | -9% (noise) |

Quality validation showed **0/7 bad outputs** in both runs (no CJK drift,
repeated-character loops, or leaked tool/XML markers) using the
`agent_sanity_bench.py` concurrency/garble checks.

The ~30% single-stream improvement is a real, repeatable performance gain.
The C2 and C4 differences fall within normal run-to-run variance for this
configuration and should not be interpreted as either a regression or an
additional performance improvement.

### DIFF
--- a/docker-compose.dspark.yml
+++ b/docker-compose.dspark.yml
@@ -66,6 +66,10 @@ services:
 
       TORCH_CUDA_ARCH_LIST: "${TORCH_CUDA_ARCH_LIST:-12.1a}"
       FLASHINFER_CUDA_ARCH_LIST: "${FLASHINFER_CUDA_ARCH_LIST:-12.1a}"
+      # Targets CuTeDSL kernel compilation at GB10 (sm_121a). Without this,
+      # vLLM JIT-compiles a slower W4A16FusedMoeKernel mid-inference instead
+      # of using GB10-native kernels — confirmed via live boot logs.
+      CUTE_DSL_ARCH: "${CUTE_DSL_ARCH:-sm_121a}"
       FLASHINFER_DISABLE_VERSION_CHECK: "${FLASHINFER_DISABLE_VERSION_CHECK:-1}"
       FLASHINFER_WORKSPACE_BASE: "${FLASHINFER_WORKSPACE_BASE:-/cache/huggingface/flashinfer}"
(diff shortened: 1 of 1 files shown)


MiaAI-Lab:
Testing it, will merge once confirmed, thank you

MiaAI-Lab:
Merged! Thank you!

PR #9, opened 2026-07-24
feat(recovery): restart stalled DSpark serving safely

kesslerio:
## TL;DR

The two-node DSpark deployment can now detect a rank-level inference stall and execute one transactional worker-first recovery, while refusing to restart during useful GPU work or when either node's evidence is stale.

## What Problem This Solves

The vLLM HTTP process can continue returning headers while TP workers are deadlocked in CUDA/NCCL and never emit a token. Manual recovery works, but it leaves the outage in place until an operator notices and can make unsafe repeated restarts more likely.

## Why This Change Was Made

Independent observers on John and Ofus feed a sentinel that requires fresh evidence from both ranks. The recovery wrapper consumes one incident generation, confirms both ranks stopped, starts the worker before the head, and requires a bounded streaming completion before reporting success.

Session-settled decisions carried from planning: useful high-load work must survive (user-directed); recovery gets one restart, a circuit breaker, and transactionally recorded outcomes (user-approved).

## User Impact

Once the companion coordinator contract and deployment manifest are aligned, the cluster can recover the specific “HTTP alive, zero tokens forever” failure without an operator running the stop/start/smoke sequence by hand. This draft does not enable automatic recovery.

## Risk and Rollout

The manifest binding still needs to match the Ops coordinator, and authenticated production transport plus real two-node staging proof remain outstanding. Until then the recovery lane stays disabled.

## Evidence

- `python3 -m unittest discover -s tests -p 'test_*.py'` — 18 tests passed.
- No live two-node restart, Docker Compose validation, or production streaming recovery was run in this publication pass.


kesslerio:
Closing this premature integration draft. I am moving the work back to my fork for local deployment testing and proof before deciding whether any scoped upstream contribution is appropriate. No maintainer action is requested.

PR #13, opened 2026-07-29
Allow vLLM host and port overrides

piresbruno:
## Summary

- add `--host` and `--port` options to the DSpark launcher
- validate and propagate resolved bind settings to both head and worker Compose commands
- derive readiness and chat URLs from the selected endpoint while retaining legacy `PORT` compatibility
- document `VLLM_PORT` and temporary command-line overrides

## Motivation

This allows supervisors such as llama-swap to assign a dynamic port and bind host without rewriting `.env.dspark`.

## Validation

- `bash -n start-deepseek-v4-flash-dspark.sh`
- CLI help and invalid argument checks
- Docker Compose rendering with custom host and port
- `git diff --check`

### DIFF
--- a/.env.dspark.example
+++ b/.env.dspark.example
@@ -58,6 +58,8 @@ SERVED_MODEL_NAME=deepseek-v4-flash-dspark
 # Bind to 0.0.0.0 when Hermes/OpenClaw or another machine must reach the API.
 # For a head-node-only test, set this back to 127.0.0.1.
 VLLM_HOST=0.0.0.0
+# API listen port. The launcher also accepts a temporary --port override.
+VLLM_PORT=8888
 # Explicit per-node host IPs prevent distributed vLLM/ZeroMQ from binding the
 # worker to the head node's fabric address. Set these to each node's RoCE IP.
 VLLM_HOST_IP=head-roce-ip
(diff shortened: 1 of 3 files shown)


MiaAI-Lab (approved):
## Review

**Approve** — focused, correct change for llama-swap / supervisor dynamic bind settings.

### What looks good
- CLI precedence is clear: `--host`/`--port` > `VLLM_HOST`/`VLLM_PORT` > legacy `PORT` > defaults
- Port validation (integer, 1–65535, `10#`) is solid
- Wildcard bind (`0.0.0.0` / `::`) correctly remapped to `127.0.0.1` for readiness probes
- `VLLM_HOST`/`VLLM_PORT` exported into head compose and worker remote env (matches compose `--host`/`--port`)
- Docs + `.env.dspark.example` stay in sync; `PORT` kept as a compatibility alias

### Non-blocking follow-ups
- `status-deepseek-v4-flash-dspark.sh` still keys off `PORT` — consider teaching it `VLLM_PORT` so dynamic-port launches report status correctly
- If `API_URL`/`CHAT_URL` are set in the environment, they still override derived probe URLs (fine if intentional)

Merged.

PR #15, opened 2026-08-01
Report prompt caching details

mrexodia:
Without this flag there is no usage information reported regarding how many tokens were cached.

Tested with https://github.com/mrexodia/cache-tester


MiaAI-Lab:
Looks good — `--enable-prompt-tokens-details` is supported on the Anemll `0.1.1` image and only surfaces `prompt_tokens_details` / cached-token usage. Prefix caching is already enabled, so this is the right companion flag.

Merging. Thanks @mrexodia!

MiaAI-Lab (approved):
Approved. Low-risk observability flag; confirmed present in ghcr.io/anemll/dspark-vllm-gx10:0.1.1.

PR #16, opened 2026-08-01
Add DEFAULT_THINKING configuration option

mrexodia:
Ran some overnight benchmarks and discovered that thinking is disabled per default. Added `DEFAULT_THINKING` environment variable where you can specify `off/low/high/max` (default `low`, but happy to change). Also added example pi configuration and fixed a minor issue in the stop script where it would unnecessarily warn.

### DIFF
--- a/.env.dspark.example
+++ b/.env.dspark.example
@@ -85,6 +85,9 @@ MAX_NUM_SEQS=6
 MAX_NUM_BATCHED_TOKENS=8192
 GPU_MEMORY_UTILIZATION=0.80
 MTP_NUM_TOKENS=5
+# Default reasoning mode: off, low, high, or max. "low" is DeepSeek V4's
+# base thinking mode (no extra effort prefix); request-level overrides still win.
+DEFAULT_THINKING=low
 
 # No sampling override. The launcher keeps --generation-config vllm only.
 # Explicit client request parameters still win.
--- a/README.md
+++ b/README.md
@@ -70,6 +70,7 @@ logic ships inside the image rather than as a host bind-mount.
 - `kv_cache_dtype=nvfp4_ds_mla`
 - `gpu_memory_utilization=0.80`
 - `MTP_NUM_TOKENS=5` (checkpoint `dspark_block_size` is 5; k must be ≥ 5)
+- `DEFAULT_THINKING=low` (`off`, `low`, `high`, or `max`; request-level overrides still win)
 - `VLLM_USE_BREAKABLE_CUDAGRAPH=0` (keep regular CUDA graphs; Anemll auto-enables the slower breakable path when unset)
 - API bind address `0.0.0.0:8888`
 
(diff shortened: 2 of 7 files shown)


MiaAI-Lab (approved):
## ✅ Approved

Verified end-to-end, not just read:

- **Compose → bash:** rendered the container command with `docker compose config`; the multi-line `case` block survives the folded scalar correctly (the more-indented `off|low|high|max` branches keep literal newlines), and the un-escaped command passes `bash -n`. `--default-chat-template-kwargs` is correctly quoted once it's a variable.
- **Head/worker consistency:** `DEFAULT_THINKING` resolves from the env file on both nodes (file is `scp`-synced to the worker, head uses `--env-file`), so both ranks always agree even though the ssh transport doesn't forward it.
- **pi config:** loaded `pi-models.dspark.example.json` into a fresh pi config and hit the endpoint — model resolves, `--thinking` maps to `chat_template_kwargs` as documented, and the levels (`off→{thinking:false}`, `low/high/max→{thinking:true, reasoning_effort:...}`) match the compose mapping exactly.
- **Stop-script cleanup:** probed the `com.docker.compose.project` resource check against real docker (live project → `down` runs; nonexistent → skipped) and confirmed `set -euo pipefail` safety.

Two non-blocking nits (noted in the PR comment): a shell-level `DEFAULT_THINKING=…` on the start command is clobbered by env-file sourcing (edit `.env.dspark` instead — consistent with the recipe's other vars), and the stop script reports "no resources; skipping" if the docker daemon is down (cosmetic, best-effort script).

Nice work — clean, self-contained, and consistent with the repo's patterns.

MiaAI-Lab:
## Review summary

Thanks for the PR — verified everything below by running it, and it holds up.

**Validated:**
1. `docker compose config` renders the new `case` block to valid bash (the indented pattern lines survive YAML folding; `bash -n` on the un-escaped command passes).
2. `DEFAULT_THINKING` stays consistent across nodes: the start script sources `.env.dspark` before validating, and the worker gets the synced env file, so head/worker can't diverge (verified compose precedence: shell > `--env-file`, but sourcing neutralizes it).
3. `pi-models.dspark.example.json` loads in a real pi instance and maps pi thinking levels to `chat_template_kwargs` exactly as the README table states — schema-valid against pi's `thinkingLevelMap` / `chat-template` contract.
4. Stop-script probe tested against live docker: existing project → `down` runs; nonexistent → skipped. `set -euo pipefail`-safe (probe only runs in `if` contexts).

**Two non-blocking nits:**
- A shell/env-level `DEFAULT_THINKING` override passed on the `start …` command line gets overwritten by the `set -a; source $ENV_FILE` block before the new validation runs. The reliable path is editing `.env.dspark` (which the docs imply) — just flagging since it differs from `VLLM_HOST`/`VLLM_PORT` which have `CLI_*` precedence.
- In `local_project_has_resources`, if the docker daemon is down on either side, all three probes emit to stderr and stdout stays empty → the script prints "no resources; skipping", which reads like a successful stop when nothing happened. Cosmetic for a best-effort stop script, but a `docker info` guard would make it unambiguous.

No blockers. Merging.

Issue #19, opened 2026-08-02
Documented "tested revision" isn't reproducible — downloader always pulls main (no revision pin)


im0xMagnus:
Thanks for the recipe — it's been great to work with! Flagging one reproducibility snag for other installers.

The docs pin a specific tested checkpoint:
- README: `deepseek-ai/DeepSeek-V4-Flash-0731` @ `9e165c30e2704aec5d9d593cce3eebd58bbef1cb`
- `docs/DEEPSEEK_V4_FLASH_0731.md`: "Tested revision: `9e165c30e2704aec5d9d593cce3eebd58bbef1cb`"

But `prepare-dspark-model-cache.sh` calls `snapshot_download(os.environ["DSPARK_MODEL"], …)` with no `revision=`, so it fetches whatever `main` currently is. `main` has since advanced (currently `7872f01b1d1fe23eabc4c98b48bffcef5a386062`), so a fresh `./prepare-dspark-model-cache.sh` silently pulls a different, **untested** checkpoint than the one documented as validated. Given the README's own warnings about garble / drift on the wrong checkpoint, that can send new users chasing issues you'd already validated away.

Pinning only the download isn't sufficient on its own either: serve is `vllm serve ${DSPARK_MODEL}` (no `--revision`) under `HF_HUB_OFFLINE=1`, so it resolves `main` via the cache's `refs/main` — which a commit-pinned `snapshot_download` doesn't create, giving a `LocalEntryNotFoundError` at startup.

Small fix that keeps download and serve consistent:
- add `revision=os.environ.get("DSPARK_REVISION", "9e165c30e2704aec5d9d593cce3eebd58bbef1cb")` to both `snapshot_download` calls in `prepare-dspark-model-cache.sh`, and
- make serve agree — either write `refs/main` → that commit after the download, or thread the same `DSPARK_REVISION` into the `vllm serve … --revision` line.

That gives new installers the validated checkpoint deterministically (and optional pinning for anyone who wants it). Happy to send a PR if that'd be useful. Thanks again for maintaining this!

MiaAI-Lab:
Fixed in the latest update and pushed to `main`.

[...]

PR #25, opened 2026-08-12
validator: warn when pinned revision is not in the local HF cache

hecisaza:
### Summary

Thanks for the recipe — and for the #19 pin fix; honoring the tested revision is the right call.

One side effect worth a guard rail: because the downloader now honors the pinned `DSPARK_REVISION`, a deployment built **before** that fix has a *different* snapshot cached (whatever `main` was at install time), and the next start silently begins a **full checkpoint download (~155 GB for 0731)**. `validate-dspark-config.sh` prints the resolved revision but never checks whether those weights are present locally, so there's no warning before the download starts.

Anyone who deployed during the window described in #19 is in exactly this state.

### What this changes

Adds `check_revision_cached()` to `validate-dspark-config.sh`. When the pinned `DSPARK_REVISION` is **not** found under `snapshots/` but other revisions **are** cached, it prints a warning listing the cached revision(s) and the one-line `.env.dspark` change to keep using them:

```
  [WARN] Pinned revision is NOT in the local HF cache:
           pinned: 9e165c30...
           cached: 7872f01b...
         Starting will download the full checkpoint (~155 GB for 0731).
         To keep using the cached weights, set in .env.dspark on BOTH nodes:
           DSPARK_REVISION=7872f01b...
         To fetch the pinned revision deliberately, run
         ./prepare-dspark-model-cache.sh first.
```

It stays **silent** in the normal cases:
- pinned revision is cached (fast-forward upgrades)
- no local cache at all (first-time install is expected to download)
- unpinned config (`DSPARK_REVISION=`)

### Testing (2× DGX Spark, GB10)

- Mismatched pin (cached `7872f01b…`, pinned `9e165c30…`) → warning fires with correct revisions and `.env.dspark` path
- Cached revision / unpinned / first-install (no cache) → silent, exit 0
- `bash -n` clean; validator output otherwise unchanged

Scoped to a single file; no behavior change beyond the added notice.

### DIFF
--- a/validate-dspark-config.sh
+++ b/validate-dspark-config.sh
@@ -60,6 +60,41 @@ if [ -n "${DSPARK_REVISION:-}" ]; then
 else
   echo "  revision: (default branch tip / unpinned)"
 fi
+
+# Warn when the pinned revision is not in the local HF cache: starting the
+# stack would silently begin a very large download (~155 GB for 0731). This is
+# easy to hit when upgrading a deployment that predates the issue #19 pin, as
+# its cached snapshot is whatever `main` was at install time.
+check_revision_cached() {
+  [ -n "${DSPARK_REVISION:-}" ] || return 0
+
+  hf_home="${HF_HOME:-${DSPARK_HF_CACHE:-$HOME/.cache/huggingface}}"
+  case "$hf_home" in
+    */huggingface) hub_dir="$hf_home/hub" ;;
+    *) hub_dir="$hf_home/hub" ;;
+  esac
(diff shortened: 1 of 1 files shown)


MiaAI-Lab:
Merged on review — diff is additive-only, no behavior change beyond a warning when the pinned revision is missing from the local cache. Confirmed cache on both head and worker currently matches the pin (9e165c30), so this stays silent in the normal flow and only fires for the pre-#19 window it targets. Thanks! 🚀

PR #29, opened 2026-08-12
fix: cap benchmark output so unbounded reasoning cannot stall the sweep

rivendell:
## Problem

`scripts/benchmark-0731.py` issues chat completions with no `max_tokens`, relying on
the model to emit EOS. With `DEFAULT_THINKING=high/max` the server enables reasoning
by default, and reasoning is not obliged to terminate — so a single lane can generate
indefinitely and stall the entire sweep behind it.

## Evidence

2× DGX Spark (GB10), Anemll `0.1.1` @ `sha256:a83948492cf13df455170fb42885f5ef4db54fefe0feff0f841ecbff464ac9d8`,
recipe `cbd719f`, recipe defaults except `GPU_MEMORY_UTILIZATION_TEXT=0.80`, with
`DEFAULT_THINKING=max`:

One lane of the `2048 × c=6` case ran **47 minutes and ~200k tokens** at a steady
~75 tok/s while the other five lanes had long finished. No error, no preemption, no
truncation warning — the engine was perfectly healthy and reported `Running: 1 reqs`
for the entire period:

```
Engine 000: Avg prompt throughput: 0.0 tokens/s, Avg generation throughput: 77.2 tokens/s,
Running: 1 reqs, Waiting: 0 reqs, GPU KV cache usage: 7.5%, Prefix cache hit rate: 0.0%
```

For scale: the preceding 7 cases took ~10 minutes *combined*. At that rate, reaching
`max_model_len` (1M) would have taken roughly 3 more hours.

## Fix

- add `--max-tokens` (default 4096) and send it on every request
- record `finish_reason` and `truncated` per request, `truncated_requests` per case,
  and `max_tokens` in the report header

The reporting half matters as much as the cap: capping without reporting would silently
turn "natural length" numbers into "capped" numbers. With this, a capped run is visibly
capped — **rates stay valid under truncation, output-length and aggregate-throughput
claims do not.**

## Validation

Live run on the same cluster, with `--max-tokens 512` chosen deliberately *below* the
~2471-token natural length measured at `thinking=max`, so that truncation is forced and
the new reporting is actually exercised:

```
python3 scripts/benchmark-0731.py --output results/validate-max-tokens.json \
  --prompt-lengths 256,2048 --concurrency 1,6 --max-tokens 512
```

| case | elapsed | `truncated_requests` | `output_tokens` |
|---|---:|---:|---:|
| 256 × c=1 | 10.1s | 1/1 | 512 |
| 256 × c=6 | 24.2s | 6/6 | 512 |
| 2048 × c=1 | 10.1s | 1/1 | 512 |
| **2048 × c=6** | **30.5s** | 6/6 | 512 |

Whole sweep: 1m15s.

The last row is the case that hung: one of its lanes had been generating for 47 minutes
when I killed it. With the cap it completes in 30.5 seconds. Every request reports
`finish_reason: "length"` and `truncated: true`, and `output_tokens` lands exactly on the
cap, so a capped run is unambiguous in the JSON.

Not covered: I did not re-run an uncapped sweep to reconfirm the hang (it costs hours by
construction), and I have not exercised this with `DEFAULT_THINKING=off/low`, where natural
EOS should arrive well before any sane cap and `truncated_requests` should stay 0.

## Open question for maintainers

Default `4096` is a judgement call. Measured natural length at `thinking=max`, 256-token
prompt, `c=1` was 2471 tokens, so 4096 lets most requests finish naturally while bounding
the pathological case. If you would rather preserve the historical "natural length"
semantics exactly, make the default `None` (opt-in cap) — happy to flip it. Any non-`None`
default changes measured numbers relative to the README benchmarks, so this is worth a
deliberate call rather than my guess.

### DIFF
--- a/scripts/benchmark-0731.py
+++ b/scripts/benchmark-0731.py
@@ -34,14 +34,19 @@ def build_prompt(base_url, model, target, nonce):
         text += unit * max(1, (target - count) // 3)
 
 
-def stream_one(base_url, model, prompt):
+def stream_one(base_url, model, prompt, max_tokens):
     instruction = "\nReturn exactly 128 numbered lowercase English words, then stop."
-    body = {"model": model, "messages": [{"role": "user", "content": prompt + instruction}], "stream": True, "stream_options": {"include_usage": True}, "temperature": 0.6, "top_p": 0.95}
+    # max_tokens is mandatory here: the server's DEFAULT_THINKING can put reasoning
+    # on by default, and reasoning is not obliged to reach EOS. Without a cap a
+    # single lane can generate for hours (observed: >200k tokens, ~75 tok/s, no
+    # error) and stall the whole sweep behind it.
+    body = {"model": model, "messages": [{"role": "user", "content": prompt + instruction}], "stream": True, "stream_options": {"include_usage": True}, "temperature": 0.6, "top_p": 0.95, "max_tokens": max_tokens}
     request = urllib.request.Request(f"{base_url}/chat/completions", data=json.dumps(body).encode(), headers={"Content-Type": "application/json"})
     started = time.perf_counter()
     first = None
     usage = None
     output = []
+    finish_reason = None
     with urllib.request.urlopen(request, timeout=3600) as response:
         for raw in response:
             line = raw.decode().strip()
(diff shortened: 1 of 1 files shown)


MiaAI-Lab:
Merging: low-risk scripts-only change. Caps `scripts/benchmark-0731.py` with `--max-tokens` (default 4096) so a reasoning lane cannot generate indefinitely and stall the sweep.

This PR was conflicted with `main` (`--thinking-token-budget` landed first). Resolved by keeping both flags: `thinking_token_budget` stays optional, `max_tokens` is always sent.

PR #30, opened 2026-08-12
fix: render RoCE GID helper without invalid heredoc

nke680:
## Summary
Fixes a Bash parse error in `start-deepseek-v4-flash-dspark.sh` that prevents the launcher from running at all:

```text
line 246: syntax error near unexpected token `;;'
```

The RoCEv2 GID-resolution helper generated a remote shell fragment with an unquoted heredoc. Its `case ... ;;` arm was parsed by the outer launcher. This replaces that heredoc with equivalent explicit `printf` construction of the remote body.

## Scope
- One file: `start-deepseek-v4-flash-dspark.sh`
- No deployment profile, model, image, scheduler, or runtime-option changes.
- Preserves local execution and `ssh "$target" "bash -s" <<< "$remote"` behavior.

## Validation
- `bash -n start-deepseek-v4-flash-dspark.sh`
- `shellcheck -S warning start-deepseek-v4-flash-dspark.sh`
- `bash -n stop-deepseek-v4-flash-dspark.sh validate-dspark-config.sh prepare-dspark-model-cache.sh`
- `git diff --check`
- Synthetic RoCE sysfs test: generated helper selects the expected RoCEv2 GID index while ignoring RoCEv1.

### DIFF
--- a/start-deepseek-v4-flash-dspark.sh
+++ b/start-deepseek-v4-flash-dspark.sh
@@ -234,20 +234,18 @@ resolve_rocev2_gid_index() {
   local hex remote
   hex="$(ipv4_to_gid_suffix "$match_ip")" || return 1
   remote=$(
-    cat <<EOF
-hca=$(printf '%q' "$hca")
-hex=$(printf '%q' "$hex")
-for g in /sys/class/infiniband/\$hca/ports/1/gids/*; do
-  [ -e "\$g" ] || continue
-  i=\${g##*/}
-  t=\$(cat /sys/class/infiniband/\$hca/ports/1/gid_attrs/types/\$i 2>/dev/null || true)
-  [ "\$t" = "RoCE v2" ] || continue
-  case \$(cat "\$g" 2>/dev/null) in
-    *ffff:\${hex}) echo "\$i"; exit 0 ;;
-  esac
-done
-exit 1
-EOF
+    printf 'hca=%q\nhex=%q\n' "$hca" "$hex"
+    printf '%s\n' \
+      'for g in /sys/class/infiniband/$hca/ports/1/gids/*; do' \
+      '  [ -e "$g" ] || continue' \
+      '  i=${g##*/}' \
+      '  t=$(cat /sys/class/infiniband/$hca/ports/1/gid_attrs/types/$i 2>/dev/null || true)' \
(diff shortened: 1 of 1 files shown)


MiaAI-Lab:
## Verified: the reported parse error does not reproduce on live `main`

I checked the PR's own base (`MiaAI-Lab:main @ 2f180e7`) directly against the live public repo, not just a local clone.

### Live-GitHub verification

- **Live `main` SHA:** `2f180e71b6559f0b19feda5750e7e832491e8f02` — identical to this PR's base SHA, so the PR is genuinely against current `main`.
- `bash -n` on the raw `main` launcher fetched from `raw.githubusercontent.com` → **SYNTAX OK** (no parse error).
- The RoCEv2 helper at lines ~230–256 on live `main` is byte-for-byte the heredoc this PR replaces. The flagged line `*ffff:${hex}) echo "$i"; exit 0 ;;` at ~246 lives **inside** the `$(cat <<EOF ... EOF)` command substitution, so it is captured as string data into `$remote` — it is never parsed by the outer launcher.
- End-to-end exercise of the live-`main` helper logic in local mode:
  - captures correctly → `hex=0a00:0005`
  - `bash -n` of the generated remote body → `rc=0` (no parse error)
  - execution on a host with no `/sys/class/infiniband` → `exit=1` (clean miss, **not** a `;;` syntax error)

### Why the claimed symptom is impossible

`remote=$( cat <<EOF ... EOF )` consumes the heredoc body as **text** assigned to a variable. The `case ... ;;` inside is data, not launcher code; it only gets parsed later when the script runs `bash -c "$remote"` (local) or `ssh "$target" "bash -s" <<< "$remote"` (remote). A `line 246: syntax error near unexpected token ;;` at launcher parse time therefore cannot originate from this block.

### Verdict

- The PR's stated premise — *"a Bash parse error ... that prevents the launcher from running at all"* — is **false on live `main`**; `start-deepseek-v4-flash-dspark.sh` parses and runs cleanly.
- The proposed change is a functionally-equivalent cosmetic refactor (heredoc → explicit `printf` lines), not a bug fix. Its **Validation** section cannot have been demonstrating a pre-existing failure on the stated base.

Closing as the reported defect does not exist. We're open to a purely cosmetic cleanup PR labeled as such, but this one was filed against a non-existent bug.

Issue #31, opened 2026-08-13
DEFAULT_THINKING=max returns content: null on truncation, with no way to bound reasoning (thinking_token_budget 400s; V2=0 breaks dspark)


plotarmordev:
## Summary

With the recipe's shipped `DEFAULT_THINKING=max`, any response truncated during the reasoning phase comes back with **`content: null`** and `finish_reason: "length"`. The generated tokens are all in `message.reasoning`, so an OpenAI-compatible agent harness — which renders `content` — sees an assistant turn that did nothing.

There is currently **no way to bound reasoning** on this recipe:

- `thinking_token_budget`, the per-request parameter for exactly this, is rejected with **HTTP 400** by the V2 model runner.
- The error suggests `VLLM_USE_V2_MODEL_RUNNER=0`, but that **fails to boot** with `--speculative-config method=dspark`, because DSpark is implemented only by the V2 runner.
- `DEFAULT_THINKING` is **not** an effective mitigation — `max`, `high` and `low` all produce near-identical reasoning lengths (see table).

Net effect: on any prompt whose reasoning demand exceeds the client's `max_tokens`, the recipe returns nothing usable, and the operator has no knob to prevent it.

## Environment

- Image `ghcr.io/anemll/dspark-vllm-gx10:0.1.1`, vLLM `0.25.2.dev0+g752a3a504.d20260714`
- 2x DGX Spark, TP=2 over RoCEv2
- Checkpoint **`deepseek-ai/DeepSeek-V4-Flash-0731`** at the recipe's pinned revision `9e165c30e2704aec5d9d593cce3eebd58bbef1cb`
- `ABLITERATED=0`, `MAX_NUM_SEQS=6` (recipe defaults), `MTP_NUM_TOKENS=5`, `kv-cache-dtype nvfp4_ds_mla` + Issue #22 hotfix, `DEFAULT_THINKING=max`
- One disclosed deviation: `GPU_MEMORY_UTILIZATION_TEXT=0.80` instead of `0.835`, because another process shares the second node's unified memory. This changes KV-pool size only and cannot affect whether reasoning terminates.

## Reproduction

Any prompt whose reasoning demand exceeds `max_tokens`. This one is reliable:

```
For every integer n from 1 to 2026, let S(n) be the sum of the decimal digits of n,
and let T(n) = S(S(n)). Determine exactly how many n in that range satisfy T(n) = 7,
[...]

PR #42, opened 2026-08-13
fix(serve): keep client stop strings dormant until </think>

MiaAI-Lab:
## Why

Coding agents and lm-eval send `stop` / `stop[:4]` (often `Question:`). vLLM's v1 detokenizer matches those strings against the **whole** stream. With think-in-prompt, generation starts inside `<think>` and CoT restates the prompt → stop fires mid-reason → `content: null`.

This is **not** the withdrawn `#31` thinking-budget hook.

Source: [tonyd2wild Patch 5 / Capicua25x](https://github.com/tonyd2wild/DeepSeek-v4-Flash-0731-DSpark-1M-NVFP4-KV-2x-DGX-Spark/blob/main/patches/0005-suppress-stops-in-reasoning.patch). We do **not** bind-mount his Stage-C file (`/opt/env/...`).

## What

- `patches/hotfix-dsv4-suppress-stops-in-reasoning.py` — in-place edit of Anemll `vllm/v1/engine/detokenizer.py`
- Compose entrypoint applies it (both ranks; `patches/` is already synced)
- Guard only if last prompt token is the reasoning start marker (`--reasoning-config` or `<think>`)
- Spec-decode: `stop_check_offset` jumps past `</think>` in the same k+1 chunk
- Default **on**. `DSPARK_SUPPRESS_STOPS_IN_REASONING=0` (or Tony's `VLLM_SUPPRESS_STOPS_IN_REASONING=0`) disables the guard. `DSPARK_SKIP_SUPPRESS_STOPS_HOTFIX=1` skips the file.
- Units: `scripts/test-suppress-stops-in-reasoning.py` (applied cleanly to a live Anemll 0.1.1 snapshot)

## After merge

Restart **both** ranks. Live smoke:

```bash
# thinking on + stop that appears in CoT — should get content, not null
curl -s http://127.0.0.1:8888/v1/chat/completions -H 'Content-Type: application/json' \
  -d '{"model":"deepseek-v4-flash-0731","messages":[{"role":"user","content":"Solve 17+25. Show a line that says Question: first."}],"temperature":0,"max_tokens":512,"stop":["Question:"],"chat_template_kwargs":{"thinking":true,"reasoning_effort":"low"}}'
```

Non-thinking requests must still honor `stop` on content.


MiaAI-Lab:
Landed on `main` as `3c9576c` (merge of this branch). Live-confirmed on the 2× Spark pair; changelog updated.

Issue #21, opened 2026-08-04
encode_arguments_to_dsml corrupts tool-call arguments when arguments is a dict


jodyalbritton:
This is an upstream issue, but reporting here because I hit it following the recs in the recipe.  Just a heads up for anyone else who has issues with tool call encoding. 

Where: encoding/encoding_dsv4.py, encode_arguments_to_dsml, lines 151–178 (bug at 164–167), checkpoint deepseek-ai/DeepSeek-V4-Flash-0731 rev 9e165c30e2704aec5d9d593cce3eebd58bbef1cb

The regression:

# current (broken) — 164-167
```
try:
    arguments = json.loads(tool_call["arguments"])
except Exception as err:                             # err unused
    arguments = {"arguments": tool_call["arguments"]} # wraps a dict input

# previous (correct) — as vendored in vllm/tokenizers/deepseek_v4_encoding.py
if isinstance(tool_call["arguments"], str):
    arguments = json.loads(tool_call["arguments"])
else:
    arguments = tool_call["arguments"]
```

Repro:

encode_arguments_to_dsml({"name": "list_services", "arguments": '{"kind": "airo"}'})
encode_arguments_to_dsml({"name": "list_services", "arguments": {"kind": "airo"}})

<｜DSML｜parameter name="kind" string="true">airo</｜DSML｜parameter>          ← correct
<｜DSML｜parameter name="arguments" string="false">{"kind": "airo"}</…>       ← corrupt

Impact: this function re-renders prior assistant tool calls into the prompt. Once one turn is corrupted, the model sees its own history in that shape and imitates it — emitting a single arguments object with the tool name missing, and unable to recover, since it's correctly matching what its context shows. First call in a session works; everything after fails. Single-turn tests don't catch it.

Fix: dispatch on type, keep the wrap only as a real fallback for a non-JSON string:

```
raw = tool_call["arguments"]
if isinstance(raw, dict):
    arguments = raw
else:
    try:
        arguments = json.loads(raw)
    except (TypeError, ValueError):
        arguments = {"arguments": raw}
```

pabl0rg:
which upstream project is this a bug in?

MiaAI-Lab:
Fixed in the latest update and pushed to `main`.

[...]

Issue #2, opened 2026-07-13
Rootless docker


nicola:
Did you manage to run this with a rootless docker setup?

From gpt-5.6-sol analyzing my setup:

> Started the launcher and identified the failure.
> 
>  Root cause: both nodes use the rootless Docker context. Rootless
>  Docker’s network_mode: host exposes only its private network:
> 
>  ```text
>    10.0.2.100 172.17.0.1
>  ```
> 
>  It cannot access the physical RoCE addresses. Consequently, vLLM fails
>  on dgx1 with:
> 
>  ```text
>    zmq.error.ZMQError: Cannot assign requested address
>    addr='tcp://169.254.186.219:57707'
>  ```
> 
>  The deployment requires host networking and /dev/infiniband, so use
>  rootful Docker on both nodes.

Have you tried to run this?

nicola:
answer is yes, i just needed to upgrade docker

PR #46, opened 2026-08-13
Fix high/max reasoning loops and tool calls

### LINKED ISSUE #37: reasoning_effort=high + tools degenerates (leaked DSML, no tool_calls); thinking_token_budget does not save it
## Summary

Follow-up to #34 / `524054d`, but a **different failure mode**.

On recipe `018c6bc`, `reasoning_effort: "high"` (also via `chat_template_kwargs.reasoning_effort`) **plus `tools`** degenerates: garbage reasoning and leaked `</｜DSML｜tool_calls>` with **no structured `tool_calls`**. `low` / `medium` + tools are clean. `thinking_token_budget` does **not** fix it.

`524054d` (default thinking budget 32k) does not address this and is the wrong default for agent tool loops.

## Environment

- Recipe commit: `018c6bc`
- Image: `ghcr.io/anemll/dspark-vllm-gx10:0.1.1`
- 2x DGX Spark, TP=2
- Direct non-streaming `POST /v1/chat/completions` (not a harness parse bug)
- Server stayed healthy (HTTP 200)

## Reproduction

One `echo_tool` function, `temperature: 0`, `max_tokens: 200`, unique salt per trial. Hermes-shaped body:

```json
{
  "model": "deepseek-v4-flash",
  "messages": [{"role": "user", "content": "SALT=<id> Call echo_tool with token=<id>. No prose."}],
  "tools": [{"type": "function", "function": {"name": "echo_tool", "parameters": {"type": "object", "properties": {"token": {"type": "string"}}, "required": ["token"]}}}],
  "max_tokens": 200,
  "temperature": 0,
  "reasoning_effort": "high",
  "chat_template_kwargs": {"thinking": true, "reasoning_effort": "high"}
}

Results (n=3 each)

┌──────────────────┬────────────────────────────────────────┐
│ Request          │ Result                                 │
├──────────────────┼────────────────────────────────────────┤
│ reasoning        │ 3/3 structured tool_calls              │
│ _effort: low +   │                                        │
│ tools            │                                        │
├──────────────────┼────────────────────────────────────────┤
│ medium + tools   │ 3/3 structured tool_calls              │
├──────────────────┼────────────────────────────────────────┤
│ high + tools     │ 3/3 fail — empty or garbage reasoning  │
│                  │ (<｜begin▁of▁sentence｜># (a) 1.1.1…)  │
│                  │ and leaked </｜DSML｜tool_calls>; no   │
│                  │ tool_calls                             │
├──────────────────┼────────────────────────────────────────┤
│ high + tools +   │ still fails (no clean think-start, so  │
│ thinking_token_  │ the #31 hook never fires)              │
│ budget: 32       │                                        │
└──────────────────┴────────────────────────────────────────┘

Cause

REASONING_EFFORT_PROMPTS["high"] prepends “Absolute maximum… do not stop reasoning”, which fights the tools template (“finish think BEFORE any DSML tool call”).

Suggested fix

If the turn has tools, do not inject the high/max prefix. Keep thinking_mode. Leave no-tools high/max unchanged.

That change measured 3/3 clean tool_calls for high+tools, with or without a 512-token budget.

ashhart:
## Summary

- replace exhaustive high/max prefixes with bounded, goal-directed reasoning guidance
- keep thinking enabled but omit the conflicting effort prefix on tool-bearing turns
- apply and sync the encoder hotfix in the standard two-node startup path
- add CPU tests for atomic application, idempotence, unknown anchors, and file round-trips

## Why

The existing high/max prefixes require exhaustive deliberation and conflict with the DSML instruction to finish reasoning before a tool call. In practice this can cause repeated analysis, leaked DSML, exhausted output budgets, or missed tool calls.

## Validation

- 4/4 focused CPU regression tests pass
- Python compilation and `git diff --check` pass
- live high/max tool-call gate: 6/6 exact tool calls, with no leaked DSML
- combined live answer/tool gate: 10/10 pass

Fixes #37.


MiaAI-Lab (changes requested):
Thanks for this — I verified the implementation and the mechanics are solid. The anchors in `hotfix-encoding-dsv4-issue37.py` match the real stock encoder (`encoding_dsv4.py`), the patch applies + is idempotent, all 4 CPU regression tests pass from a proper repo layout, and the compose/worker wiring mirrors the already-merged issue #21 hotfix cleanly. The guard change (`and not tools`) correctly fixes the first-request high+tools failure mode described in #37.

Requesting changes before merge for three reasons:

### 1. This PR is still a Draft
`draft: true`. Please mark it "Ready for review" before merge — drafts signal not-ready and shouldn't be merged as-is.

### 2. `Fixes #37` overstates the scope and would prematurely close the issue
The reporter on #37 confirmed #46 fixes the **original first-request** bug, but explicitly noted a **second, uncovered failure** ([#37 comment](https://github.com/MiaAI-Lab/DeepSeek-v4-Flash-DSpark-2x-DGX-Spark/issues/37#issuecomment)):

> "#46 looks like the right fix for the original first-request bug… There is a second failure that #46 does not cover."
>
> After the prefix skip, a single request with tools still works. What breaks is the next request after a couple of real tool rounds, when the body includes the official prior `tool_calls` + tool results plus a Hermes-sized tools list (~20 functions)… 6/6 leaked DSML (thinking on or off).

Merging with `Fixes #37` will auto-close #37 and bury the remaining third-round re-encoding failure (~15.5k prompt tokens, official tool history + large tools list). That harder bug is out of scope here but still needs to stay tracked.

Please change `Fixes #37` → `Refs #37` / `Partially addresses #37`, and either:
- leave #37 open, or
- open a follow-up issue for the "prior `tool_calls` re-encoding + large tools list" failure and have #46 reference both.

The PR title ("…and tool calls") also implies more coverage than the first-request-only fix provides — worth softening.

### 3. The no-tools `high`/`max` prompt rewrite is out of scope vs. the issue's requested fix
The reporter's suggested fix was minimal and explicit: *"If the turn has tools, do not inject the high/max prefix. Keep thinking_mode. Leave no-tools high/max unchanged."* The guard change alone implements that.

This PR additionally rewrites the **no-tools** `high`/`max` prompt text to inject advisory 128/768/1536 "token cap" guidance. That changes behavior for **all** non-tool high/max turns, was not requested, and the CPU tests only exercise patch mechanics — not output quality. The caps are prompt-text only and aren't enforced.

Two options:
- **(preferred)** Ship the guard-only change (`and not tools`) here and move the no-tools prompt-text rewrite to a separate, behavior-validated PR so high/max non-tool answer quality can be vetted independently.
- **(or)** Keep both in this PR, but soften the prompt text to a guidance phrasing rather than specific unreachable token targets, and add a CI/behavior note documenting why the caps are advisory-only.

What's good and shouldn't change: the split `patch_text` status model (`applied|skipped|missing`, refuse partial), idempotence, and the consistent compose+scp wiring. If #2 and #3 (and #1) are addressed, the code quality here is merge-ready.

plotarmordev:
Could you rebase this onto current main? It conflicts as-is (main moved a lot this week). Happy to re-test right after — thanks!

plotarmordev:
Closing because there is no valid test target at this head:

- The PR is a stale draft with unresolved conflicts and outstanding requested changes.
- Issue #37 no longer reproduces on current `main`, so the baseline arm is clean and an A/B run cannot demonstrate a fix.

If the defect reproduces again on current `main`, please open a new/rebased PR with the exact failing prompt and a reduced, fail-closed patch.

Issue #45, opened 2026-08-13
Upgrade base image to a vLLM that implements Concurrent Partial Prefill (unblocks issue #43's whole-request fairness bar)


MiaAI-Lab:
## Context

Issue #43 reports whole-request per-lane decode min/max ≈ 0.10–0.24 across six cold cells (32K/62K x2/x4/x8) and asks for ≥ 0.50. The root of that asymmetry is **serialized prefill** — the #27 fix (`max_num_partial_prefills=1`) admits only one in-flight prefill at a time, which staggers TTFT across lanes. Per-step decode service is already fair (the #43 floor in PR #44 keeps decode skips at 0); the spread is TTFT stagger, a direct consequence of the #27 fix.

The only scheduler lever that crosses #43's bar is **concurrent partial prefills** (`max_num_partial_prefills>1`), which overlaps prefills and collapses the stagger. Simulation of the vLLM v1 scheduler shows the direction is real and monotonic:

| cap | 62K×8 whole-request min/max | decode skips | makespan |
|---:|---:|---:|---:|
| 1 (current #27) | 0.42 | 0 | 752 |
| 2 | 0.63 | 0 | 504 |
| 3 | 0.72 | 0 | 442 |
| fully parallel | 1.00 | 0 | 318 |

Post-TTFT (decode-window-only) min/max stays 1.00 at every cap. So raising the cap satisfies both #43's bar _and_ its no-makespan-regression ask.

## Problem on the current base image

`ghcr.io/anemll/dspark-vllm-gx10:0.1.1` (vLLM `0.25.2.dev0+g752a3a504.d20260714`) does **not implement** concurrent partial prefill. `vllm/engine/arg_utils.py:_check_feature_supported()` rejects it before engine init:

```python
# No Concurrent Partial Prefills so far.
if (
    self.max_num_partial_prefills != SchedulerConfig.max_num_partial_prefills
    or self.max_long_partial_prefills
        != SchedulerConfig.max_long_partial_prefills
):
    _raise_unsupported_error(feature_name="Concurrent Partial Prefill")
# -> NotImplementedError: Concurrent Partial Prefill is not supported.
```

[...]

Issue #43, opened 2026-08-13
[Bug] #27 remains unfair with exact 32K/62K x2/x4/x8 cold lanes despite both advertised fixes


bpc-oss:
## Summary

The current `main` #27 solution is installed exactly (`max_num_partial_prefills=1` admission hotfix plus `long_prefill_token_threshold=1024`), but a fresh six-cell retest on the original reporter deployment still shows severe cold/mixed per-lane decode asymmetry. The original issue is closed, so this follow-up tracks the remaining scheduler defect.

## Exact runtime

- 2x DGX Spark GB10, TP=2
- vLLM `0.25.2.dev0+g752a3a504.d20260714`
- MiaAI `27a5569513db` lineage; #27 scheduler postimage SHA-256 `46d56c3dffb4716bd8850b8b05b430e85578af2bb955c4efe75362ff98d05dac`
- `max_num_partial_prefills=1` marker verified in live scheduler
- `long_prefill_token_threshold=1024`
- `max_num_seqs=8`, `max_num_batched_tokens=8192`, GPU util `0.82`, MTP5
- async scheduling, chunked prefill, prefix caching, NVFP4 DS-MLA
- Issue #36 v2 coordinator + retention interval 4096
- user checkpoint `drowzeys/keys-DeepSeekV4-Flash-GA-0731-Dspark-Abliterated-32-32` at pinned revision; served alias `deepseek-v4-flash-0731-ablit`

## Strict reproduction

For each cell, start idle and send simultaneous independent lanes with:

- exact 32,768 or 63,488 input tokens per lane;
- unique cold prefix per lane (all `cached_tokens=0`; server cache-hit delta 0);
- 256 output tokens, `min_tokens=256`, `ignore_eos=true`, temperature 0;
- streaming; decode rate = `(completion_tokens-1)/(last_token-first_token)`;
- server counters close exactly; all requests HTTP 200 and full length.

## Results

| shape | min tok/s | max tok/s | min/max |
|---|---:|---:|---:|
| 32K x2 | 9.28 | 38.97 | 0.238 |
| 32K x4 | 5.68 | 35.24 | 0.161 |
| 32K x8 | 4.85 | 37.35 | **0.130** |
| 62K x2 | 6.38 | 46.46 | 0.137 |
| 62K x4 | 5.78 | 41.69 | 0.139 |
| 62K x8 | 4.45 | 41.60 | **0.107** |

[...]

Issue #39, opened 2026-08-13
2026-08-13 commits (#26/#27/#31/#34): 4.7x decode regression at 32k ctx, plus tool-name fidelity loss — measured A/B on 2x GB10


ecohash-co:
Thanks for this recipe — it has been running well for us since 2026-08-12.

We adopted today's four commits (#26, #27, #31, #34) together on 2× DGX Spark GB10 (TP=2), measured
before/after, and hit **two independent regressions**. We have reverted to `cbd719f` and re-verified
clean. Sharing the measurements in case they reproduce on your cluster.

## Setup

```
image        ghcr.io/anemll/dspark-vllm-gx10:0.1.1@sha256:a839484… (digest-pinned, identical both nodes)
model        deepseek-ai/DeepSeek-V4-Flash-0731 @ 9e165c30 (pinned)
topology     2× GB10, TP=2, ConnectX-7 p2p
serve        MAX_MODEL_LEN=1048576  MAX_NUM_SEQS=6  MAX_NUM_BATCHED_TOKENS=8192
             GPU_MEMORY_UTILIZATION=0.835  MTP_NUM_TOKENS=5  kv-cache-dtype nvfp4_ds_mla
             DEFAULT_THINKING=max  --enable-prefix-caching --enable-chunked-prefill
before/after identical except the four commits (cbd719f ↔ 524054d)
```

---

## Regression 1 — `DEFAULT_THINKING_TOKEN_BUDGET` costs 4.7× decode throughput at 32k context

Decode throughput measured client-side from the **first streamed token** (so prefill is excluded and
the two context depths are comparable), n=2 per point, median:

| stack | @1k ctx | @32k ctx | retained |
|---|---|---|---|
| `cbd719f` | 15.8 tok/s | **14.1 tok/s** | 0.89× |
| + #26/#27/#31/#34, `DEFAULT_THINKING_TOKEN_BUDGET=32768` | 14.3 tok/s | **3.0 tok/s** | **0.21×** |
| + #26/#27/#31, `DEFAULT_THINKING_TOKEN_BUDGET=0` | 15.6 tok/s | **13.8 tok/s** | 0.88× |

The third row isolates it: with **#26/#27/#31 still applied** and only the budget default turned off,
throughput returns to baseline. So the cost is specifically the budget path.

### Why

[...]

Issue #27, opened 2026-08-12
[Bug] Severe decode lane starvation at x8 concurrency (1.10-21.69 tok/s, zero preemptions)


bpc-oss:
## Summary

At x8 concurrency on the 2x DGX Spark recipe, decode service is extremely uneven across requests. At 62K x 8, one lane decoded at 21.69 tok/s while another decoded at 1.10 tok/s in the same cold wave. The slow lane's mean inter-token latency reached 912 ms.

The effect grows with prompt length and persists on warm replays whenever prefix-cache hits are lost:

| Shape | Cold decode min / median / max | Cold mean ITL range | Cold TTFT range | Wave elapsed |
|---|---:|---:|---:|---:|
| 8K x 8 | 7.06 / 11.67 / 19.19 tok/s | 52.1-141.6 ms | 7.843-32.209 s | 46.119 s |
| 16K x 8 | 4.04 / 6.82 / 20.70 tok/s | 48.3-247.7 ms | 12.194-64.863 s | 78.496 s |
| 32K x 8 | 2.12 / 3.95 / 20.53 tok/s | 48.7-472.6 ms | 20.017-131.609 s | 144.237 s |
| 62K x 8 | 1.10 / 2.10 / 21.69 tok/s | 46.1-912.1 ms | 33.332-262.125 s | 273.882 s |

There were **zero preemptions** in every wave, all requests returned HTTP 200 with the exact input and output lengths, and MTP remained active. This looks like severe decode scheduling unfairness or starvation while other lanes are still in prefill, rather than request failure.

## Environment

- Hardware: 2x NVIDIA DGX Spark / GB10, TP=2
- Recipe revision tested: [`5448679afe7a1e6e3c73a76c65eb4eea0b1d58d1`](https://github.com/MiaAI-Lab/DeepSeek-v4-Flash-DSpark-2x-DGX-Spark/commit/5448679afe7a1e6e3c73a76c65eb4eea0b1d58d1)
- Current `main` is three commits ahead (`cbd719f`); the intervening HF-cache warning, issue #24 structured-output hotfix, and image digest pin do not appear scheduler-related.
- Image: `ghcr.io/anemll/dspark-vllm-gx10:0.1.1`
- Local image ID: `sha256:3430d6614a8e2925f34d059af6caf05aff42387326db4d05639a60f10f2654d8`
[...]

PR #47, opened 2026-08-13
Adapt prefill chunks around active decode

ashhart:
## Summary

- retain large prefill chunks for idle, prefill-only work
- switch to a configurable 1024-token chunk while a decode lane is live
- detect live speculative decode from durable output/request state across the async scheduler registry
- preserve compatibility with the existing issue #43 fairness/diagnostic hotfix
- add environment documentation and CPU regression guards

## Why

A single static chunk size forces a latency/throughput tradeoff: larger chunks improve isolated cold ingestion, but can create multi-second inter-token gaps when a long prompt arrives during active generation. This change selects the threshold from scheduler state so those two workloads can be tuned independently.

## Validation

- focused adaptive-prefill tests: 4/4 pass
- patch applies to the real scheduler from the pinned container image
- the issue #43 patch applies afterward and the combined scheduler compiles
- reapplying the issue #43 layer is idempotent
- measured cold 32K ingestion improved from about 18.4s to 16.2s; mixed-load p99 after initial tokenization stayed around 0.69-1.09s in the tested setup

The early tokenizer stall remains outside this scheduler change and is not claimed as fixed.

### DIFF
--- a/.env.dspark.example
+++ b/.env.dspark.example
@@ -114,6 +114,10 @@ MAX_NUM_BATCHED_TOKENS=8192
 # hotfix (max_num_partial_prefills actually enforced) it collapses x8 spread
 # from ~46x to ~1.05x. Safe across MAX_NUM_SEQS and prompt lengths.
 LONG_PREFILL_TOKEN_THRESHOLD=1024
+# The issue-27 adaptive scheduler uses this smaller threshold only while a
+# decode lane is active. This permits a larger LONG_PREFILL_TOKEN_THRESHOLD
+# for faster idle/cold ingestion without multi-second reply stalls.
+DSPARK_DECODE_ACTIVE_PREFILL_THRESHOLD=1024
 
 # Issue #43: max concurrent in-flight partial prefills. Raising to 2/3
 # overlaps prefills and is the lever that improves the whole-request min/max
--- a/docker-compose.dspark.yml
+++ b/docker-compose.dspark.yml
@@ -115,6 +115,9 @@ services:
       # Checkpoint dspark_block_size is 5; k<5 truncates draft blocks (silent on 0.25.2 image).
       MTP_NUM_TOKENS: "${MTP_NUM_TOKENS:-5}"
       DEFAULT_THINKING: "${DEFAULT_THINKING:-low}"
+      # Large chunks when prefill-only, bounded chunks while a reply decodes.
+      # Must be a multiple of the scheduler block size.
+      DSPARK_DECODE_ACTIVE_PREFILL_THRESHOLD: "${DSPARK_DECODE_ACTIVE_PREFILL_THRESHOLD:-1024}"
       DSPARK_SKIP_HOTFIX: "${DSPARK_SKIP_HOTFIX:-0}"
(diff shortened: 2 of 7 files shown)


bpc-oss:
Live validation of PR #47 (`6aba5a75791b327ed4b62d67e5028d1ff28814b3`) on the reporter's 2x DGX Spark deployment: **FAIL for Issue #45's whole-request fairness gate; do not adopt as the fairness fix.**

Candidate identity:

- image `sha256:1da3e7cfba6660793ea21c0b1712d6d6e714e8b153768e34b44f1bae2f8cfc80`
- current DSpark base + #27 serialized-prefill safety + PR #44 decode floor + PR #47 adaptive mixed-prefill threshold
- adaptive threshold `1024`, idle chunk `8192`, `max_num_seqs=8`, retention `4096`, MTP enabled
- custom checkpoint/revision and GPU utilization unchanged
- candidate residency remained restart=0/OOMKilled=false; cold preemptions=0

The exact x8 cold shapes still show severe whole-request decode asymmetry:

| Cell | min tok/s | max tok/s | min/max |
|---|---:|---:|---:|
| 32Kx8 | 5.219 | 43.084 | **0.1211** |
| 62Kx8 | 5.118 | 38.865 | **0.1317** |

Both are far below the Issue #45 acceptance line of 0.50 and are effectively unchanged from the original failure class. The result confirms that reducing the chunk only after a decode lane is already active does not remove the large TTFT/prefill-wave stagger that dominates whole-request fairness.

Cache retention remains healthy and independent of this failure:

- 32Kx8 warm1/warm2: every lane cached 32,512 tokens
- 62Kx8 warm1/warm2: every lane cached 63,232 tokens
- 256Kx1 warm20: 20/20 full hits, 261,888 cached tokens, no miss
- 512Kx1 warm20: 20/20 full hits, 524,032 cached tokens, no miss
- all listed waves reported 0 preemptions

Strict API regression under this candidate/runtime also failed: catalog/tokenize PASS; Chat FAIL (`chat content invalid`); appended multiturn FAIL (`shared-prefix prompt token count invalid`); Responses FAIL (`responses capability matrix failed`); disconnect FAIL (`disconnect request was never observed running`). These are reported separately from the scheduler verdict.

Issue #39 supplemental checks did not reproduce its universal regression path in this deployment: forced `ping` tool calls were 10/10 exact with valid `{}` arguments and no blank response; single-lane 2K to 32K cold decode retained 80.6% (59.29 -> 47.78 tok/s). The deployment does not set `DEFAULT_THINKING_TOKEN_BUDGET`, so this does not validate PR #48's explicit-budget path.

The dedicated six-cell discriminator was blocked before network requests by a local evidence-runner authority check defect. This does not weaken the two failing x8 measurements above: they are the same exact 32Kx8/62Kx8 payload shapes, completed successfully, with per-lane timings, cache counters, MTP counters and zero-preemption counters captured.

Conclusion: PR #47 may improve mixed-load ITL after decode becomes active, but it does **not** resolve Issue #45's whole-request fairness requirement. A subsequent fix needs to address admission/TTFT stagger, and must be tested live against the exact six cold cells rather than inferred from post-TTFT ITL alone.

ashhart:
Closing because live 2x DGX Spark validation showed that this change does not meet Issue #45's whole-request fairness gate. It may improve post-TTFT mixed-load ITL, but it does not address the admission/TTFT stagger.

Issue #34, opened 2026-08-13
Issue #31 follow-up: standard OpenAI harnesses still get blank turns without a server-side thinking budget


Kristianaaron:
## Summary

This is a follow-up to #31 after applying commit `018c6bc`.

The V2 `thinking_token_budget` hotfix works when the client explicitly sends the field. However, common OpenAI-compatible agent harnesses (observed with OMP/pi) do not expose a reliable way to add this non-standard top-level field. With the recipe default `DEFAULT_THINKING=max`, those clients can therefore still receive a blank assistant turn.

The practical result is that #31 is fixed at the API capability level, but remains reproducible through otherwise compatible agent harnesses.

## Environment

- Recipe commit: `018c6bc3620b775a0c4feaf8f9350052a9657291`
- Image: `ghcr.io/anemll/dspark-vllm-gx10@sha256:a83948492cf13df455170fb42885f5ef4db54fefe0feff0f841ecbff464ac9d8`
- vLLM: `0.25.2.dev0+g752a3a504.d20260714`
- 2x DGX Spark, TP=2 over RoCEv2
- Official `DeepSeek-V4-Flash-0731` checkpoint, `ABLITERATED=0`
- `DEFAULT_THINKING=max`
- `MTP_NUM_TOKENS=5`
- `MAX_NUM_SEQS=6`
- `MAX_NUM_BATCHED_TOKENS=8192`
- `GPU_MEMORY_UTILIZATION_TEXT=0.835`
- Issue #26, #27, #31 and grammar hotfixes confirmed applied on both ranks
- Server stayed healthy throughout; requests returned HTTP 200

## Controlled reproduction

Prompt:

```
I would like to continue working on Atlas. Briefly state the next concrete action and then give a useful answer.
```

Request shape (authentication omitted):

```json
{
  "model": "deepseek-v4-flash-dspark",
  "messages": [{"role": "user", "content": "<prompt above>"}],
  "temperature": 0,
  "max_tokens": 256
}
```

Results:

| Request | Trials | Blank `content` | Finish reason | Reasoning chars |
|---|---:|---:|---|---:|
| No `thinking_token_budget` | 4 | **3/4** | `length` for blank turns | 1083–1156 |
| `thinking_token_budget: 64` | 4 | **0/4** | `stop` | 258–305 |
[...]

PR #48, opened 2026-08-13
Fix V2 thinking budgets without decode host scans

### LINKED ISSUE #31: DEFAULT_THINKING=max returns content: null on truncation, with no way to bound reasoning (thinking_token_budget 400s; V2=0 breaks dspark)

ashhart:
## Summary

- support explicit `thinking_token_budget` on the DSpark V2 runner
- keep request state, boundary enforcement, and accepted-token observation on GPU
- preserve the existing sampler fast path when the field is omitted
- warm the budget kernels in the startup smoke and add CPU regression gates
- allow the benchmark harness and Pi example to exercise explicit budgets

This replaces the withdrawn #31/#34 implementation. It does not restore its Python prefix scans, device-to-host copies, or omit-field defaults.

## Live TP=2 validation

On two DGX Sparks with the official 0731 checkpoint:

- budgets `0`, `16`, `32`, `64`, and `256` returned visible answers
- four simultaneous requests with different budgets completed without state bleed
- an auto tool call and the following tool-result turn completed without a loop
- short-context budgeted decode: **67.2 tok/s**
- fresh-prefix 65K budgeted decode: **59.3 tok/s**; prefill: **1,815 tok/s**
- fresh-prefix 65K no-budget control: **37.2 tok/s** at a 1,024-token length stop; prefill: **1,693 tok/s**
- health remained 200 with no CUDA/Triton runtime errors after the tests

The output lengths differ in the 65K control, so that pair establishes absence of the prior ~4.5 tok/s cliff rather than a strict A/B speedup claim.

## Tests

- `bash scripts/ci-validate.sh` on Linux: pass
- patch apply and idempotent reapply against the installed V2 source: pass
- real CUDA kernel boundary / accepted-token smoke: pass
- live Pi low-thinking request with a custom budget: pass

Closes #31.


MiaAI-Lab:
Merged locally and pushed to `main` as `2689b1f` (your branch was behind `main` after the 2026-08-14 README rewrite, so GitHub showed `mergeable: dirty`; the only conflicts were README.md / CHANGELOG.md — both resolved by keeping our rewrite and folding in your capability statements, plus correcting the stale "rejects HTTP 400" lines to opt-in framing).

### Live validation on 2× DGX Spark TP=2 (official 0731 @ 9e165c30, Anemll 0.1.1 image)

- **Patch applies cleanly**: all 11 monkey-patch anchor strings match the pinned image byte-for-byte; `docker logs` shows `[issue31-gpu-hotfix] done` (11/11 sites). Idempotent on re-run (rc=0, `already applied` everywhere).
- **Boot smoke passes**: the new `thinking_token_budget:1` probe now exercises the budget Triton kernels successfully on GB10/SM12.1 — no CUDA/Triton errors, so a healthy boot no longer false-fails under `DEFAULT_THINKING=max`.
- **Omit-field path unchanged (off by default)**: `POST /v1/chat/completions {"max_tokens":256,"temp":0}` -> `stop`, `content:"PING-OK-17"`. No budget injected, no kernel fires, no sync — this is the exact pre-#48 fast path. `thinking_token_budget` is opt-in per request; the server injects no default.
- **New capability works**: `thinking_token_budget:512` + `effort:high` -> `stop`, reasoning 2344 chars, content 12457 chars, no HTTP 400. `budget:0` -> reasoning 0 chars, content returned. Natural ` Immutable Map Idx Value ~ Nat ` untouched.
- **No #39 decode regression**: A/B c=1-6 @ 256/4096, budget=1024 vs no-budget — per-stream median decode rate at-or-above baseline at every concurrency (d -0.0% to +19.4%, never materially below). Aggregate swings are cap-hit artifacts, not budget-kernel cost. The withdrawn host-scan cliff is structurally absent.
- **CPU CI gate** (`scripts/ci-validate.sh`) passes 24/24, including the new guards: `GPU thinking-budget hot path has no CPU sync or token-buffer scan`, `withdrawn CPU-scanning thinking-budget patch stays absent`, `compose applies GPU-resident V2 thinking budget`.

### Closes

- Closes #48 (this PR).
- Closes #31 (the capability gap — `thinking_token_budget` is again accepted on the V2 runner, without the withdrawn host-scan cost).
- Closes #34 (clients can now bound reasoning without server-side omit defaults; blank-turn risk on `DEFAULT_THINKING=max` + small `max_tokens` is mitigated by sending the field).

### Not closed by this merge (deliberate)

- #56 stays open: `DEFAULT_THINKING=max` is unchanged in `.env.dspark.example`, and omit-field requests keep the stock fast path. #48 gives the *capability* to bound reasoning; #56 asks to reconsider the shipped *default*, which is a separate decision (lower to `high`, or document the ~50k-char magnitude).

Closing as merged. Thanks @ashhart.

Issue #26, opened 2026-08-12
[Bug] Prefix cache collapses at 32Kx8 and 62Kx8 while 8Kx8/16Kx8 and single-lane warmx20 pass


bpc-oss:
## Summary

On the current 2x DGX Spark recipe, eight independent prefixes are retained reliably at 8K and 16K, but prefix-cache retention collapses at 32K and 62K:

- `8K x 8`: 8/8 lanes hit on both immediate exact replays.
- `16K x 8`: 8/8 lanes hit on both immediate exact replays.
- `32K x 8`: 0/8 on warm replay 1; only 1/8 on warm replay 2.
- `62K x 8`: only 1/8 on warm replay 1; 0/8 on warm replay 2.

This is not explained by a general inability to retain large prefixes. In a control run on the same live engine, a single 256K prefix survived 20/20 exact replays and a single 512K prefix survived 20/20 exact replays, with the expected full-block cache hit every time.

The evidence therefore points to an interaction with **multiple concurrently completed prefixes** (retention, eviction, or later prefix discoverability), rather than simple age-based decay of one prefix. I am not claiming a root cause yet.

## Environment

- Hardware: 2x NVIDIA DGX Spark / GB10, TP=2
- Recipe revision tested: [`5448679afe7a1e6e3c73a76c65eb4eea0b1d58d1`](https://github.com/MiaAI-Lab/DeepSeek-v4-Flash-DSpark-2x-DGX-Spark/commit/5448679afe7a1e6e3c73a76c65eb4eea0b1d58d1)
- Current `main` is three commits ahead (`cbd719f`); the intervening changes are the HF-cache warning, the issue #24 structured-output hotfix, and the image digest pin. None appears related to prefix-cache retention.
- Image: `ghcr.io/anemll/dspark-vllm-gx10:0.1.1`
- Local image ID: `sha256:3430d6614a8e2925f34d059af6caf05aff42387326db4d05639a60f10f2654d8`
- vLLM: `0.25.2.dev0+g752a3a504.d20260714`
- Checkpoint: `drowzeys/keys-DeepSeekV4-Flash-GA-0731-Dspark-Abliterated-32-32`
- Checkpoint revision: `6c47916f85e52b5e712223ca8f93952f90255714`

The serving configuration matches the MiaAI recipe except for the checkpoint, `max_num_seqs`, and GPU utilization:

| Setting | Value |
|---|---:|
[...]

PR #53, opened 2026-08-14
fix: emit a generation header when a request ends with an assistant turn

### LINKED ISSUE #52: Request ending with an assistant message renders without a generation header — model generates from a dead state (self-sustaining empty-turn loop)
## Summary

When a chat-completions request's `messages` array **ends with an `assistant` message**, the rendered prompt ends on a bare EOS with no generation header. The model is then asked to generate from a closed, turnless state and produces either an immediate EOS (1–2 tokens, `finish_reason: "stop"`) or raw `<｜DSML｜tool_calls>` markup emitted as plain text.

Because the tool-call parser silently discards a DSML block that contains no valid invoke, the client records an empty turn — which leaves the transcript ending with an assistant message again. **The failure is self-sustaining**: one bad turn locks an agent loop until the session is abandoned.

Requests reach this shape routinely: any harness that retries after a mid-stream error and re-sends the partial assistant turn produces exactly it. That is why the same prompt works for a long session and then suddenly does not.

## Cause

`encoding_dsv4.py` → `render_message()` appends the generation header (`ASSISTANT_SP_TOKEN` + thinking token) only for `user` / `developer`:

```python
elif messages[index].get("role") in ["user", "developer"]:
    # Normal generation: append Assistant + thinking token
    prompt += ASSISTANT_SP_TOKEN
    ...
```

A trailing `assistant` message has `task is None` and a role that is not `user`/`developer`, so nothing is appended after its EOS.

Rendered tails, same transcript:

```
assistant-final: '...to the reviewer for a fresh verdict:<｜end▁of▁sentence｜>'
user-final:      '...verdict:<｜end▁of▁sentence｜><｜User｜>continue<｜Assistant｜><think>'
```

## Reproduction

```bash
  -d '{"model":"<served-name>",
       "messages":[
         {"role":"system","content":"You are a coding agent."},
         {"role":"user","content":"Has the fix been committed yet? Answer precisely."},
         {"role":"assistant","content":"Not yet — nothing was committed or pushed. The working tree still has the bench branch checked out."}
       ],
       "max_tokens":400,"temperature":0}'
```

Before the fix this returns ~16 generated tokens that are a **continuation fragment** of the previous sentence (e.g. `" on branch …, with uncommitted changes present."`) rather than an answer. With a partial trailing turn it instead returns 1–2 tokens and `finish_reason: "stop"`.

## Measurements

Generated tokens for one prompt, sent raw to `/v1/completions` so nothing differs but the trailing token(s):

| trailing assistant turn | stock (EOS) | reopen turn (`wo_eos`) | EOS + generation header |
|---|---|---|---|
| partial (ends mid-thought) | 32 | 208 | 260 |
| complete (finished answer) | 16 — sentence fragment | **1 — empty, dead** | 140 — healthy |

**Please note the middle column.** `assistant_msg_wo_eos_template` already exists in the file and reopening the trailing turn is the tempting one-line fix. It is wrong: a *complete* assistant turn has nothing left to continue, so the model emits EOS immediately and the dead state simply moves. We shipped that variant first and it visibly degraded output — half-restarted answers and mangled tokens — before we caught it.

The EOS-closed state is also where the stray markup comes from. Same prompt, differing only by that one token:

- turn left open → 183 tokens, `stop`, coherent prose
- **EOS appended → 400 tokens of raw `<｜DSML｜tool_calls>` / `<｜DSML｜parameter name=...>` emitted as text**

which is why the corrupted turns contain `</parameter>`, `</name>`, `<value>` fragments.

## Fix

PR #<PR> — extend the transition branch so a trailing `assistant` message also receives the generation header, i.e. `add_generation_prompt=True` semantics. Only the final message of a request is affected, and only when it is an assistant turn.

Verified on 2× DGX Spark, TP=2, `ghcr.io/anemll/dspark-vllm-gx10:0.1.1`, vLLM `0.25.2.dev0+g752a3a504.d20260714`, upstream `018c6bc`:

- assistant-final announcement → well-formed tool call instead of an unreasoned wrong-schema one
- transcript already poisoned with 3 degenerate turns → **141 tokens, coherent content and a valid tool call** (was 1 token, empty) — so the fix also recovers a stuck session
- 18-turn streaming soak with tools → 17 turns clean, no empty turns, no stray markup. The one exception was `finish_reason: length` with ~11.3k chars of reasoning, i.e. unrelated reasoning overrun.

## Possibly related

We separately observe tool calls returned with `finish_reason: "tool_calls"` while `completion_tokens` exactly equals `max_tokens`, sometimes carrying unterminated JSON arguments (filed separately as #<C>). A client that stores and replays such a call gets `400 Unterminated string` on every later request, and its retry can leave the transcript assistant-final — a **plausible, not demonstrated** route into this bug.

plotarmordev:
Fixes #52.

## What

`render_message()` appends the generation header (`ASSISTANT_SP_TOKEN` + thinking token) only when the trailing message is `user` / `developer`. When a request's `messages` array ends with an **assistant** message that branch is skipped, so the prompt ends on a bare EOS with no header and the model generates from a closed, turnless state — 1–2 tokens then `stop`, a sentence fragment continuing the previous turn, or raw `<｜DSML｜tool_calls>` markup emitted as text.

This extends that branch so a trailing assistant message also receives the header. Only the final message of a request is affected, and only when it is an assistant turn; consecutive assistant messages mid-transcript are untouched.

This implements `add_generation_prompt=True` semantics — a **new** assistant turn — not prefill continuation.

## Why not `wo_eos`

`assistant_msg_wo_eos_template` already exists and reopening the trailing turn is the shorter change. It is the wrong one. Generated tokens, one prompt, `/v1/completions`:

| trailing assistant turn | stock (EOS) | reopen (`wo_eos`) | EOS + header (this PR) |
|---|---|---|---|
| partial | 32 | 208 | 260 |
| complete | 16 — fragment | **1 — empty, dead** | 140 — healthy |

A complete assistant turn has nothing left to continue, so reopening it emits EOS immediately and merely relocates the dead state. We shipped the `wo_eos` variant first and it degraded output in normal use before we caught it — flagging it here so the shorter fix isn't reached for later.

OpenAI-compatible clients cannot reach the `wo_eos` path anyway (it is only readable from a per-message key), so there is no behaviour to preserve there.

## Verification

2× DGX Spark, TP=2, `ghcr.io/anemll/dspark-vllm-gx10:0.1.1`, vLLM `0.25.2.dev0+g752a3a504.d20260714`, on `018c6bc`:

- assistant-final announcement → well-formed tool call (previously an unreasoned wrong-schema call)
- transcript already poisoned with 3 degenerate turns → 141 tokens, coherent content, valid tool call (previously 1 token, empty). The fix recovers a stuck session, not just prevents entry.
- trailing-user control → unchanged
- 18-turn streaming soak with tools → 17 clean. The one exception was `finish_reason: length` with ~11.3k chars of reasoning (unrelated reasoning overrun), not an empty turn.

## Notes

- Idempotent, and skips cleanly if the encoder was never copied in.
- Mounted read-only and applied in the compose entrypoint after the Issue #21 encoder hotfix, since it patches the file that entrypoint copies into the image; synced to the worker by `start-deepseek-v4-flash-dspark.sh` alongside the existing hotfixes.
- Self-verifies after patching by rendering a trailing-assistant transcript and asserting the header is present.
- If a future checkpoint moves the anchor the script exits non-zero and boot continues unpatched — same failure mode as the other hotfixes in `patches/`.

Happy to adjust placement or naming to match your conventions.


plotarmordev:
Rebased onto `103af68`. Resolved the one conflict in `docker-compose.dspark.yml`: kept your new mounts (#43 decode fairness, #26 v2, suppress-stops) and appended the patch invocation to the encoding block, so it still runs after the Issue #21 encoder hotfix on the file the entrypoint copies in.

Re-checked against current `main`: none of the 16 commits since `018c6bc` touch `encoding_dsv4.py`, `render_message()` or the transition block, so the bug and this fix are unaffected by the thinking-budget revert. This PR is independent of that path — no budget hook involved, and it reproduces on a three-message transcript.

plotarmordev:
Pushed one hardening change: the new condition now also requires `not messages[index].get("wo_eos")`.

Reasoning: a caller that sets `wo_eos` on the final assistant message is asking for prefill continuation, not a new turn, so appending a generation header after a deliberately unclosed turn would be wrong. That path is **unreachable today** — `wo_eos` appears nowhere in the image's vLLM tree outside `encoding_dsv4.py` itself, nothing in the serving layer sets it, and there is no `continue_final_message` support — but it costs nothing and keeps the patch correct if an integration ever does.

I also ran a shape-by-shape regression audit against the deployed encoder, since a no-regression bar is the right one for this file. Rendering is byte-identical for: user/developer-final; non-final user/developer; assistant mid-transcript including non-final members of multi-assistant tails; assistant followed by `latest_reminder`; `latest_reminder`-final; and any message carrying a `task` (that branch takes precedence). Tool transcripts are unaffected — `merge_tool_messages`/`sort_tool_results_by_call_order` only reshape user content blocks and cannot change the final role. With `drop_thinking` on, the final message is never dropped, so the firing branch cannot shift, and an assistant-final message always satisfies `index >= last_user_idx`, giving `<think>` in thinking mode and `</think>` in chat mode — mirroring user-final semantics exactly.

Behaviour changes only when the transcript-final message is an assistant turn without a `task` — the patch's declared scope.

Unrelated but worth flagging separately: one shape I checked is `latest_reminder`-final, which is **also** headerless and would hit the same dead state. I have not seen it in practice and it is out of scope here, but you may want it covered by the same branch.

plotarmordev:
Live A/B on the 2-node DGX Spark deployment, rebased head f08cd6c.

**Environment:** TP head + worker, image `ghcr.io/anemll/dspark-vllm-gx10:0.1.1`, upstream base f752cd0, checkpoint `drowzeys/keys-DeepSeekV4-Flash-GA-0731-Dspark-Abliterated-32-32` (1M ctx), `temperature=0`, `thinking=high`. States: **A** = upstream, no patch (mount/entrypoint/sync removed, verified absent in-container); **B** = exact PR patch file (md5 `86a216e5…`, applied at boot: `[assistant-final-hotfix] patched … / verified: generation header present`). Full container restart between states.

**1. Render ground truth** (in-container `encode_messages`, assistant-final request):

| state | prompt tail |
|---|---|
| A | `…<｜Assistant｜><think></think>…verdict:<｜end▁of▁sentence｜>` — bare EOS, no header (dead shape confirmed) |
| B | `…verdict:<｜end▁of▁sentence｜><｜Assistant｜><think>` — generation header appended (+2 prompt tokens) |

The patch changes exactly what it claims, and nothing else.

**2. Chat probe** (`deadstate_ab.py`, tools offered, `max_tokens=1200`): completion tokens / finish

| case | A (no fix) | B (PR) |
|---|---|---|
| assistant-final (announcement) | 557, `tool_calls`, alive | 498, `tool_calls`, alive |
| assistant-final + 3 degenerate turns | 552, `tool_calls`, alive | 156, `tool_calls`, alive |
| control: trailing user msg | 1200 `length` | 1200 `length` |

**3. Continuation matrix** (rendered prompt → `/v1/completions`, `max_tokens=300`):

| trailing turn | A (bare EOS) | B (EOS + header) |
|---|---|---|
| partial | 300 (`length`); **continues the closed turn** — leading `.` glued to `…verdict:`, then valid DSML tool call | 300 (`length`); **starts a fresh turn** — `Let me look at the repository…</think>` then valid DSML tool call |
| complete | 170 (`stop`); continuation-style `: Let me check the review result.` | 142 (`stop`); fresh-turn text (`The user asked me to fix…`) |

**Honest finding:** the dead-state failure (≤10-token empty `stop` turns, DSML soup) did **not** reproduce on the unpatched baseline — unlike the incident-era measurements in the commit message (32/16 tokens, fragments), the current 0731-GA checkpoint recovers from the bare-EOS tail into valid tool calls on every probe case. So this A/B cannot demonstrate a rescue effect; what it does show:

1. The render delta is real and exactly as designed (verified at the token level, both states).
2. Without the header the model *continues the closed turn* (leading `.`/`:` fragments); with the header it starts a clean new turn — the semantic the patch intends, matching `add_generation_prompt=True`.
3. No regression on any case; control (user-final) rendering is untouched.

Caveat on token counts: identical control prompts produced 244 vs 1200 tokens across boots (spec-decode/NVFP4 numerics are not run-stable), so only the dead-vs-alive signal and the render tails are load-bearing here.

Recommendation unchanged: merge — the patch is correct-by-construction, verified live, and the failure mode it guards against was measured in production on the previous checkpoint even though the current one tolerates it.

plotarmordev:
Follow-up on the rebase delta: the deployed variant of this hotfix carried an extra guard, `and not messages[index].get("wo_eos")`, which this PR drops. Verified live that the guard is unreachable through the served API, so dropping it is safe:

1. `wo_eos` as an extra message key is stripped by the chat API — `/v1/chat/completions/render` output is token-identical with and without it.
2. `add_generation_prompt=False` and `continue_final_message=True` (top-level or via `chat_template_kwargs`) are likewise ignored by this encoder path — all renders byte-identical.
3. `grep -r wo_eos vllm/entrypoints/` → no hits: nothing server-side ever sets the flag. The `wo_eos` branch is only reachable by calling `encode_messages` directly, which no request path does.

For completeness, the encoder-level behavior difference this disposes of: a direct `encode_messages` call with `wo_eos=True` on the final assistant turn renders `…verdict:` (open turn) stock vs `…verdict:<｜Assistant｜><think>` patched. Unreachable in the served recipe.

plotarmordev:
## Maintainer final verification — exact head `f75f962fbb7ad7a3ed8c0b5996c950c78f45776d`

The final head was re-reviewed after its last update.

### Important correction

The first enabled boot at `d4b31daf` failed closed before serving: a review-requested `add_generation_prompt` guard referenced a variable that does not exist in the real checkpoint encoder. The self-check restored the original bytes and both containers refused to serve. The corrected code is commit `0864014`; final head `f75f962` adds evidence docs only. Earlier `aa940bd` → `d4b31daf` history was rewritten with force-with-lease during wording cleanup; later commits are normal descendants.

### CPU and source gates

- `python3 scripts/test-assistant-final-continuation.py -v`: **13/13 PASS**
- `bash scripts/ci-validate.sh`: **PASS** (shell syntax, py_compile 14 files, all CPU recipe/unit gates)
- `git diff --check`: **PASS**
- focused secret scan of changed files: **no matches**
- corrected patcher run against untouched real checkpoint encoder copies inside both live containers: **patched and self-check verified**
- independent adversarial review read the real checkpoint encoder, approved `0864014`, then confirmed `f75f962` is docs-only and **merge-ready with no blockers**

### Serialized live OFF / failure / ON proof

**OFF (`DSPARK_ENABLE_ASSISTANT_FINAL_HOTFIX=0`)**
- full two-rank boot + cluster verification: **PASS**
- effective flag `0` on both ranks
- installed assistant-final marker count `0` on both ranks
- assistant-final `/v1/chat/completions/render`: 98 tokens ending in stock `<think></think>alpha<|end of sentence|>`

**Fail-closed anchor drift**
- deliberate derived test tree changed only the expected anchor
- both containers: `exit=1`, restart/failure state; API never served (`curl rc=56`)
- log: `[FAIL] [assistant-final-hotfix] anchor not found`

**ON (`DSPARK_ENABLE_ASSISTANT_FINAL_HOTFIX=1`)**
- exact corrected code two-rank boot + cluster verification: **PASS**
- effective flag `1` on both ranks; both patched and verified
- ON render preserved every one of the 98 OFF token IDs, then appended exactly token IDs `[128804,128821]` = `<|Assistant|><think>`
- live trailing-assistant completion: `content="alpha beta"`, `finish_reason="stop"`, 40 completion tokens
- manual rerun on both ranks: `already applied and verified` (idempotence)

### Restore and gaps

- owner cluster restore: **PASS**; minimal chat succeeded
- restored active env hashes exactly match pre-test captures:
  - `.env.dspark` `150f43334a123b597ee1047634acc2dca43344482739dcbddd679ce1f73817ef`
  - `.env.abliterated` `fa3f7614bba5f9a8727174d33a1b09b5cf48c50df54829318c8afe7cb70bcf7c`
- no rescue claim: the reported stuck harness loop was not reproduced in this session
- GitHub reports **no hosted checks** on this branch; that is unavailable evidence, not green. Exact-head local CPU, two-rank OFF/ON, negative fail-closed, restore, and independent review gates above are complete.

plotarmordev:
## Maintainer final verification — synchronized exact head `26584e5`

`f75f962` was behind current `main`, so I merged `origin/main` normally and resolved the three additive conflicts in `CHANGELOG.md`, `docker-compose.dspark.yml`, and `scripts/ci-validate.sh`. The result is mergeable and preserves both PR #53 and the newly merged spin-wait/Responses API work.

### Exact-head gates

- The final head was re-reviewed after `26584e5` became the head.
- `bash scripts/ci-validate.sh`: **PASS** on `26584e5` (CPU recipe suite, including 14 assistant-final tests and merged main gates).
- Secret scan of the PR file set: **clean**.
- Independent merge-resolution review: **APPROVE**. It verified a normal two-parent merge, byte-identical live-tested patcher/test files, one copy of each env mapping/invocation, and no lost main behavior.
- GitHub currently reports `MERGEABLE` / `CLEAN`.

### Fresh two-rank live proof on `26584e5`

Both boots used the owner-approved pinned DSpark checkpoint and GPU utilization `0.80`.

**OFF / default**
- Effective flag `0` on head and worker.
- No assistant-final patch marker on either rank.
- Rendered stock tail unchanged: bare EOS, `98` token IDs.
- API smoke passed.

**ON**
- Effective flag `1` on head and worker.
- Both ranks reported patched-and-verified; manual rerun was idempotent on both.
- Same render preserved all `98` stock token IDs and appended exactly one `<|Assistant|><think>` generation header (`100` total).
- Live continuation completed with `finish_reason=stop` and exact content `alpha beta`.
- API smoke passed.

**Restore**
- Owner `.env.dspark` and `.env.abliterated` were restored byte-for-byte (SHA-256 matched captures).
- Two-rank restore and API verification passed; cluster is serving `deepseek-v4-flash-abliterated`.

The fail-closed anchor-drift boot proof remains from code head `0864014`; the patcher/test bytes are unchanged at `26584e5`, and the exact-head CPU suite plus independent merge review revalidated that path.

There are **no hosted checks** on this branch; that evidence is unavailable, not green. Exact-head local CI, live OFF/ON, restore, scope/secret gates, and independent review are complete. **Merge-ready.**

plotarmordev:
Merged normally as `d29936a`. The merge commit tree is byte-identical to tested head `26584e5`.

**Follow-up obligation:** no hosted checks were available for this merge. If hosted CI is configured or becomes available, run it against merged `main` / `d29936a`; until then, the recorded evidence is exact-head local CI plus live two-rank OFF/ON and restore, not hosted CI.

PR #59, opened 2026-08-14
stability-quick: skip vision checks on the text-only serve profile

hecisaza:
`scripts/stability-quick.py` fails immediately on the recipe's **default** serve profile,
before any of its checks run.

### Reproduce

Default text-only deployment (`ENABLE_VL_SIDECAR` unset → 0), healthy 2× Spark cluster:

```console
$ python3 scripts/stability-quick.py
[11:55:31Z] Phase 0: readiness smoke
[11:55:31Z] FAIL readiness: <urlopen error [Errno 111] Connection refused>
```

The server is healthy; only the vision sidecar is absent — which is what the default profile
asks for:

```console
$ curl -s -o /dev/null -w '%{http_code}\n' http://127.0.0.1:8888/v1/models   # text
200
$ curl -s -o /dev/null -w '%{http_code}\n' http://127.0.0.1:8889/v1/models   # VL sidecar
000
```

### Cause

Phase 0 probes the sidecar unconditionally, and phase 3 is vision-only:

```python
main_models = request_json(f"{args.base_url}/models", timeout=30)
vl_models   = request_json(f"{args.vl_url}/models",   timeout=30)   # default :8889
```

`--vl-url` exists, but there is no way to express "there is no sidecar". The last commit to
touch this file is `2727c14` *"Ship text-only 0731 default; defer vision to experimental
path"* — the change that made text-only the default left this script requiring vision.

### Fix

Adds `--skip-vl`, defaulting to **on unless `ENABLE_VL_SIDECAR=1`**, so the default profile
works with no arguments and vision deployments are unaffected:

- phase 0 skips the sidecar `/models` probe and the vision smoke
- phase 3 records `{"skipped": true, "ok": true}` instead of failing
- text phases (readiness smoke, context ladder, soak) are untouched

18 added lines, 1 changed; the rest of the diff is re-indenting phase 3 under the new
`else:` — review with `git diff -w`.

### Verified

2× DGX Spark (GB10), TP=2 over RoCEv2, recipe `103af68`, image
`ghcr.io/anemll/dspark-vllm-gx10:0.1.1`, checkpoint 0731, text-only profile:

| case | result |
|---|---|
| unpatched, no sidecar | `FAIL readiness` at phase 0 |
| patched, default (env unset) | **PASS overall**; `vision '(skipped)'`, phase 3 skipped |
| patched, `ENABLE_VL_SIDECAR=1` | still probes VL and fails — no silent skip |
| patched, `--skip-vl` with `ENABLE_VL_SIDECAR=1` | explicit flag wins, skips |

Patched default-profile run reached every text phase:
`Phase 1 PASS 8192 (76.7 tok/s)` → `Phase 2 PASS round 1 (median 35.2 tok/s)` →
`Phase 3 SKIPPED` → `PASS overall`.

### DIFF
--- a/scripts/stability-quick.py
+++ b/scripts/stability-quick.py
@@ -6,6 +6,7 @@ from __future__ import annotations
 import argparse
 import base64
 import json
+import os
 import statistics
 import sys
 import time
@@ -155,6 +156,14 @@ def main() -> int:
     parser.add_argument("--model", default="deepseek-v4-flash-0731")
     parser.add_argument("--vl-model", default="qwen3-vl-4b")
     parser.add_argument(
+        "--skip-vl",
+        action="store_true",
+        default=os.environ.get("ENABLE_VL_SIDECAR", "0") != "1",
+        help="Skip vision-sidecar checks (readiness probe and phase 3). Defaults to "
+             "skipping unless ENABLE_VL_SIDECAR=1, so the text-only serve profile "
+             "runs out of the box.",
+    )
+    parser.add_argument(
         "--ladder",
         default="8192,32768,131072,262144",
         help="Comma-separated prompt token targets for the context ladder",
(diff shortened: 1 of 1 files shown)


MiaAI-Lab:
Merging: low-risk scripts-only change. `stability-quick.py` was failing Phase 0 on the default text-only profile because it always probed the VL sidecar. Skipping vision checks unless `ENABLE_VL_SIDECAR=1` matches the current serve default and does not touch runtime.

PR #62, opened 2026-08-14
fix: bound opt-in Responses API store

bpc-oss:
## What changed

- add an exact-source backport for the pinned Anemll vLLM `752a3a504`
- bound terminal Responses records with `DSPARK_RESPONSES_STORE_MAX_ENTRIES`
- remove matching response/message/event state together
- protect queued, in-progress, background, and foreground-streaming work
- refresh recency when `previous_response_id` is reused
- converge successful, failed, and cancelled background tasks to terminal state
- expose the native store switch in compose, defaulting it to disabled
- sync and apply the patch on both ranks, with CPU regression and compose gates

## Why

The pinned vLLM build supports `VLLM_ENABLE_RESPONSES_API_STORE=1`, but its
`response_store`, `msg_store`, and `event_store` never evict. Stateful
`previous_response_id` is therefore unsafe for a long-lived server: enabling
it can grow process memory without bound.

Upstream vLLM PR https://github.com/vllm-project/vllm/pull/48454 addresses the
same root problem for future vLLM versions but is still open and currently
conflicted. This PR is a fail-closed compatibility backport for the exact image
this repository ships; it can be retired when the base image contains an
upstream store implementation.

## Operator impact

There is no behavior change with the default:

```env
VLLM_ENABLE_RESPONSES_API_STORE=0
DSPARK_RESPONSES_STORE_MAX_ENTRIES=256
```

Operators who need stateful Responses continuation opt in by setting the first
value to `1`. State remains process-local and is lost on restart.

## Validation

- `python3 scripts/test-responses-store-hotfix.py -v` — 4 passed
- exact `fe3a48ab...` preimage transformed to `caeeb0ed...`
- transformed `serving.py` compiled and a second apply was idempotent
- enabled compose render carried store=`1`, capacity=`17`, and the mounted patch
- `bash scripts/ci-validate.sh` — passed
- `git diff --check` — passed

Before extraction, the same lifecycle policy ran on a 2x DGX Spark deployment:
260 sequential stored responses evicted the oldest response while retaining
the newest; stateful tool continuation, streaming, structured output,
reasoning, and disconnect cleanup passed with both ranks restart=0/OOM=false.

### DIFF
--- a/.env.dspark.example
+++ b/.env.dspark.example
@@ -143,6 +143,13 @@ DSPARK_ISSUE43_SCHED_DIAG=0
 # Unset/empty is not supported via compose default.
 VLLM_PREFIX_CACHE_RETENTION_INTERVAL=4096
 
+# Stateful Responses API continuation is disabled by default. Enable it when
+# clients use `store=true` plus `previous_response_id`. The pinned vLLM build's
+# native stores are otherwise unbounded; the bundled backport keeps at most
+# this many terminal responses and removes matching message/event state.
+VLLM_ENABLE_RESPONSES_API_STORE=0
+DSPARK_RESPONSES_STORE_MAX_ENTRIES=256
+
 # Keep client `stop` strings dormant until </think> (Anemll port of Tony /
 # Capicua25x Patch 5). Default on. Set 0 if a harness must bound reasoning
 # with a stop string. DSPARK_SKIP_SUPPRESS_STOPS_HOTFIX=1 skips the patch.
--- a/README.md
+++ b/README.md
@@ -469,6 +469,23 @@ This is the **Stage C padded NVFP4** path (584-byte sparse-MLA envelope via
 Optional GB10 hybrid plugin: `ENABLE_VLLM_GB10_PATCH=1 ./start-…`
 (`--quantization modelopt_gb10_hybrid`). Default off.
 
+### Optional stateful Responses API
+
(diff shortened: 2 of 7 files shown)


plotarmordev:
## Maintainer review: BLOCKED — author rework required

- This PR is draft, `CONFLICTING/DIRTY`, and far behind current `main`; automated staging stops with `needs-rebase` before any test.
- The patch currently applies while the store feature is OFF and fails open; independent review also found background-store status/cleanup defects.

Please rebase onto current `main`, resolve the conflicts, gate application on store=1 with `|| exit 1`, and add lifecycle tests. The new SHA needs fresh review before any benchmark is scheduled.

plotarmordev:
Parked as draft pending author rework. Required: rebase current main, resolve conflicts, apply only when the Responses store is enabled, fail closed on patch errors, fix background lifecycle cleanup, add lifecycle tests, and obtain fresh exact-head review.

plotarmordev:
Superseded by #205. The replacement retains @bpc-oss’s core diagnosis and exact-source backport strategy, with attribution, while rebasing onto current main and addressing the required exact-one/default-off gate, all-rank fail-closed preflight, background failure/cancellation lifecycle, continuation/eviction races, atomic rollback, and focused tests. Closing this draft avoids two competing implementation lanes; please continue review on #205. Thank you for the original investigation and patch direction.

PR #49, opened 2026-08-13
perf(dsv4): backport upstream vLLM #51967 — compile-time top-k index kernel

Saolence:
## Summary

Backport of [vllm-project/vllm#51967](https://github.com/vllm-project/vllm/pull/51967) *"[Perf][DSV4] Optimize global top-k index kernel with compile-time constants"* to the Anemll `ghcr.io/anemll/dspark-vllm-gx10:0.1.1` image (vLLM `0.25.2.dev0+g752a3a504.d20260714`).

Ships as an idempotent hotfix: `patches/hotfix-dsv4-topk-compile-time-consts.py`.

## Why (前因)

`_compute_global_topk_indices_and_lens_kernel` (in `vllm/models/deepseek_v4/common/ops/cache_utils.py`) took `global_topk_indices_stride`, `topk_indices_stride`, `topk`, `block_table_stride` and `block_size` as **runtime** arguments. Triton JIT-compiles a generic loop with dynamic bounds checks for runtime values — every prefill step pays that overhead on the global top-k pass, which is on the hot path of DSV4's sparse attention.

## The fix (后果)

Promote those five arguments to `tl.constexpr`. The JIT now knows `topk=128`, `block_size=256` (DSV4 values) at compile time, fully unrolls the loop and drops dynamic bounds checks. **Numerically identical** — no math change, only instruction-count reduction.

## Validation (2x DGX Spark, TP=2, DSV4-Flash-0731, 1M ctx)

| Metric | Baseline | With #51967 | Δ |
|---|---:|---:|---:|
| decode C1 | 35.9 tok/s | 37.3 tok/s | **+2.8%** |
| prefill 2K/8K/32K | — | — | within ±5% noise |
| sanity (chat/tool/reasoning) | pass | pass | — |

## Notes

- Follows the repo's existing upstream-backport convention (see `hotfix-dsv4-skip-topk-49486.sh`).
- Applies on both ranks; the hotfix script is idempotent (re-run skips applied hunks).

PR #67, opened 2026-08-16
perf(dsv4): backport upstream vLLM #51967 — compile-time top-k index kernel (re-new)

Saolence:
Re-submission of #49 (fork was deleted; branch restored from `refs/pull/49/head`).

## Summary

Backport of [vllm-project/vllm#51967](https://github.com/vllm-project/vllm/pull/51967) *"[Perf][DSV4] Optimize global top-k index kernel with compile-time constants"* to the Anemll `ghcr.io/anemll/dspark-vllm-gx10:0.1.1` image (vLLM `0.25.2.dev0+g752a3a504.d20260714`).

Ships as an idempotent hotfix: `patches/hotfix-dsv4-topk-compile-time-consts.py`.

## Why

`_compute_global_topk_indices_and_lens_kernel` took 5 stride/topk/block-size values as runtime args. Triton JIT emits dynamic bounds checks for runtime values; promoting them to `tl.constexpr` lets the compiler fully unroll and drop the checks. Kernel is numerically identical (constexpr promotion only).

## Measured (2026-08-17 re-test, 2x DGX Spark TP=2, DSV4-Flash-0731, thinking=max + MTP-5 + nvfp4_ds_mla)

Decode C1 (256-token prompt, 128-token output):
- standalone: 46.3 → 47.8 tok/s (**+3.3%**)
- as part of the full #51967+#49283+#49160 stack: 46.3 → 52.0 tok/s (**+12.2%**, Welch p=0.002, n=16/16)

32K/128K cold prefill: within measurement noise.

## Notes

- Upstream #51967 merged 2026-08-16 (previously unmerged — the original withdrawal reason).
- Idempotent, applies on both ranks, follows repo convention.


plotarmordev:
Not ready to merge; the implementation is technically validated, but the product value is not yet proven.

**What changed in the review**
- The earlier fork-local `+3.3%` claim was withdrawn: it had no sample count, dispersion, or reproducible artifact, and the only statistical result belonged to a three-patch stack.
- This fork's effect is therefore unmeasured. Upstream measured roughly `+0.50%` output throughput / `-0.98%` TPOT at TP=8, which does not establish a useful gain on this TP=2 deployment.
- The feature is default OFF. Merging it now adds an eight-file hotfix path and maintenance burden without changing production behavior.

**What the existing proof does show**
- Current head `c4953d22b67fd96da85bcf8caf967c489df54f63` passes exact merged-tree CPU CI.
- The pre-merge feature commit `0f07d8df787bba36f0b99230ce12fe77e6f1cef2`, tested over the same `main` now contained in the current head, applied on both ranks, passed smoke, and restored cleanly.

That proves apply/rollback correctness, not usefulness. Recommendation: do not merge unless a local A/B shows a repeatable gain above the current noise floor. Otherwise close it and rely on a future image containing upstream #51967.

plotarmordev:
Closing without merge. The fork-local performance benefit is unmeasured; the earlier +3.3% claim was withdrawn, and a default-OFF eight-file hotfix path is not justified without a repeatable local A/B gain above noise. Reopen with that evidence if the pinned image still needs the upstream change.

PR #50, opened 2026-08-13
perf(dsv4): backport upstream vLLM #49283 — split Q/KV RMSNorm blocks

Saolence:
## Summary

Backport of [vllm-project/vllm#49283](https://github.com/vllm-project/vllm/pull/49283) *"[DSv4 Perf] Right-size per-task RMSNorm blocks for the fused QK norm"* to the Anemll `ghcr.io/anemll/dspark-vllm-gx10:0.1.1` image (vLLM `0.25.2.dev0+g752a3a504.d20260714`).

Ships as an idempotent hotfix: `patches/hotfix-dsv4-qk-rmsnorm-split-blocks.py`.

## Why (前因)

`_fused_q_kv_rmsnorm_kernel` (in `vllm/models/deepseek_v4/common/ops/fused_qk_rmsnorm.py`) launched both the Q row and the KV row with a **single shared** `BLOCK_SIZE = triton.next_power_of_2(max(q_size, kv_size))`. On DeepSeek-V4 the Q row is narrower than the KV row, so the Q task ran with a wide block whose excess lanes were masked (`mask = block < SIZE`) — pure ALU waste on every attention layer of every prefill/decode step.

## The fix (后果)

Extract a reusable `_rmsnorm_row` helper and give Q and KV their own block widths (`Q_BLOCK = next_pow2(q_size)`, `KV_BLOCK = next_pow2(kv_size)`). Narrow Q rows no longer pay for KV-wide reductions. fp32 math is unchanged — **numerically identical**.

## Validation (2x DGX Spark, TP=2, DSV4-Flash-0731, 1M ctx)

| Metric | Baseline | With #49283 | Δ |
|---|---:|---:|---:|
| prefill 32K | 11704.9 tok/s | 11962.8 tok/s | **+2.2%** |
| decode C1/C4 | 36.3/20.1 | 36.9/20.2 | +1.7%/+0.5% |
| prefill 2K/8K | — | — | within noise |
| sanity | pass | pass | — |

## Notes

- Complements #51967 (different file); both can be applied together.
- Idempotent hotfix, applies on both ranks, follows repo convention.

PR #68, opened 2026-08-16
perf(dsv4): backport upstream vLLM #49283 — split Q/KV RMSNorm blocks (re-new)

Saolence:
Re-submission of #50 (fork was deleted; branch restored from `refs/pull/50/head`).

## Summary

Backport of [vllm-project/vllm#49283](https://github.com/vllm-project/vllm/pull/49283) *"[DSv4 Perf] Right-size per-task RMSNorm blocks for the fused QK norm"* to the Anemll image (vLLM `0.25.2.dev0`).

Ships as an idempotent hotfix: `patches/hotfix-dsv4-qk-rmsnorm-split-blocks.py`.

## Why

`_fused_q_kv_rmsnorm_kernel` launched Q and KV rows with one shared `BLOCK_SIZE = max(q_size, kv_size)`. On DSV4 the Q row is narrower, so Q ran with masked wide blocks — ALU waste on every attention layer. Splitting into `Q_BLOCK`/`KV_BLOCK` via a reusable `_rmsnorm_row` helper removes it. fp32 math unchanged.

## Measured (2026-08-17 re-test, 2x DGX Spark TP=2, DSV4-Flash-0731, thinking=max + MTP-5 + nvfp4_ds_mla)

Decode C1 (256-token prompt, 128-token output):
- on top of #49(single): 47.8 → 49.7 tok/s (**+4.0%**)
- as part of the full #51967+#49283+#49160 stack: 46.3 → 52.0 tok/s (**+12.2%**, Welch p=0.002)

32K/128K cold prefill: within measurement noise.

## Notes

- Upstream #49283 still open (original withdrawal reason #2 — kept for the stack).
- Original #49's measured `32K prefill +2.2%` did NOT reproduce under the current hotfix set + cold-prefill protocol; reported honestly.
- Idempotent, applies on both ranks, follows repo convention.


plotarmordev:
Not recommended for merge in its current form.

**Why the original value case does not transfer here**
- The upstream premise is inverted for this checkpoint: Q is 1024 and already unchanged; only the KV task narrows from 1024 to 512.
- Upstream's 1.19–1.34× kernel result used different dimensions on Intel B60 hardware. It is not evidence for this GB10 deployment.
- The earlier fork-local `+4.0%` and stacked `+12.2%` claims were removed because neither was attributable to this patch.
- There is no end-to-end performance measurement. Upstream #49283 remains open, and its target file was later deleted from upstream `main`.
- This would add a default-OFF, twelve-file, 1,300+ line hotfix path with no demonstrated benefit.

**What the existing proof does show**
- Current head `4e802743a0ab15118f16b62a850701aff6f398d2` passes exact merged-tree CPU CI.
- The pre-merge feature commit `ba0f5fe8876c11babf76d806330aa6a4a930d473`, tested over the same `main` now contained in the current head, applied on both ranks, passed smoke, and restored cleanly.

That proves apply/rollback correctness only. The boring recommendation is to close this PR unless a focused local benchmark first demonstrates a repeatable end-to-end gain.

plotarmordev:
Closing without merge. The upstream rationale does not match this checkpoint, the earlier performance claims were not attributable to this patch, and no end-to-end gain was measured. A default-OFF 1,300+ line hotfix path is not justified. Reopen only with a focused, repeatable local benchmark showing real value.

PR #70, opened 2026-08-16
feat(scripts): serving audit suite — RULER-lite eval, spec-decode acceptance probe, tool battery, garble sweep

TysAIs:
## What

Adds a battle-tested serving audit suite to `scripts/` (verified live 2026-08-16 on 2x DGX Spark / Anemll 0.1.1):

- **`ruler-lite.py`** — RULER-style synthetic long-context eval (S/MK-NIAH retrieval, variable tracking, common-words extraction) using canonical NVIDIA RULER templates incl. answer prefixes; tokenize-verified context lengths; CI-able exit code
- **`spec-acceptance.py`** — overall + per-position DSpark draft acceptance from /metrics
- **`tool-battery.py` + `deepctx-tool-battery.py`** — agent tool-calling incl. issue55 truncation safety, at 32k/131k+
- **`bench-miaai.py`** — throughput matrix matching the recipe's 08-14 methodology
- **`context-garble-sweep.py`** — cold-prefill garble sweep with a **corrected filler ratio** (the old 1.3 words/token under-pads ~1.2x on the DS4 tokenizer, so near-1M sweeps hit the context ceiling and scored the server's correct 400 as a false failure)
- **`run-audit.sh`** — one-shot orchestrator; **`EVAL.md`** — methodology + expected values

## Verified numbers (our cluster, 2026-08-16)

C1 decode 73-76 tok/s · acceptance ~68-75% · RULER-lite 8/8 at 8k/32k · tool battery 7/7 · garble CLEAN through 900k tokens.

## Why it matters

Naive benchmarks misread this stack badly (per-chunk SSE counting undercounts ~4.5x; pathological prompts collapse the drafter into a fake 40% regression; word-count filler over-runs the 1M ceiling). This suite encodes the correct methodology so regressions are caught before they ship.

### DIFF
--- a/AUDIT.md
+++ b/AUDIT.md
@@ -1,0 +1,46 @@
+# Serving audit suite (verified on 2x DGX Spark / Anemll 0.1.1, 2026-08-16)
+
+This fork adds a **serving audit suite** — tests that answer "is this deployment
+actually good?" beyond "does it respond?" They were designed from first
+principles after a false-alarm hunt (a naive benchmark read 45 tok/s where the
+truth was 73+; a garble sweep scored the server's *correct* context-limit 400
+as a model failure). Methodology and rationale: [`scripts/EVAL.md`](scripts/EVAL.md).
+
+## Run everything
+
+```bash
+bash scripts/run-audit.sh --base-url http://127.0.0.1:8888/v1 --model deepseek-v4-flash-0731
+```
+
+Or each phase individually:
+
+| Phase | Tool | Checks |
+|---|---|---|
+| Throughput | `scripts/bench-miaai.py` | C1/C6 decode tok/s (excl TTFT), TTFT |
+| Spec-decode health | `scripts/spec-acceptance.py` | Draft acceptance rate + per-position curve |
+| Long-context quality | `scripts/ruler-lite.py` | RULER-style retrieval / multi-hop tracing / aggregation at 8k→262k |
+| Tool calling | `scripts/tool-battery.py` | Agent tool calls incl. truncation safety (issue55) |
+| Deep-context tools | `scripts/deepctx-tool-battery.py` | Tool battery at 32k/131k+ |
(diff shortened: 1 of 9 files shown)


MiaAI-Lab:
Not merging yet — serve-safe (add-only scripts, no compose/start/hotfix), but the advertised one-shot runner is broken.

**Does not touch production.** 9 new files only. Fine as optional evals.

**Blockers for “good to merge as documented”:**

1. `scripts/run-audit.sh` documents `bash scripts/run-audit.sh --base-url … --model …` (and `AUDIT.md` uses that). The script never parses argv; only `BASE_URL`/`MODEL` env vars. Flags are silently ignored.
2. Phase 6 will fail immediately: `GARBLE_LENGTHS=2048,32768,131072` is passed as `--lengths "$GARBLE_LENGTHS"`, but `context-garble-sweep.py` uses `nargs="+"` / `type=int`, so `2048,32768,131072` is not a valid int.
3. `bench-miaai.py` uses `urllib.error.URLError` without `import urllib.error` — retry path NameErrors.
4. Garble script default `--model` is `SuperDeepseek-V4-Flash-abliterated-MQ-2xDGX`, not this recipe’s `deepseek-v4-flash-0731` (orchestrator overrides it; standalone does not).
5. `deepctx-tool-battery.py` still uses the 1.3 words/token pad that `EVAL.md` says is wrong, and the summary hardcodes `single=True` even when the call failed.

Happy to merge once (1)+(2) are fixed so `run-audit.sh` actually runs. The rest can follow.

MiaAI-Lab:
Merged with follow-up fixes on `main` (`83aa4f6`) so the advertised runner actually works:

- `run-audit.sh` now parses `--base-url` / `--model` / length flags
- garble `--lengths` accepts the comma-separated CSV from the orchestrator
- phase pass/fail uses the python exit code (not `tee`)
- `import urllib.error` in `bench-miaai.py`
- recipe default model on the garble script
- deep-context pad uses the 1.56 tok/word ratio; summary reports the real single-call result

Issue #22, opened 2026-08-09
Suspected: nvfp4_ds_mla KV degrades decode at long context (fp8_ds_mla ~16x faster at ~600K)


plotarmordev:
# Suspected: nvfp4_ds_mla KV cache degrades decode badly at long context (fp8_ds_mla ~16x faster at 600K) — 3-node Spark cluster

First: thank you for this recipe, it's been the backbone of our setup for weeks.

Reporting a **suspected** issue rather than a confirmed one — I have a large
measured difference but I have **not** run a properly controlled A/B, so please
read the caveats before spending time on it.

## Summary

On our cluster, switching `--kv-cache-dtype` from `nvfp4_ds_mla` (recipe
default) to `fp8_ds_mla` appeared to make long-context decode roughly **16x
faster**, while also giving a **slightly larger** KV pool. Short-context speed
was unchanged and matches your published number.

If that reproduces for you, the default may be costing deep-context users a lot.
If it doesn't, I'd love to know what's different about our setup.

## Environment

- Recipe at commit `232bd96` (2026-07-30)
- Image `ghcr.io/anemll/dspark-vllm-gx10:0.1.1`, vLLM `0.25.2.dev0+g752a3a504`
- **3 DGX Sparks**, model served TP=2 across two of them (the third is an ops
  node running no model). Cluster built with the NVIDIA Sync Cluster Assistant.
- Model: a **0731-GA-based NVFP4 DSpark checkpoint** (not the stock HF
  checkpoint — a third-party build of the same generation/format, so weights
  are a possible difference from your validation runs)
- `MAX_MODEL_LEN=1048576`, `MAX_NUM_SEQS=4`, `MAX_NUM_BATCHED_TOKENS=8192`,
  `GPU_MEMORY_UTILIZATION=0.80`, `MTP_NUM_TOKENS=5`

## What we measured

| Context | `nvfp4_ds_mla` | `fp8_ds_mla` |
|---|---|---|
| ~630K tokens | **~1.0–1.1 tok/s** | **17.3 tok/s** |
| ~400K tokens | not measured | 16.3 tok/s |
| ~120K tokens | not measured | 22.0 tok/s |
| short prompt, C=1 | not measured | 66.5 tok/s |

That 66.5 tok/s matches the 66.6 tok/s in your README's concurrency table
almost exactly, which is why we suspect the published figures are shallow
context and the regression only appears at depth.

[...]

PR #73, opened 2026-08-17
fix(patches): give issue22 hotfix a --status mode, and honour VLLM_ROOT

sethforprivacy:
`patches/hotfix-nvfp4-ds-mla-issue22.sh` is the only hotfix in `patches/` that doesn't implement the `"<label> : APPLIED|NOT APPLIED"` `--status` contract its siblings (`hotfix-dsv4-*.sh`) all provide. It parses no arguments at all, so `--status` is silently ignored and the script runs its normal apply path instead.

### Why it matters

**1. A status query performs a write.** On an already-patched tree the patch is a no-op, so it's harmless in practice — but `--status` shouldn't mutate.

**2. Parity tooling can't cover this hotfix.** Anything that gates hotfix parity by comparing `--status` text between ranks gets no `APPLIED`/`NOT APPLIED` token here, so a `grep -oE '(APPLIED|NOT APPLIED)'` matches nothing and the hotfix is skipped — *silently*, which is the exact failure mode such a gate exists to prevent. Under TP=2, a hotfix applied on one rank and not the other leaves the ranks running different attention code, and that doesn't necessarily crash — it can just serve subtly wrong output.

We hit this on a 2x DGX Spark deployment: our runtime gate reported the other hotfixes in `patches/` fine and quietly ignored this one.

### A second, unrelated bug in the same file

Both embedded Python heredocs hardcode `/usr/local/lib/python3.12/dist-packages/vllm`, ignoring the `VLLM_ROOT` that the shell half already respects:

```bash
VLLM_ROOT="${VLLM_ROOT:-/usr/local/lib/python3.12/dist-packages/vllm}"   # respected
...
root = Path("/usr/local/lib/python3.12/dist-packages/vllm")              # hardcoded
```

So `VLLM_ROOT=/custom bash hotfix-nvfp4-ds-mla-issue22.sh` prints the custom root in its banner while patching and verifying the *default* tree. A `--status` implementation reading a hardcoded path would be wrong for the same reason, so it's fixed here too. The siblings already do this correctly (`root = Path("$VLLM_ROOT")`).

### Changes

- add `ACTION="${1:-}"` and a `status()` function that reports state and exits without touching the tree, matching the sibling format and label padding
- handle `--status` before any write happens
- pass `$VLLM_ROOT` into the apply heredoc instead of hardcoding
- replace the bespoke `=== Verification ===` block with `bash "$0" --status`, which is what the siblings do and keeps one source of truth for the check

No change to the fix itself — the `use_fp8_cache` dispatch patch is untouched.

### Verification

Against `ghcr.io/anemll/dspark-vllm-gx10:0.1.1` and a synthetic tree:

| case | result |
|---|---|
| `--status` on pristine tree | `nvfp4_ds_mla routed to fast FP8 path : NOT APPLIED` |
| `--status` leaves file byte-identical | `cmp` clean — read-only confirmed |
| apply with `VLLM_ROOT` set | patches the tree it names |
| `--status` after apply | `... : APPLIED` |
| re-apply | idempotent, `Skipped: 1` |
| patched module imports | `imports OK` |

### DIFF
--- a/patches/hotfix-nvfp4-ds-mla-issue22.sh
+++ b/patches/hotfix-nvfp4-ds-mla-issue22.sh
@@ -9,24 +9,55 @@
 #   docker exec <container> bash /path/to/hotfix-nvfp4-ds-mla-issue22.sh
 #   # Then restart the vLLM process inside the container.
 #
+#   docker exec <container> bash /path/to/hotfix-nvfp4-ds-mla-issue22.sh --status
+#   # Report patch state only; makes no changes.
+#
 # Safe to re-run (idempotent — skips already-applied patches).
 set -euo pipefail
 
 VLLM_ROOT="${VLLM_ROOT:-/usr/local/lib/python3.12/dist-packages/vllm}"
+ACTION="${1:-}"
 
 if [ ! -d "$VLLM_ROOT" ]; then
   echo "ERROR: vLLM not found at $VLLM_ROOT" >&2
   exit 1
 fi
 
-echo "=== Hotfix: nvfp4_ds_mla long-context decode (Issue #22) ==="
-echo "vLLM root: $VLLM_ROOT"
-
-python3 <<'PYEOF'
+# Report patch state without touching the tree, in the same
+# "<label> : APPLIED|NOT APPLIED" form the other hotfixes here use, so tooling
+# can compare status text across ranks.
+status() {
+  python3 - "$VLLM_ROOT" <<'PY'
 import sys
 from pathlib import Path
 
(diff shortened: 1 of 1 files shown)


MiaAI-Lab:
Merging: tooling-only change to `patches/hotfix-nvfp4-ds-mla-issue22.sh`. Adds a read-only `--status` mode and makes the apply path honour `VLLM_ROOT` (same default as today). The Issue #22 flashmla dispatch string is unchanged, and compose still runs the script with no args plus `|| true`, so default boot/serve behavior is unchanged.

PR #76, opened 2026-08-17
fix(patches): add a read-only --status mode to the seven .py hotfixes

sethforprivacy:
The always-on `.py` hotfixes (`hotfix-encoding-dsv4-issue21.py`, `hotfix-dsv4-issue26-hybrid-swa-min.py`, `hotfix-dsv4-issue27-partial-prefill-concurrency.py`, `hotfix-dsv4-issue31-v2-thinking-budget-gpu.py`, `hotfix-dsv4-issue43-decode-fairness-and-diag.py`, `hotfix-dsv4-issue55-tool-truncation.py`, `hotfix-dsv4-suppress-stops-in-reasoning.py`) had no way to report their applied state — unlike the `.sh` hotfixes, which all expose a `<label> : APPLIED|NOT APPLIED` `--status` mode (see #73). The compose entrypoint runs them chained with `;` and no `set -e`, so a hotfix whose anchor drifts (e.g. after an image bump) prints a traceback and vLLM boots unpatched with nothing recording what happened — and any parity tooling that greps for `APPLIED|NOT APPLIED` (as the `.sh` contract documents) cannot cover these seven scripts at all.

### Changes

Each of the seven `.py` hotfixes gains a read-only `--status` mode that:

- reads the same marker(s) the apply path writes and prints `<label> : APPLIED|NOT APPLIED` (same token format as the `.sh` hotfixes),
- mutates nothing and exits 0,
- degrades gracefully when the target file is absent (`NOT APPLIED`, exit 0),
- keeps its existing positional-arg behavior (`--status` is consumed before any path argument; issue31 reports per-artifact across its 5 patched files).

### Contract rationale

Same spirit as #73: a status query is a query — it must not write, must be parseable, and must operate inside the container where the hotfixes run.

### Verification

Inside `ghcr.io/anemll/dspark-vllm-gx10:0.1.1` (throwaway containers, no GPU):

| state | result |
|---|---|
| pristine image, `--status` | `NOT APPLIED` for all seven |
| apply then `--status` | `APPLIED` for all seven (issue21 correctly reports `NOT APPLIED` on the pristine image: its encoder does not carry the broken anchor there) |
| host without the vLLM tree | clean `NOT APPLIED`, exit 0 |

### DIFF
--- a/patches/hotfix-dsv4-issue26-hybrid-swa-min.py
+++ b/patches/hotfix-dsv4-issue26-hybrid-swa-min.py
@@ -20,6 +20,7 @@ entrypoint before ``exec vllm serve``.
 """
 from __future__ import annotations
 
+import sys
 from pathlib import Path
 
 P = Path(
@@ -28,6 +29,16 @@ P = Path(
 
 MARK_V2 = "# [issue26-hotfix-v2] SWA may shrink the common hit (#36)"
 MARK_V1 = "# [issue26-hotfix] SWA groups must not shrink the hybrid common hit"
+
+if len(sys.argv) > 1 and sys.argv[1] == "--status":
+    status_src = P.read_text(encoding="utf-8") if P.is_file() else ""
+    if MARK_V2 in status_src:
+        print("issue26 hybrid-SWA-min v2         : APPLIED")
+    elif MARK_V1 in status_src:
+        print("issue26 hybrid-SWA-min v2         : NOT APPLIED (v1 inject present)")
+    else:
+        print("issue26 hybrid-SWA-min v2         : NOT APPLIED")
+    raise SystemExit(0)
 
 STOCK = (
     "                _new_hit_length = len(hit_blocks[0]) * spec.block_size\n"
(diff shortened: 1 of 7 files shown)


sethforprivacy:
Follow-up on #73: that PR gives `hotfix-nvfp4-ds-mla-issue22.sh` the `--status`/`APPLIED|NOT APPLIED` contract (and fixes its hardcoded VLLM_ROOT). This PR extends the same contract to the seven always-on `.py` hotfixes, which had no status mode at all — so parity tooling that greps `docker exec … --status` (as #73 documents) can cover the whole applied-patch surface, not just the `.sh` set.

Independent and mergeable in either order (touches only `patches/*.py`, no overlap with #73's file).

MiaAI-Lab:
Verified on the live cluster before merge.

`docker exec` `--status` on head and worker, all APPLIED:
- issue21, issue26 v2, issue27, issue43, issue55, suppress-stops
- issue31: thinking_budget_gpu.py, input_processor.py, config/vllm.py, sampler.py, model_runner.py

`scripts/ci-validate.sh` passed. Merging.

PR #77, opened 2026-08-17
fix(serve): run issue31/issue55 .py hotfixes unconditionally; warn when the encoder is missing

sethforprivacy:
In `docker-compose.dspark.yml`'s entrypoint, the issue31 (GPU-resident V2 thinking-token-budget) and issue55 (tool-call truncation safety) hotfixes were chained **inside** `if [ -n "$ENCODING_SOURCE" ] && [ -f "$ENCODING_SOURCE" ]; then … fi` — the block that locates and copies `encoding/encoding_dsv4.py`. Neither hotfix depends on the encoder.

When `ENCODING_SOURCE` cannot be resolved (e.g. `ABLITERATED=1` with the keys checkpoint, a custom model path, or a snapshot without `encoding_dsv4.py`), both correctness hotfixes were silently skipped **and** the `reasoning_effort` remap was skipped, with no warning. `scripts/ci-validate.sh` stays green because it only greps the compose file for the mount lines / invocation strings — which are present regardless of the gating.

### Changes

- `issue31` + `issue55` now run unconditionally, alongside the other `.py` hotfixes at the top level of the entrypoint.
- The encoder copy + reasoning-effort remap + issue21 stay gated on `ENCODING_SOURCE`, but now emit a loud `WARN` to stderr when the encoder cannot be resolved (previously fully silent).

### Verification

- `docker compose config` renders cleanly.
- Functional test of the entrypoint order: encoder missing → `WARN` printed, issue31+issue55 run, issue21 skipped; encoder present → issue21+issue31+issue55 all run.
- Full `scripts/ci-validate.sh` (CPU gates): pass, exit 0.


MiaAI-Lab:
Verified on a full restart of the 2x cluster before merge.

Boot logs show issue31 GPU hotfix wrote/patched all targets (`[issue31-gpu-hotfix] done`) and issue55 applied. No `WARN: encoding_dsv4.py not found` (official 0731 snapshot present). `--status` reports issue31+issue55 APPLIED on both ranks. Merging.

Issue #72, opened 2026-08-17
start-…sh cannot run after a reboot: restart: unless-stopped (#38) restores the ranks, then the "container already exists" guard exits 1


sethforprivacy:
## Summary

Since #38 added `restart: ${DSPARK_RESTART_POLICY:-unless-stopped}` to `docker-compose.dspark.yml`, dockerd restores both vLLM ranks by itself at boot. The documented start command then always aborts:

```
$ ./start-deepseek-v4-flash-dspark.sh
DSpark head container already exists for project deepseek-v4-flash. Stop it first or use PROJECT_NAME=...
$ echo $?
1
```

The cluster is genuinely **up and serving** at that point — the ranks dockerd restored form the TP=2 group and answer normally. So the launcher reports a fatal error against a healthy cluster, and the only documented way forward is `./stop-…sh` (tearing down a working cluster) plus a ~6–7 min cold start.

## Environment

- 2× DGX Spark (GB10), TP=2, RoCEv2 fabric
- recipe @ `5cdb9aa` (2026-08-13)
- image `ghcr.io/anemll/dspark-vllm-gx10:0.1.1`
- model `deepseek-ai/DeepSeek-V4-Flash-0731` @ `9e165c30e2704aec5d9d593cce3eebd58bbef1cb`
- Ubuntu 24.04, systemd 255, stock (non-rootless) docker

## Reproduce

1. Bring the cluster up normally with `./start-deepseek-v4-flash-dspark.sh`.
2. Reboot both nodes (or lose power).
3. After boot, `docker ps` on each node shows `deepseek-v4-flash-vllm-dspark-1` `Up` — dockerd restored them, and the API serves.
4. Run `./start-deepseek-v4-flash-dspark.sh` → exits 1 on the guard.

## Cause

Two individually-reasonable decisions that interact:

- `docker-compose.dspark.yml:11` — `restart: ${DSPARK_RESTART_POLICY:-unless-stopped}` (added in #38, so a crashed rank comes back)
- `start-deepseek-v4-flash-dspark.sh:514` — unconditional `exit 1` when the head container is already running (and `:523` for the worker)

Together they mean the launcher can never run cleanly after a reboot, because the restart policy guarantees the precondition the guard rejects.

## Impact

[...]

PR #78, opened 2026-08-17
fix(stop): report non-zero and warn when the worker cannot be stopped

sethforprivacy:
Every remote step in `stop-deepseek-v4-flash-dspark.sh` is `ssh "$WORKER_HOST" … || true`. When the worker is unreachable (power cut, worker reboot — the common cases) the script prints nothing out of the ordinary and **exits 0**, even though nothing on the worker was stopped.

### Why it matters

- A supervisor that runs `./stop && ./start` (e.g. a systemd `ExecStop`/`ExecStart` pair) sees success and launches a fresh start while the worker still holds a stale rank: VRAM, the NCCL master port, and a container that `restart: unless-stopped` resurrects on the worker's next boot. The fresh start then wedges or fails on that state with misleading guidance.
- `start-…sh` already hardens its reachability check (`ssh -o BatchMode=yes -o ConnectTimeout=10 … true`), but `stop` did not — and a stop script hanging on an interactive password prompt is worse than failing.
- This is the mirror image of #72: an exit code that says "stopped" when it did not.

### Changes

- Precheck worker reachability up front with `ssh -o BatchMode=yes -o ConnectTimeout=10 "$WORKER_HOST" true`; when unreachable, warn loudly and skip remote ops (local head cleanup still runs).
- Replace the bare `|| true` on remote stop steps (`force_rm` remote, VL-sidecar worker, main worker) with a per-step warning + failure counter.
- Exit `1` with an actionable message when any remote stop step failed, instead of the previous unconditional success.

### Verification

- Reachable worker, non-matching project: exit 0, nothing torn down.
- Unreachable worker (dead IP): loud `WARN`s, exit 1, local head cleanup still attempted.
- Live cluster untouched in both runs; `bash -n` and full `ci-validate.sh` pass (bash 5).

### DIFF
--- a/stop-deepseek-v4-flash-dspark.sh
+++ b/stop-deepseek-v4-flash-dspark.sh
@@ -23,6 +23,22 @@ WORKER_DIR="${WORKER_SCRIPT_DIR:-${WORKER_DIR:-$SCRIPT_DIR}}"
 WORKER_HF_CACHE="${WORKER_HF_CACHE:-${HF_CACHE:-}}"
 WORKER_VLLM_HOST_IP="${WORKER_VLLM_HOST_IP:-}"
 
+# A stop that cannot reach the worker must not report success: a powered-down
+# worker resurrects its stale rank (compose restart: unless-stopped) the next
+# time it boots, and nothing here will have stopped it. Mirror start's ssh
+# hardening (BatchMode/ConnectTimeout) so stop never hangs on a prompt either.
+STOP_FAILURES=0
+stop_warn() {
+  echo "WARN: $*" >&2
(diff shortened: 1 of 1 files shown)


MiaAI-Lab:
Verified worker reachability on the live cluster (did not take the serving ranks down for a negative SSH test).

`ssh -o BatchMode=yes -o ConnectTimeout=10 zurih@<worker> true` succeeds, which is the new stop precheck. `scripts/ci-validate.sh` includes the stop-script syntax gate. Merging; a dead-worker exit-1 path remains the intended contract when SSH fails.

PR #83, opened 2026-08-18
fix(autotune): cover production dense budgets for sparse-MLA DSv4 decode

fche:
The fork's _decode_dsv4_tuning_config() buckets only the token axis; the dense/extra_topk axis is matched exactly (not bucketed) against the flashinfer autotune cache. Boot tuning only ever physically runs dense budgets {0,128,512}, so every production shape (dense 256/1024/2048/8192) is a cache miss and falls back to the heuristic SparseMlaDecodeV3Runner, measured at ~+82% decode cost on the LB-1 (DGX-GB10) in our deployment.

This hotpatch (house patches/ style):
- merges 24 production-dense configs (T axis [1..64], dense 256/1024/2048/8192) into the cache at every boot via a rewrite of write_flashinfer_autotune_cache that concatenates the seed supplement instead of replacing the file
- wires it into docker-compose entrypoint (skip-flag: DSPARK_SKIP_DSV4_AUTOTUNE_DENSE_HOTFIX) and start.sh worker sync (2-node)

The supplement is generated by the real autotuner, so its extras fingerprint matches the deployed model build; users with a different build can regenerate via the README snippet in the hotfix.

Operational fix; the durable upstream fix is to bucket the dense axis in flashinfer's _decode_dsv4_tuning_config() so the boot pass itself covers it.

P.S. This patch was built by Hermes Agent, running under this very VLLM, having noticed the autotune WARNINGs in the docker logs.


plotarmordev:
The corrected implementation is now in #88. Your original commit and supplemental tuning data are preserved there byte-for-byte. The replacement was necessary because this fork rejected the maintainer fast-forward push despite GitHub reporting maintainer edits enabled. Thank you for the production-dense measurements and initial implementation.

PR #85, opened 2026-08-19
fix(ruler-lite): long-context gate never reached the requested depth, and >=500k could not complete

newjordan:
Two independent defects in `scripts/ruler-lite.py`. Both cause the long-context gate to report something other than what it measured, in opposite directions.

Found while auditing a 2× DGX Spark deployment (DeepSeek-V4-Flash-0731 @ `9e165c30` weights, `ghcr.io/anemll/dspark-vllm-gx10:0.1.1` @ `sha256:a839484…`, TP=2, 1M ctx).

---

### 1. `pad_to_length()` capped every context at ~5.6k and still reported PASS at the requested depth

It appended one `HAYSTACK_SENTENCE` per `/tokenize` round trip under a fixed `guard < 200`. That sentence is ~25 tokens, so the achievable context was ~`200*25 + base` ≈ 5.6k. Any `--lengths` value ≥ 8192 was evaluated at that ceiling — and printed as a PASS *at the depth requested*.

Before the change, `--lengths 8192,32768`:

```
ctx=4840  4868  5396  5631      <- the 8192 arm
ctx=4840  4869  5396  5631      <- the 32768 arm, identical
=== RULER-lite: 8/8 passed ===
```

Byte-identical context sizes across a 4× depth difference. The consequence is that **`AUDIT.md`'s expected value "RULER-lite: 8/8 at 8k/32k" has not been exercising long context**, and neither has any CI invocation of this script.

Fix pads in deficit-sized chunks (measuring the padding unit once) and **raises `RuntimeError` rather than returning a short prompt**, so a depth that cannot be reached is visible instead of scored as a pass.

### 2. The hardcoded 900 s client timeout is shorter than a single deep prefill

`request_json()` pinned `timeout=900`. A cold prefill near the 1M ceiling exceeds that by itself, so the deepest lengths this script advertises could never complete:

```
FAIL task_sniah    ctx=? (0s)  pred='timed out'   (x4)
```

Server-side, every `/tokenize` returned 200 OK and **no chat completion ever reached the engine** — the client hung up mid-prefill and the harness scored its own limit as a model failure.

Adds `--request-timeout`, default **unchanged at 900 s** so existing behaviour and CI are untouched.

---

### Verification after both fixes

Same deployment, real tokenize-verified depths:

| depth | result | time/case |
|---|---|---|
| 8,199–8,204 | 4/4 | 2–10 s |
| 32,775–32,781 | 4/4 | 16–19 s |
| 131,079–131,084 | 4/4 | ~80 s |
| 262,144–262,167 | 4/4 | 165–184 s |
| 524,295–524,300 | 4/4 | 463–482 s |
| 900,015–900,020 | 4/4 (`--request-timeout 3000`) | 998–1004 s |

The 900k timings are why 900 s could never work: every case needs ~1000 s. Anything at or above ~500k needs the timeout raised.

Scope is one file; no behaviour change at default settings other than actually reaching the requested depth.


plotarmordev:
Ready to merge.

**What this PR does**
- Fixes RULER-lite padding so requested long-context depths are actually reached instead of silently stopping near 5k tokens.
- Adds a validated request-timeout override so very long prefills are not mislabeled as model failures.

**No-regression proof**
- Exact head `23a752a3397d61c90061bc470b9de94fd55a02dc` is clean and mergeable against current `main`.
- Nightshift exact-head gates passed: shell syntax, Python compile, session premises, full CPU recipe CI, and denylist.
- A fresh local merge with current `main` passed `git diff --check` and the complete CPU recipe CI, including the RULER-lite padding/timeout regression tests.
- Scope is evaluation tooling (`ruler-lite`, audit invocation, tests, changelog); no model-serving or GPU runtime path changes.

PR #88, opened 2026-08-19
fix(autotune): cover production dense budgets for sparse-MLA DSv4 decode

plotarmordev:
Supersedes #83 because the contributor fork rejected the maintainer fast-forward push.

Preserves Frank Ch. Eigler's original commit and supplemental tuning data byte-for-byte, then fixes the release-blocking integration defects found in review:
- hooks the actual DSv4 sparse-MLA cache path before broadcast
- injects the skip flag and adds a repo-wide inert-flag gate
- validates metadata compatibility and pinned overlay anchors
- preserves both-rank cache identity and refuses half-pair deployment
- adds focused behavioral tests and honest CPU/static evidence wording

Local exact-head evidence at 8d29da1d:
- 31/31 focused tests pass
- scripts/ci-validate.sh passes with 56 checks
- independent Opus and HELP/Kimi K3 reviews approve
- GPU_MEMORY_UTILIZATION is unchanged

Nightshift GPU A/B is mandatory before any performance claim, not before merge.


plotarmordev:
Still relevant, but not ready to merge.

**Why it remains relevant**
- Current-main live 2-node boot logs still show `No tuned config covers sparse_mla_sm120_decode_dsv4` for production dense-8192 shapes at 24, 12, and 6 tokens on both ranks, followed by tactic `-1` fallback.
- Main's #27 partial-prefill change does not add these autotune buckets or this cache merge hook.

**Why it must stay draft**
- The hotfix is effectively default ON: `DSPARK_SKIP_DSV4_AUTOTUNE_DENSE_HOTFIX` defaults to `0`.
- The compose entrypoint runs the patcher with `|| true`, so an application/anchor failure does not fail the boot. It may log and retain stock behavior, but the launch itself does not enforce that the intended tuning data was installed.
- CPU tests prove the merge logic, metadata checks, anchors, and fallback branches. They cannot prove that these tactics improve—or even avoid regressing—the pinned GB10 runtime.

**Current repair state**
- Current head `a6126e86185c102b15db23a18c4e6c94e9d5f32d` is clean against `main` and passes the 31-test focused suite, the current #27 regression tests, complete CPU recipe CI, and added-content secret scan.

**Required before merge**
- Run generic live T2 plus a dedicated skip-flag A/B on the exact head.
- Require a repeatable end-to-end gain and clean both-rank cache installation. If that evidence does not appear, close it. If it remains unproven but useful for experiments, make it explicit opt-in/default OFF rather than changing every boot.

plotarmordev:
Parked as draft. Do not merge until the exact head passes generic live T2 plus a dedicated skip-flag A/B, with repeatable end-to-end gain and clean both-rank cache installation. If it remains unproven, make it default OFF or close it.

fche:
I don't have an explanation why you're seeing what you're seeing.  Here, on both ranks, with what I believe is exactly the proposed patch applied, we're seeing signs that the #83 code was effective (#88 is not here), and no "No tuned ..." warnings.

<code>
11:18:24 [flashinfer_sparse_mla_warmup.py:124] Autotuning ... with cache: .../5012efb6.../autotune_configs.json
11:18:37 [Autotuner]: Loaded 48 configs from .../5012efb6.../autotune_configs.json
11:18:37 [Autotuner]: Config cache hit for sparse_mla_sm120_decode_dsv4 (runner=..., source=config file)
11:18:44 cache loaded on rank 0 from .../5012efb6.../autotune_configs.json
…Capturing CUDA graphs (PIECEWISE) 7/7, (FULL) 4/4  [zero warnings]
</code>

Those "48 configs ... autotune_configs.json" are an indication that the json config bundles were used (and effective).

fche:
A local AI went on a quest to understand why we're seeing something different here than you're seeing: it seems to come down to a difference in base image versions:

"One data point that may matter for where the A/B hook lands: the image our cluster currently runs (built 2026-07-14) does DSv4 sparse-MLA decode autotune in flashinfer_sparse_mla_warmup.py, which calls write_flashinfer_autotune_cache() (line 177) before AutoTuner.get().load_configs() — i.e., the function #83 patched is on the DSv4 path in that image. The overlay's newer kernel_warmup.py doesn't include that module, which may explain the difference between the #83 "silent no-op" you saw in review and the live behavior here. If the A/B is run on that existing image, the new .py hook won't apply (no anchor match); if it's run on a fresh build, the target code differs from what's deployed."

"Before the live T2/A/B, I'd suggest one setup check: confirm the patch's anchor actually applies to the kernel_warmup.py the A/B image ships with — the running production image (built 2026-07-14) does DSv4 autotune in flashinfer_sparse_mla_warmup.py, which calls write_flashinfer_autotune_cache() at line 177, so the hook location differs by build. Verifying that first means the A/B either demonstrates the dense-merge path or fails cleanly for a known reason instead of silently no-opping."

plotarmordev:
Closing after the planned exact-head comparison.

Tested current main `be07fcd9`, the PR #83 mechanism, and corrected PR #88 head `ecb8223f` on the pinned Anemll 0.1.1 image, with stock bookends and one frozen stock cache restored before every arm. Both mechanisms installed the same byte-identical 48-entry DSv4 cache on both ranks and removed the covered fallback warnings.

Neither produced the predeclared useful end-to-end effect at concurrency 4. Median aggregate throughput was 183.29 / 184.28 tok/s for the stock bookends, 182.81 for PR #83, and 188.56 for PR #88. Against pooled stock, the ratio of medians was -0.80% for PR #83 (independent-bootstrap 95% interval 0.942–1.050) and +2.32% for PR #88 (0.971–1.073), below the 5% useful-effect threshold. PR #88 versus PR #83 was +3.15% (0.981–1.075); their caches and median performance met the predeclared equivalence rule. All 624 measured requests completed, candidate logs had no critical errors, and the deployment was restored and verified.

The corrected PR #88 implementation is safer than PR #83 because it is exact-image-pinned, opt-in, and fail-closed. But the served-throughput benefit is not reproduced on current main, so carrying the runtime patch would add complexity without demonstrated value. Closing without merge.

fche:
Thanks for testing!

plotarmordev:
Follow-up: separating “the cache mechanism works” from “the complete server gets materially faster.”

First, an acknowledgment and correction: @fche was right that #83 is active on our deployed Anemll 0.1.1 image. Our early review mixed that image’s `flashinfer_sparse_mla_warmup.py` path with a newer overlay layout and therefore questioned the wrong hook. We corrected the setup and tested #83 directly. It loaded 48 configs on both ranks, removed the covered `No tuned config` warnings, preserved the stock entries, and produced the same final DSv4 key/value map as corrected #88. So #83 is not a silent no-op on this image.

What we tested:

- pinned current main `be07fcd9`
- pinned Anemll 0.1.1 image and the deployed `flashinfer_sparse_mla_warmup.py` path
- order: stock A1 → #83 mechanism → corrected #88 head `ecb8223f` → stock A2
- one identical frozen 24-entry stock cache restored before every arm
- 12 measured repeats at concurrency 1, 2, 4, and 6
- 277 actual prompt tokens + 256 generated tokens per request, temperature 0, ignore-EOS
- 624 successful measured requests total
- both-rank cache capture and log checks after every arm

Median aggregate served throughput:

| concurrency | pooled stock tok/s | #83 tok/s (change) | #88 tok/s (change) |
|---:|---:|---:|---:|
| 1 | 78.81 | 78.73 (-0.10%) | 78.47 (-0.43%) |
| 2 | 120.08 | 120.75 (+0.56%) | 118.47 (-1.34%) |
| 4 | 184.28 | 182.81 (-0.80%) | 188.56 (+2.32%) |
| 6 | 158.33 | 157.60 (-0.46%) | 158.00 (-0.21%) |

At the primary concurrency 4, the independent-bootstrap 95% ratio interval versus pooled stock was `0.942–1.050` for #83 and `0.971–1.073` for #88. #88 versus #83 was +3.15%, interval `0.981–1.075`. Stock bookend drift was 0.54%. Candidate logs had zero critical errors and zero covered-shape fallback warnings.

This does not refute the ~82% kernel-level decode-cost measurement in #83’s description. We reproduced the creator’s cache-installation and warning results on our deployed image; we did not rerun the private LB-1 kernel microbenchmark behind that number or claim to have reproduced the creator’s separate build. The result here is narrower: under this end-to-end serving workload on current main, the kernel-level advantage did not become a clear or material served-throughput gain.

Why neither implementation looks ideal as the durable fix:

- #83 is a valid operational hotpatch, but rewrites an installed cache writer at boot and injects build/device-specific seed data.
- #88 delivers the same data through the exact deployed DSv4 path and makes activation opt-in, pinned, and fail-closed. That is safer packaging, not a better tuning algorithm, and it adds more image-specific runtime machinery.
- Both remain coupled to image internals. The image-path difference @fche identified is exactly the maintenance risk.

@fche: do these end-to-end results fit your expectation, or does this request shape miss the condition where your kernel result should materially affect served throughput? If it misses, could you point us to the exact reproducible request or benchmark shape you consider decisive? Also, given that both mechanisms install the same cache but show no clear served gain here, would you still recommend the runtime seed, or would you prefer moving the fix into the tuner/image itself?

Maintainer decision: based on the current evidence, we are keeping #88 closed. We think the actual upgrade belongs at the tuning layer: either (a) change and validate the dense-axis bucketing/matching in `_decode_dsv4_tuning_config()`, as #83’s description suggests, or (b) tune/package the required production-dense shapes while building the exact image. Either route should include kernel-level coverage tests for the production axes and then the same end-to-end A/B after rebuilding the image. That would remove the boot-time source rewrite and supplement injection rather than making those patches permanent.

Nothing was merged. We will reconsider the technical direction if additional exact, reproducible workload evidence changes the result.

fche:
Thanks for your detailed explanation, we concur.  We also submitted the cache autotune tweaks to flashinfer upstream for their consideration, where the threshold for improvement may be lower than the hotpatch-justifying 5% here.

de1tydev:
Closing data point from a third cluster, no action requested — this only corroborates what @fche reported and what the retarget already fixed.

We ran an earlier head of this PR (the one anchored at `model_executor/warmup/kernel_warmup.py`) on a stock 2× DGX Spark GB10 deployment: pinned Anemll `0.1.1`, official `DeepSeek-V4-Flash-0731`, from checkout `f752cd0`. It reported **`prereq-missing` at every boot** and never touched the image — i.e. the fail-closed path behaved exactly as designed, and the cause was the hook location rather than anything about our deployment. That matches @fche's diagnosis and the `20037b3` retarget to `flashinfer_sparse_mla_warmup.py`.

One repo-side note in case a future patch anchors to overlay sources: `recipe/overlay/vllm/model_executor/warmup/kernel_warmup.py` is only consumed by `recipe/Dockerfile.dspark-runtime-overlay`, i.e. the Stage-C image from `build-dspark-vllm-runtime.sh`. `docker-compose.dspark.yml` does not mount it, so on the default Anemll lane the container always runs the image's own warmup module. Worth one line wherever such a patch gets documented, so the next person reads a `prereq-missing` as "wrong lane" rather than "broken patch".

Also agreeing with the close: an exact-image-pinned, opt-in, fail-closed patch that does not move served throughput is complexity without a payoff, and upstreaming the tuning data to flashinfer is the better home for it.

Issue #66, opened 2026-08-15
issue31 GPU thinking-budget hotfix: ~4.5 tok/s decode cliff recurs on the omit-field path (2x DGX Spark, TP=2)


cwschroeder:
## Summary

On 2× DGX Spark TP=2 at `f752cd04`, enabling `patches/hotfix-dsv4-issue31-v2-thinking-budget-gpu.py` reproduces a decode cliff in the **omit-field** path — requests that never send `thinking_token_budget`, with `DEFAULT_THINKING=off`.

The magnitude matches the cliff the CHANGELOG reports as resolved:

> The withdrawn implementation's ~4.5 tok/s cliff did not recur.

It does recur here. Under 4-way concurrency **every** stream drops to 1.7–2.5 tok/s, and one single-stream cell fell to 5.11 tok/s.

The effect is **intermittent, not uniform** — in the same boot, 2k and 78k cells were normal while the 25k cell collapsed. That is probably why the #48 A/B missed it.

## Why #48's verification did not catch this

The closing note reports `prompt=256, max_tokens=4096, budget=1024` — a *budgeted* request — and concludes "Omit-field path unchanged (no server-side default injected)". The omit-field path appears not to have been measured for throughput. That is the path that regresses.

## Controlled A/B

Same boot-to-boot procedure, `LONG_PREFILL_TOKEN_THRESHOLD=0` held constant, warm autotune/compile caches on both runs, containers force-recreated between runs. The **only** variable is the issue31 mount, redirected to a no-op script via `DSPARK_ISSUE31_GPU_HOTFIX`.

| Metric (C1 unless noted) | #31 disabled | #31 enabled |
|---|---|---|
| 2k decode | 46.72 tok/s | 47.63 |
| **25k decode** | **41.24** | **5.11** |
| 78k decode | 41.01 | 42.08 |
| 25k prefill | 1151.7 | 770.6 |
| **4× parallel, wall** | **29.12 s** | **63.3 s** |
| **4× parallel, aggregate decode** | **13.19 tok/s** | **6.07** |
| **4× parallel, per stream** | **6.03 / 23.99 / 4.31 / 10.33** | **2.12 / 1.89 / 2.46 / 1.69** |

[...]

PR #89, opened 2026-08-19
feat(auth): multi-key API auth via DSPARK_API_KEYS (default off)

sethforprivacy:
## Problem

The recipe has no way to require an API key, so the only documented option is auth at the reverse proxy. Basic auth is a poor fit there: OpenAI-family SDKs send `Authorization: Bearer <api_key>` and offer no way to send Basic instead, so clients that expose only an `apiKey` field (agent harnesses, most OpenAI-compatible tools) cannot authenticate at all, and the ones that can need per-request header overrides.

## Why vLLM can do this itself

- `--api-key` is a dataclass `list[str]` field, so **one flag carries N keys**.
- `AuthenticationMiddleware` compares a SHA-256 of the presented Bearer token against each configured key with `secrets.compare_digest` (constant time).
- It guards `GUARDED_PREFIX = ("/v1", "/v2", "/inference")` only, so `/health`, `/metrics`, `/version` and `/ping` stay open, and `/invocations` is an unguarded inference route — network-level control of the server port is still required.

Auth at vLLM rather than at the edge also covers LAN and tailnet callers, not just a public route.

## Change

`VLLM_API_KEY` is the existing single-key option (vLLM's native env var); `DSPARK_API_KEYS` (space-separated, single-line, **empty by default**) adds multiple independently revocable keys:

- The two variables are **mutually exclusive**: if both are set, the container entrypoint and the `start-`/`smoke-`/`status-` probe scripts exit 2 naming both variables — before any encoder copy, hotfix apply, or install work. Nothing is silently chosen or combined.
- Compose passes `DSPARK_API_KEYS` through and expands it into `--api-key` **only when the value parse correctly** (unset/empty/whitespace-only means no auth; separators trimmed/collapsed preserving order and duplicates; newline/CR and `-`-leading tokens rejected with exit 2).
- `start-`, `smoke-` and `status-deepseek-v4-flash-dspark.sh` fall back to the **first parsed key** for their probes (byte-identical parse in all four contexts).
- When keys are set, `patches/hotfix-vllm-redact-api-key-log.sh` keeps them out of the startup Docker log: the `non-default args` line shows `'api_key': ['<redacted:N value(s)>']` (the single-flag count check still works). Keys remain readable in container argv/env (`docker inspect` / host `ps`) — that is by design and documented in both docs.

## Tests

`scripts/test-dspark-api-keys.py` (CPU only, wired into `scripts/ci-validate.sh`; stdlib-only, 34 tests) **executes the real code it gates** rather than matching source text:

- unset/empty/whitespace-only emits **no** `--api-key` flag — the default-off guarantee;
- a parsed value becomes **exactly one** `--api-key` carrying every key, order preserved, duplicates allowed (repeating the flag *overwrites* rather than appends, so a per-key loop would silently leave only the last key valid);
- literal glob characters and metacharacters stay data;
- newline / CR / `-`-leading tokens are rejected with exit 2 (the single-line guard runs before tokenization in all four contexts);
- both variables set fails closed with exit 2 naming both — in the entrypoint (before any patch work) and in all three probes;
- `VLLM_API_KEY` alone keeps the single-key behavior unchanged;
- a **real-compose handoff layer** (when `docker compose` can render) passes synthetic keys through `.env` → `environment:` → rendered `command` → the real entrypoint block → asserted `exec vllm serve` argv — proving the passthrough chain, not just the embedded shell. Negative controls prove that dropping the `environment:` passthrough or regressing `$$` to `$` makes the gate fail;
- hostile ambient-parent-env tests (parent carrying both variables cannot change any outcome);
- committed-key scan covers compose, env example, probes, `docs/ENVS.md`, `CHANGELOG.md`, and the test file itself.

`bash scripts/ci-validate.sh` passes end to end (34/34 on docker-less hosts; the compose-handoff layer skips cleanly when `docker compose` cannot render).

## Live verification

On the 2× DGX Spark TP=2 clusters serving `deepseek-v4-flash-0731` (setup details in the dgx-sparks deployment notes), restarted onto this branch via the recipe's own stop/start:

- every configured key → 200 on `/v1`; keyed chat → 200
- no `Authorization` header → 401; wrong key → 401; Basic auth → 401
- `VLLM_API_KEY` alone: restarted with it set → 200 with that key, 401 with a different key, 401 keyless (then restored to the multi-key state, re-verified 4/4 → 200)
- both variables → the probes exit 2 immediately naming both (the container entrypoint runs the same check before any patch work)
- the launcher's health poll, smoke, and status probes pass against the keyed server
- `/invocations` inspected on the pinned image: unauthenticated request reaches the handler (400 validation, not 401) — it is unguarded; `/ping` and `/metrics` keyless → 200
- startup logs inspected: the `non-default args` line shows `'api_key': ['<redacted:4 value(s)>']`, no key bytes anywhere in Docker logs

## Scope note

vLLM does not attribute requests to a key, so this provides revocation and blast-radius control, **not** per-user logging.


plotarmordev:
Parked as draft pending a product and security decision. The repository already supports single-key Bearer auth through VLLM_API_KEY; this PR is justified only if per-caller revocation is actually needed. Before review: rebase current main and define/test behavior when VLLM_API_KEY and DSPARK_API_KEYS are both set, because probes currently prefer the former while compose also passes the latter through --api-key.

plotarmordev:
Thanks — I agree that independently revocable API keys are useful, so I want to continue with this feature. Please keep the PR in draft while revising it.

Required contract and acceptance criteria:

1. Rebase onto current `main` and preserve all newer CI entries.
2. Keep `VLLM_API_KEY` as the existing single-key option and `DSPARK_API_KEYS` as the multi-key option. If both are non-empty, the server entrypoint and all three probe scripts must name both variables and exit 2. Do not silently choose one or combine them.
3. Fix Compose expansion: container-side references must use `$$`. Do not interpolate key text into shell source.
4. Build `--api-key` arguments with a Bash array and expand it as `"${API_KEY_ARGS[@]}"`; do not use an unquoted argument string.
5. Parse consistently: unset, empty, or whitespace-only means no auth; trim/collapse separators; preserve order; duplicates may remain; reject tokens beginning with `-`; probes use the first parsed key.
6. Correct the docs: the repository already supports one key via `VLLM_API_KEY`; this PR adds multiple independently revocable keys. Document the exact protected and unprotected endpoints, verify `/invocations` on the pinned image, describe argv/environment exposure and stop/start rotation, and state that this provides revocation rather than per-key attribution.
7. Replace source-text confidence with behavioral coverage for unset/empty/whitespace-only, one/many keys, leading/repeated whitespace, literal glob characters, a `-`-leading token, both variables, either variable alone, exactly one `--api-key` flag, probe first-key behavior, and negative controls for `$` versus `$$` and repeated flags.
8. Re-run the live matrix after the fixes: each valid key → 200; missing/wrong key → 401; `VLLM_API_KEY` alone remains compatible; both variables → immediate exit 2; health/startup probes succeed; inspect `/invocations` and startup logs.

Please keep the implementation focused: no auth framework, user database, roles, attribution system, hot reload, reverse-proxy configuration, new dependency, or generic shell library.

sethforprivacy:
Rebased onto current `main` (`58c84eb`) and reworked per the contract. The branch now contains **only** the 9 files (it previously carried ~30 stacked commits from other PRs); all newer CI entries in `scripts/ci-validate.sh` are preserved. One commit, `cd1bc30`.

## Item-by-item

**1 — Rebase.** Done. Base is `58c84eb` (MiaAI `main`), `ci-validate.sh` keeps every newer entry (spec-acceptance, ruler-lite, inflight-cap, …); only the new test file is added.

**2 — Mutual exclusion (both non-empty → name both + exit 2).** The compose entrypoint and all three probe scripts now run the same check: if `VLLM_API_KEY` and `DSPARK_API_KEYS` are both meaningful, they print `error: VLLM_API_KEY and DSPARK_API_KEYS are both set; set exactly one of them` and `exit 2`. Nothing is silently chosen or combined — important because vLLM 0.25.1 evaluates `args.api_key or [envs.VLLM_API_KEY]`, i.e. `--api-key` silently wins when both exist.

**3 — `$$` (no host interpolation of key text).** Every reference to the key variables inside the compose `command` uses `$${...}`, including the bash locals (`$${_dspark_keys_set}`, `$${_dspark_keys[@]}`, `$${_dspark_key}`), since Compose interpolates `$VAR` as well as `${VAR}`. Verified on the live head: `docker compose config` output contains **no key text in the shell command**; keys appear only in the container `environment` surface.

**4 — Bash array, quoted expansion.** The entrypoint builds `API_KEY_ARGS=(--api-key "${_dspark_keys[@]}")` and the `exec vllm serve` line expands `"${API_KEY_ARGS[@]}"`. Live proof on the restarted cluster: the running vLLM process cmdline is exactly one `--api-key` followed by the four configured keys.

**5 — Consistent parsing.** The three probe blocks are byte-identical (asserted by a test). Semantics in all four contexts: unset/empty/whitespace-only ⇒ no auth; separators trimmed/collapsed; order preserved; duplicates allowed; a token starting with `-` ⇒ `exit 2` naming the token; probes send `Authorization: Bearer <first parsed key>`.

**6 — Docs.** `.env.dspark.example` + `docs/ENVS.md` (+ CHANGELOG) now state: `VLLM_API_KEY` is the existing single-key option (vLLM-native env); `DSPARK_API_KEYS` adds many independently revocable keys; both-set failure mode; exact endpoint coverage; argv/env exposure; stop/start rotation (no hot reload); revocation, not per-key attribution. **`/invocations` verified on the pinned image (Anemll 0.1.1 / vLLM 0.25.1):** it is registered by the SageMaker router outside `GUARDED_PREFIX = ("/v1", "/v2", "/inference")` and is **not** authenticated (same class as `/health`, `/metrics`, `/ping`) — confirmed live (see 8).

**7 — Behavioral tests.** `scripts/test-dspark-api-keys.py` (stdlib-only, 24 tests) extracts and **runs** the real entrypoint block and each real probe block through a shell. Coverage: unset/empty/whitespace-only; one/many keys; leading/repeated whitespace; duplicates; literal glob characters (with glob-matching files in cwd); `-`-leading token rejection; both variables → exit 2 naming both; either variable alone; exactly one `--api-key` flag (full argv integration against the real exec tail); probe first-key behavior; identical probe blocks; negative controls for `$` vs `$$` and for repeated flags (mutated variants are provably caught). Wired into `scripts/ci-validate.sh`; full suite passes in a clean `debian:bookworm-slim` container under bash 5 + stdlib Python (no new dependency).

**8 — Live matrix re-run.** Cluster (2× DGX Spark, TP=2, `deepseek-v4-flash-0731`) restarted onto the new code via the recipe's own stop/start:

- each valid key → **200** on `/v1/models` (4/4), keyed chat → 200;
- no `Authorization` header → **401**; wrong key → **401**; Basic auth → **401**;
- `VLLM_API_KEY` alone: cluster restarted with it set → 200 with that key, 401 with a different key, 401 keyless (then restored to the multi-key state and re-verified 4/4 → 200, keyless/wrong → 401);
- both variables: probe scripts exit **2 immediately** naming both (entrypoint runs the same block — unit + rendered-command verified; no container was boot-looped to prove a negative);
- health/startup probes: the launcher's health poll and minimal chat succeeded with keys set; `status-` and `smoke-` probes pass against the keyed server;
- `/invocations` inspected on the pinned image: unauthenticated request reaches the handler (**400** validation error, not 401) — unguarded; `/ping` and `/metrics` also keyless 200;
- startup logs inspected: the `non-default args` line shows `'api_key': ['<redacted:4 value(s)>']` (the deployment's local log-redaction hotfix applied), no key material in logs.

Scope stayed tight: no auth framework, user database, roles, attribution system, hot reload, reverse-proxy config, new dependency, or generic shell library.

Deployment note: the reviewed files are synced to the recipe checkout on both live clusters; cluster B has been verified after a full stop/start, cluster A picks the same files up at its next normal rotation. The deployment keeps its local `hotfix-vllm-redact-api-key-log.sh` layered on the recipe (out of PR scope).

This was run through a 4-model quorum review before pushing; that panel returned no findings.

plotarmordev:
Thanks @sethforprivacy — the feature is genuinely useful, and the implementation is much stronger after your revision.

The core shape is good:

- existing `VLLM_API_KEY` single-key behavior is preserved;
- multi-key auth remains opt-in/default-empty;
- one quoted Bash array produces exactly one `--api-key` flag;
- key order and duplicates are preserved;
- globs/metacharacters stay data rather than executing;
- dash-leading tokens fail closed;
- all three probes share the same parsing behavior and use the first parsed key;
- Compose’s container-side references use `$$` correctly;
- the current-main merge preview passes full CPU CI;
- the normal real-Compose handoff and argument mutation controls pass.

I think this feature is worth finishing. One consolidated revision is still needed before the final live auth matrix:

### 1. Rebase current `main`

The current head conflicts in:

- `CHANGELOG.md`
- `docs/ENVS.md`
- `scripts/ci-validate.sh`

These are additive unions. Please preserve all current-main entries—including the env-normalization, fabric, sampler, and CI gates—plus this PR’s auth rows/tests.

### 2. Decide and handle startup-log key exposure

Your live evidence reports clean startup logs because the deployment had a private `hotfix-vllm-redact-api-key-log.sh` layered outside this PR.

On the stock pinned image, the non-default-arguments startup line can print the configured `api_key` list into Docker’s startup/retained logs.

Recommended: include an equivalent focused repository patch that replaces the values with something like:

```text
api_key: ["<redacted:4 keys>"]
```

A key count is useful; key bytes must not be logged.

If you prefer not to include the redaction patch, please say so explicitly. In that case **@MiaAI-Lab**, as repository owner, needs to decide whether the project accepts plaintext API keys being visible in process arguments, environment inspection, startup logs, and retained Docker logs. If accepted, that exposure must be documented prominently rather than inferred from private deployment behavior.

### 3. Reject conflicting auth variables before side effects

In the Compose entrypoint, the mutual-exclusion check currently runs after encoder copying, hotfix application, optional installs, and other setup.

Please move the auth validation/array-build block immediately after the initial exports and before those operations.

Required both-set behavior:

```text
first error: names VLLM_API_KEY and DSPARK_API_KEYS
exit code: 2
patch/install work before rejection: none
```

### 4. Reject multiline/CR key lists

`read -r -a` consumes only the first line. Values such as:

```text
key1
key2
```

silently drop later keys; a later `-bad` token can also escape validation.

The smallest clear contract is to keep keys single-line and reject `\n`/`\r` with exit 2 in the entrypoint and all three probe blocks. Add newline/CRLF/dash-after-newline cases.

### 5. Make the tests hermetic

`scripts/test-dspark-api-keys.py` inherits the caller’s real `VLLM_API_KEY` / `DSPARK_API_KEYS`.

In the current merge preview:

- ambient `DSPARK_API_KEYS` causes 5 failures;
- ambient `VLLM_API_KEY` causes 11 failures and 1 error.

Clear both variables before every case and add a hostile-parent regression run.

### 6. Exercise the actual Compose handoff

The suite executes extracted shell blocks, but removing the real Compose `environment:` mapping can leave the current tests green.

Please render the real Compose config with synthetic keys and assert:

```text
.env value
→ Compose environment
→ entrypoint array
→ exactly one --api-key flag
```

Negative controls should prove that removing the passthrough or changing `$$` to `$` makes the gate fail.

### 7. Clarify `/invocations` and exposure documentation

`/invocations` is not merely a health probe; on the pinned image it is an unauthenticated inference route outside:

```text
GUARDED_PREFIX = (/v1, /v2, /inference)
```

This PR does not need new middleware. Please state clearly that multi-key auth does not protect `/invocations` and that keyed deployments still require network-level access control of the server port.

Also correct the pinned runtime version to:

```text
0.25.2.dev0+g752a3a504.d20260714
```

and update the PR body to describe mutual exclusion—not key priority—and the final evidence accurately.

### 8. Expand the committed-key scan

Include:

- `docs/ENVS.md`
- `CHANGELOG.md`
- `scripts/test-dspark-api-keys.py`

in addition to Compose, the env example, and probe scripts.

After that revision:

1. hosted `validate` must pass on the exact rebased head;
2. we will rerun the full CPU/mutation matrix;
3. then run one final functional auth matrix on the pinned image:
   - every configured key → 200;
   - missing/wrong/Basic → 401;
   - `VLLM_API_KEY` alone remains compatible;
   - both variables → immediate exit 2 before patch work;
   - start/smoke/status probes pass;
   - keyless `/invocations` behavior is recorded;
   - startup and retained logs contain no key bytes under the chosen logging policy.

No performance A/B is required. This is a functional/security gate only.

Thanks again—the feature and overall implementation direction are good. The remaining work is bounded hardening and evidence, not a redesign.

sethforprivacy:
Round-2 hardening landed — branch rebased onto current `main` (`6d00e4a`, mergeable clean), **3 commits** covering the whole consolidated contract. New state addresses all 8 items plus the follow-ups from the quorum panel:

**1 — Rebase.** Done. The rebase resolved `CHANGELOG.md`, `docs/ENVS.md`, and `scripts/ci-validate.sh` as additive unions: every current-main entry is preserved (env-normalization, fabric passthrough, `DRAFT_SAMPLE_METHOD` gate, hosted `validate` workflow, #66 thinking-budget opt-in, plus the auth rows/tests from this PR).

**2 — Startup-log key exposure.** Decided: include the redaction patch in the recipe. New file `patches/hotfix-vllm-redact-api-key-log.sh` (verified against the real pinned vLLM `api_utils.py` anchor and proven behaviorally via `--status`) is appended to the existing compose `_hf` hotfix loop as one line — same loop, no new loop, `|| true` semantics consistent with the siblings. It redacts `api_key` inside `log_non_default_args` so the startup `non-default args` line shows `'api_key': ['<redacted:N value(s)>']` (count kept so the single-`--api-key`-flag diagnostic still works). Scope is the log channel only; argv/env exposure remains by design and documented. One known-forward note: the compose loop invokes every `_hf` sibling with `|| true` and the entrypoint only runs the file when it exists, so on a future image where upstream vLLM fixes the log path natively the patch exits non-fatal while `--status` reports NOT APPLIED and upstream's own redaction covers it — no silent leak either way. (Quorum caught a related gap: the patch was wired into compose but not the start script's worker `_hf_sync` whitelist — fixed: same filename appended so both ranks receive it; tests + a ci-validate guard now cover the sync.)

**3 — Before-side-effects rejection.** The entrypoint auth block now sits immediately after the initial exports (`export LD_LIBRARY_PATH=...`) and before `ENCODING_SOURCE`/encoder copy/hotfix/install work. Both-set fails with `error: VLLM_API_KEY and DSPARK_API_KEYS are both set; set exactly one of them`, exit 2, before any patch or install runs. Verified by execution against the extracted block: exit 2, exact message, and nothing else runs (no encoder copy, no hotfix, no install) before rejection. All three probe blocks are byte-identical (same md5).

**4 — Reject multiline/CR key lists.** `read -r -a` consumes only the first line; values like `k1\nk2` would silently drop `k2` and a `-bad` token could escape validation after a newline. The contract is now single-line: in all four contexts, newline or CR in the value is rejected BEFORE tokenization with `error: DSPARK_API_KEYS must be a single-line space-separated list` and exit 2. Newline/CRLF/dash-after-newline cases are asserted in all four contexts.

**5 — Hermetic tests.** `scripts/test-dspark-api-keys.py` now constructs a hermetic base env (any parent `VLLM_API_KEY` / `DSPARK_API_KEYS` are removed unless a case explicitly sets them) and includes hostile-parent regression runs proving ambient key variables cannot change any outcome.

**6 — Real Compose handoff.** A new `ComposedHandoff` layer (conditional: runs only when `docker compose` can render, otherwise skips cleanly) drives the actual chain: a stub `.env` with synthetic keys → `docker compose --env-file stub -f docker-compose.dspark.yml config --format json` → the rendered `environment` and rendered `command[2]` → the extracted auth block + exec tail executed under that environment → asserts exactly one `--api-key` carrying every key. Negative controls: dropping the `environment:` passthrough line breaks the chain (no `--api-key`); regressing `$$` to `$` causes compose to inline the keys into shell source and the interpolation gate catches it. This complements the direct execution of the extracted entrypoint block and probe blocks that was already in place.

**7 — `/invocations` and exposure docs.** `.env.dspark.example` + `docs/ENVS.md` (+ CHANGELOG) now state the pinned runtime `0.25.2.dev0+g752a3a504.d20260714` (Anemll 0.1.1), guarded prefix `(/v1, /v2, /inference)`, the unguarded endpoints `/health`, `/metrics`, `/version`, `/ping`, and that **`/invocations` is an unguarded inference route** — auth does not protect it, so keyed deployments still require network-level access control on the server port. The exposure contract is explicit: keys are container argv/env (`docker inspect` / host `ps` read them) and additionally appear in the startup Docker log unless the redaction hotfix is enabled — the docs recommend keeping the hotfix enabled whenever keys are used. PR body updated accordingly (mutual exclusion, not priority).

**8 — Committed-key scan.** The scan now covers compose, `.env.dspark.example`, all three probes, `docs/ENVS.md`, `CHANGELOG.md`, and `scripts/test-dspark-api-keys.py` itself.

**Final gates** (the reviewer's post-revision checklist): hosted `validate` is expected green on the rebased head; the full CPU suite passes on the head (`scripts/test-dspark-api-keys.py` = 36/36 with the handoff layer when docker can render, 36/36 with the layer skipped on docker-less hosts); `bash scripts/ci-validate.sh` passes end-to-end in a clean Linux container (bash 5 + stdlib python, no PyYAML). The final live auth matrix is ready for your run on the pinned image; the deployed clusters were updated to this state during round-1 and will pick the redaction + worker-sync improvements up at their next normal rotation.

All verification artifacts are in the branch. No credentials, IPs, or key material appear in the repo or in this comment.

PR #91, opened 2026-08-19
fix(spec-acceptance): per-position parse dropped every position, so the curve never printed

msdjzmst:
## What

`spec-acceptance.py`'s per-position curve has never printed anything. Three problems, all in that one feature:

**1. The parse drops every position (the reason nothing prints).**

On the vLLM build in the pinned `ghcr.io/anemll/dspark-vllm-gx10:0.1.1` image, `position` is emitted as the *last* label:

```
vllm:spec_decode_num_accepted_tokens_per_pos_total{engine="0",model_name="deepseek-v4-flash-0731",position="0"} 40263.0
```

so `line.split("position=")[1].split(",")[0].strip('"')` yields `0"} 40263.0`, `int()` raises `ValueError`, and the bare `except: pass` swallows it. `per_pos` is empty on every run. Fixed by matching the label directly with a regex, which is also order-independent if a future build reorders labels.

**2. Anchored the per-position branch on `_total`.** The old condition (`"accepted_tokens_per_pos" in line`) also matches the sibling `_created` gauge, whose value is a Unix timestamp (`1.787e9`). That was harmless only because the parse failed; with the parse fixed it would have contaminated the counts.

**3. The curve is now a window delta, and expressed as a rate.** It read `m2` alone — `m1["per_pos"]` was collected but never used — so even when populated it would have reported the container's lifetime totals rather than the burst just run. `drafts_total` is now read so counts can be divided into the rates `EVAL.md` already documents (`pos0 ~0.93 -> pos4 ~0.33`). The hardcoded `(k=5 -> max 5)` line is dropped; it printed the same value as the acceptance rate directly above it.

## Verified

Live 2x DGX Spark, TP=2, pinned `0.1.1` image, `MTP_NUM_TOKENS=5`:

- before: `per_pos` = `{}` (empty)
- after: `per_pos` = `{0: 44086, 1: 34923, 2: 26651, 3: 19687, 4: 14456}`
- **per-position values sum to `accepted_tokens_total` exactly** (139,803 == 139,803)
- derived draft width from `draft_tokens_total / drafts_total` = 5.00, matching `MTP_NUM_TOKENS=5`
- old parse reproduced independently: `ValueError: invalid literal for int() with base 10: '0"} 40263.0'`
- `python3 -m py_compile` clean

Sample output on that cluster (a real mixed agentic workload, not a synthetic burst — noting it only to show the shape, not as an expected value):

```
per-position acceptance (this window):
  pos0: 0.850
  pos1: 0.673
  pos2: 0.514
  pos3: 0.379
  pos4: 0.279
```

## Scope / non-goals

- **No behaviour change outside the per-position curve.** Overall acceptance is untouched.
- **Not a general Prometheus parser.** This keeps the existing simple line parsing and stays scoped to the pinned Anemll/vLLM output. Timestamped samples and OpenMetrics exemplars are out of scope, as they are for the rest of the file.
- **Counter reset is not addressed.** If the server restarts mid-window, `m2 < m1` and the run falls back to raw counts. That is a pre-existing property of the overall delta too, so it is left for a separate change rather than widened here.
- **Backwards compatible.** If a build does not expose `drafts_total`, the curve falls back to printing counts as before.
- **No doc changes.** `EVAL.md` and `AUDIT.md` already promise a per-position acceptance curve with rates; this makes the tool match what they describe.

`run-audit.sh` pipes this script's stdout to a log file and does not parse it, so the changed output format is not breaking.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@
 
 ### Fixed
 
+- **Speculative-acceptance per-position curve was always empty ([PR #91](https://github.com/MiaAI-Lab/DeepSeek-v4-Flash-DSpark-2x-DGX-Spark/pull/91))**: the parser assumed `position` was followed by another label, but Anemll `0.1.1` emits it last, so every sample raised inside a swallowed exception. The parser now matches the label independent of order, excludes the sibling `_created` timestamp gauge, and reports this measurement window as accepted tokens per draft rather than container-lifetime totals.
+
 - **RULER-lite never reached its advertised context lengths ([Issue #81](https://github.com/MiaAI-Lab/DeepSeek-v4-Flash-DSpark-2x-DGX-Spark/issues/81))**: `pad_to_length` appended one haystack sentence per loop with `guard < 200`, so every cell capped at ~4.8k tokens while still exiting 0. It now bulk-pads and `run_case` fails if `/tokenize` is under 97% of the target.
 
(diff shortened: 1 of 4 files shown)


plotarmordev (approved):
Approved at exact head `f6ffdf5002cd1ececbd9f67230e18ed07e35738f` after merging current `main`.

Confirmed the original defect against the committed Anemll 0.1.1 metrics shape: `position` is the final label, the sibling `_created` gauge carries timestamps, and the old parser drops the curve. Maintainer additions on the contributor branch add four regression tests, wire both the script and tests into CPU CI, and record the fix in the changelog.

Verification:
- `python3 scripts/test-spec-acceptance.py -v`: 4/4 passed
- `bash scripts/ci-validate.sh`: passed; 20 Python files compiled and all CPU recipe gates green
- diff hygiene and candidate-range secret scan: passed; 0 findings
- exact-head live CLI on the pinned image: 690 drafted, 508 accepted, 138 drafts, width 5.00; positions 0–4 printed; per-position counts sum exactly to 508; curve is monotone 0.986 → 0.449
- no restart or deployment mutation was required

No unresolved correctness or maintenance defect found. Hosted checks are not reported for this branch; the exact local equivalent passed.

Issue #92, opened 2026-08-20
scripts/spec-acceptance.py: per-position curve silently always empty; crashes when spec-decode is off; hardcoded k=5


wbaguley:
Three bugs in `scripts/spec-acceptance.py`, found while running `scripts/run-audit.sh` against a 2× DGX Spark serve. Patch attached at the end — verified against real `/metrics` output from a live `nvfp4_ds_mla` serve at `MTP_NUM_TOKENS=7`.

## 1. The per-position acceptance curve is ALWAYS empty (silent)

```python
pos = int(line.split("position=")[1].split(",")[0].strip('"'))
```

`position` is the **last** label in the metric line, so there is no trailing comma to split on:

```
vllm:spec_decode_num_accepted_tokens_per_pos_total{engine="0",model_name="m",position="0"} 16131.0
```

`line.split("position=")[1]` → `"0"} 16131.0`, and `.split(",")[0]` leaves it unchanged. `int('0"} 16131.0')` raises, and the surrounding `except Exception: pass` swallows it. Result: `per_pos` is always `{}`, so the report prints the header

```
per-position acceptance (pos0..pos4):
```

and then nothing at all. There is no error, so it reads as "no per-position data available" rather than a parsing bug.

This is the costly one, because the per-position curve is the datum that actually answers "is my `MTP_NUM_TOKENS` set too high?" On our serve, recovering it showed conditional acceptance flat at ~80% out to position 6 — which is only visible once the parse works:

```
pos0 16458   pos3  8819 (80.4%)   pos6 4626 (78.6%)
pos1 13540   pos4  7302 (82.8%)
pos2 10953   pos5  5897 (80.7%)
```

## 2. `TypeError` when spec-decode is not enabled

```python
m1 = get_metrics(args.base_url)
print(f"before: drafted={m1['drafted']:.0f} ...")   # line 58 — crashes here
...
if m1["drafted"] is None or m2["drafted"] is None:  # line 66 — the guard, 8 lines too late
    print("NO draft counters in /metrics — is spec-decode on? (check MTP_NUM_TOKENS)")
```

Against a serve without `--speculative-config`, the counters are absent and the f-string formats `None`:

```
TypeError: unsupported format string passed to NoneType.__format__
```

[...]

PR #94, opened 2026-08-20
feat(fabric): pass NCCL merge-NIC and GDR knobs through compose

de1tydev:
## Issue

`docker-compose.dspark.yml` forwards a fixed set of `NCCL_*` keys. Four fabric knobs are not in that set, so there is no way to set them from `.env.dspark` without editing the compose file itself (and keeping that edit alive across `git pull`):

- `NCCL_IB_MERGE_NICS` — NCCL's merge permission for compatible dual-port NICs (NCCL default `1`)
- `NCCL_NET_GDR_LEVEL`, `NCCL_NET_GDR_READ`, `NCCL_DMABUF_ENABLE` — GPUDirect RDMA / DMA-BUF controls

## Change

Passthrough only, with a true-unset contract:

```yaml
      NCCL_IB_MERGE_NICS: "${NCCL_IB_MERGE_NICS:-}"
      NCCL_NET_GDR_LEVEL: "${NCCL_NET_GDR_LEVEL:-}"
      NCCL_NET_GDR_READ: "${NCCL_NET_GDR_READ:-}"
      NCCL_DMABUF_ENABLE: "${NCCL_DMABUF_ENABLE:-}"
```

A compose map cannot conditionally omit a key, so an unconfigured knob would arrive as a *defined-but-empty* variable — and NCCL loads config-file values (`/etc/nccl.conf`, `NCCL_CONF_FILE`) with `setenv(..., overwrite=0)`, so that empty definition could mask them. The entrypoint therefore unsets empty definitions again before `exec`:

```sh
if [ -z "${NCCL_IB_MERGE_NICS:-}" ]; then unset NCCL_IB_MERGE_NICS; fi;
# … same for the other three
```

Observable contract: **unset stays absent** in the serving process (built-in defaults *and* config-file values still apply); **configured values pass through unchanged** on head and worker (same service definition on both nodes).

## What each retained knob is for

- `NCCL_IB_MERGE_NICS` already defaults to `1` inside NCCL: it *permits* merging of compatible dual-port NICs into one logical device. It does not select HCAs and does not force arbitrary links together — `NCCL_NET_MERGE_LEVEL` / `NCCL_NET_FORCE_MERGE` participate in that topology decision, and HCA selection stays with `NCCL_IB_HCA`. The meaningful override here is `0` (disable merging).
- `NCCL_DMABUF_ENABLE=0` is the meaningful workaround control (disable DMA-BUF probing where it misbehaves).
- `NCCL_NET_GDR_LEVEL` / `NCCL_NET_GDR_READ` are passed through as upstream overrides. No effect was demonstrated on the stack this was submitted from; they are documented as such.

What was observed on the submitting stack (2× DGX Spark GB10, TP=2, driver `580.173.02`, Anemll `0.1.1`): the container reported `CU_DEVICE_ATTRIBUTE_DMA_BUF_SUPPORTED=0` and boot logs showed host-staged transport (`via NET/IB/x`, no `/GDRDMA`). That is stated as an observation about that stack only, not as proof that every GDR route is unavailable there, and no performance claim is attached to any of the four knobs.

## Multi-HCA / GID coordination

The earlier "multi-HCA `NCCL_IB_HCA` requires `NCCL_IB_GID_AUTO=0` + a pinned index" guidance was stale and has been removed from this PR's docs: the launcher-side handling of multi-HCA selector forms under `NCCL_IB_GID_AUTO=1` is #95's scope, and the CHANGELOG entry here links to it instead of restating it.

## Validation

- `bash scripts/ci-validate.sh`: pass on this head.
- The previous source-text gate is replaced by `scripts/test-nccl-fabric-passthrough.sh` (wired into ci-validate), which exercises the extracted entrypoint normalization behaviorally:
  - all four defined-empty (what compose injects when `.env` leaves them unconfigured) → all four truly unset afterwards;
  - all four configured → byte-identical afterwards;
  - mixed set/unset → each key keeps its own state;
  - wiring checks (map entries present; normalization ordered before `exec vllm`);
  - where `docker compose` is available, `docker compose config` is rendered and checked (configured knob appears verbatim, unconfigured renders empty and is then unset by the entrypoint).

No live GPU run is claimed for this head; per review, this stays a truthful passthrough-only PR.


plotarmordev:
Thanks — the passthrough can be useful, but I cannot merge this head yet. Please revise it as follows:

1. Preserve true-unset behavior. The current `${VAR:-}` Compose mappings define each key as an empty environment variable when the user did not configure it. NCCL loads config-file values with `setenv(..., overwrite=0)`, so a defined-empty key can mask `/etc/nccl.conf` / `${NCCL_CONF_FILE}` values. Unset must remain absent; configured values must pass through unchanged on head and worker. Replace the current source-text gate with CPU evidence for that observable contract.
2. Correct the merge-NIC explanation everywhere. `NCCL_IB_MERGE_NICS` already defaults to `1`; it permits compatible dual-port merging but does not select HCAs or automatically fuse arbitrary links. `NCCL_NET_MERGE_LEVEL` / `NCCL_NET_FORCE_MERGE` participate in the topology decision.
3. Remove or reframe the reported performance numbers. The submitted comparison changed the HCA selection and did not isolate any of the four newly forwarded variables; the logs show alternating physical NICs, not a `Made virtual device ... ndevs=2` merge.
4. Describe the GDR/DMA-BUF controls precisely. Lack of DMA-BUF support does not by itself prove every GDR route is unavailable. State only what was observed on the tested stack.
5. Remove the “compose and .env must be rolled back together” warning; that failure mode is not real.
6. Make each retained knob intentional: `MERGE_NICS=0` and `DMABUF_ENABLE=0` are the meaningful disable/workaround controls; label the GDR controls as upstream overrides with no demonstrated effect on the submitted stack, or drop knobs that have no concrete contract.
7. Reconcile the multi-HCA/GID guidance and CHANGELOG with #95.

No live GPU gate is required if performance claims are removed and this remains a truthful passthrough-only PR. I will re-review the revised exact head.

de1tydev:
Revised at `7ef68fe`; PR body rewritten to match. Point by point:

1. **True-unset preserved.** A compose map can't conditionally omit a key, so the entrypoint now unsets empty definitions before `exec`: unset stays absent in the serving process (so `/etc/nccl.conf` / `${NCCL_CONF_FILE}` values loaded with `overwrite=0` are no longer masked by a defined-empty key), configured values pass through unchanged, and head and worker share the same service definition. The source-text ci-validate gate is replaced by `scripts/test-nccl-fabric-passthrough.sh`, which runs the *extracted* entrypoint fragment in both directions (all-empty → all absent; configured → byte-identical; mixed → each key keeps its own state), checks that the normalization is ordered before `exec vllm`, and — where `docker compose` is available — asserts the rendered `config` environment values. 6/6 locally.
2. **Merge-NIC explanation corrected everywhere** (compose comment, `.env.dspark.example`, `docs/ENVS.md`, CHANGELOG, PR body): `NCCL_IB_MERGE_NICS` defaults to `1` and *permits* compatible dual-port merging; it does not select HCAs or fuse arbitrary links; `NCCL_NET_MERGE_LEVEL`/`NCCL_NET_FORCE_MERGE` participate in the topology decision; HCA selection stays with `NCCL_IB_HCA`.
3. **Performance numbers removed** from the CHANGELOG and PR body. Agreed on the confound — the comparison changed HCA selection and the logs showed alternating physical NICs, not a `Made virtual device … ndevs=2` merge. This is now a passthrough-only PR with no behavior claim attached to any knob.
4. **GDR/DMA-BUF wording narrowed to the observation**: on the submitting GB10 stack (driver `580.173.02`), the container reported `CU_DEVICE_ATTRIBUTE_DMA_BUF_SUPPORTED=0` and boot logs showed `via NET/IB/x` with no `/GDRDMA`; stated as that stack's observation only, with no claim that every GDR route is unavailable.
5. **Rollback-together warning removed** from compose comment, env example, ENVS.md, and CHANGELOG.
6. **Knob intent labeled**: `NCCL_IB_MERGE_NICS=0` and `NCCL_DMABUF_ENABLE=0` documented as the meaningful disable/workaround controls; `NCCL_NET_GDR_LEVEL`/`NCCL_NET_GDR_READ` labeled upstream overrides with no demonstrated effect on the submitted stack. All four kept since the passthrough contract (unset-absent / set-verbatim) is now tested.
7. **Reconciled with #95**: the stale "multi-HCA requires `NCCL_IB_GID_AUTO=0` + pinned index" guidance is dropped from this PR entirely; the CHANGELOG entry links launcher-side multi-HCA selector handling to #95 instead of restating it.

`bash scripts/ci-validate.sh` passes on the head (`CI validate passed (CPU recipe gates only)`).

plotarmordev:
Thanks — I re-reviewed the updated exact head `7ef68fee19f07829aea00adf289e0a67a0be2aa9` against current `main` `21da90f1d2000661618fdb21dcd70d0afdddc7bf`.

The seven requested semantic/documentation corrections are substantially resolved: true-unset normalization is now behavior-covered, NCCL merge-NIC/GDR wording is appropriately bounded, performance claims and rollback warning are gone, retained knobs are intentional, and #95 owns selector/GID semantics.

Remaining before merge:
1. Rebase/merge current `main`; the current head is dirty with a real `CHANGELOG.md` conflict. Preserve main's existing 2026-08-20 entries and this PR's addition.
2. Change “configured values pass through unchanged” to the precise contract: **configured non-empty values pass through unchanged; empty is normalized to absent**.
3. Keep any GB10 runtime statements explicitly framed as contributor observations unless exact logs are attached.
4. Push the contributor-owned resolved head and require hosted `validate` success on that exact SHA.

No GPU A/B is required if this remains a passthrough-only change with no performance claim.

de1tydev:
Revised at `9108b2ed6a74a4bfdd883d75723d3797c0739b6e`. Point by point:

1. **Rebased.** The branch is now on current `main` `5d5a00d81539a2f3708718fbe185a61dd8c387cc` — that is two commits past the `21da90f1` you cited, so I took the newer tip rather than the one named. The `CHANGELOG.md` conflict is resolved by keeping main's existing 2026-08-20 `### Changed` (issue #66) and `### Fixed` (hub timeouts, `spec-acceptance.py` #92) entries untouched and placing this PR's `### Added` entry above them, per Keep-a-Changelog ordering. Nothing from main was dropped.

2. **Precise contract wording.** "Configured values pass through unchanged" is replaced everywhere by the exact contract: **configured non-empty values pass through unchanged; an empty value is normalized to absent** (and so cannot be used to force an empty setting). Updated in the compose comment, `.env.dspark.example`, `docs/ENVS.md` — where it is now stated once as the contract for all four knobs — and the CHANGELOG.

   `scripts/test-nccl-fabric-passthrough.sh` covers the boundary rather than just the two ends: alongside empty→absent and set→byte-identical, a new case asserts that a value that is *non-empty but unusual* (leading/trailing whitespace, mixed case) is still a configured value and arrives byte-identical rather than being trimmed or dropped. 7/7 locally.

3. **GB10 statements bounded.** The DMA-BUF/GDR observations are labelled contributor-reported, from the submitting stack only, in `.env.dspark.example`, `docs/ENVS.md` and the CHANGELOG, each with an explicit note that this is not a claim about GDR availability in general. No exact logs are attached, so nothing is stated beyond what was observed on that one stack.

4. **Hosted gate.** `validate` passes on this exact SHA. `bash scripts/ci-validate.sh` also passes locally.

Still passthrough-only with no performance claim attached to any knob.

PR #95, opened 2026-08-20
fix(launcher): NCCL_IB_GID_AUTO FATALs on multi-HCA NCCL_IB_HCA syntax

de1tydev:
## Issue

`start-deepseek-v4-flash-dspark.sh` defaults to `NCCL_IB_GID_AUTO=1` and resolves the RoCEv2 GID index per node from sysfs. It did that by pasting the `NCCL_IB_HCA` value straight into a sysfs path:

```sh
for g in /sys/class/infiniband/\$hca/ports/1/gids/*; do
```

`NCCL_IB_HCA` is not a bare device name. NCCL (`parseStringList` in `src/misc/utils.cc`, applied by `ncclIbInit` in `net_ib.cc`) accepts an optional leading `^` (exclusion), then an optional `=` (exact name match instead of prefix match), then a comma-separated list of `name[:port[:rail[:plane]]]` tokens. With any list form — e.g. the exact list `NCCL_IB_HCA==<dev0>,<dev1>`, which is *recommended filtering* for deterministic member selection on a multi-HCA node (not something `NCCL_IB_MERGE_NICS` itself requires) — the glob became `/sys/class/infiniband/=<dev0>,<dev1>/ports/1/gids/*`, matched nothing, and the launcher FATALed before starting either rank.

## Fix

The resolver runs a selector engine on the node that owns the sysfs tree, mirroring NCCL's semantics rather than flattening the spec.

**Selector grammar** — `^` exclusion and `^=` exact exclusion, `=` exact lists, default prefix matching, explicit `dev:port` tokens, omitted-port = all ports of the device, empty selector = every candidate, include = `match XOR exclude` per (device, port) pair. The port field follows `parseStringList` exactly: absent *or empty* is `-1` (any port), a non-empty field is `atoi()` — base 10, so `:08` is port 8 and `:010` is port 10, stopping at the first non-digit. A port too wide to represent is **clamped, not evaluated**: shell arithmetic wraps modulo 2^64, and `18446744073709551615` wraps to exactly `-1`, so an unrepresentable port would otherwise turn into the "any port" wildcard and silently *widen* the selection. Rail/plane fields are parsed off and ignored.

**Candidate universe** — the selector is applied to the same set `ncclIbInit` builds: ports that are `IBV_PORT_ACTIVE`, link layer Ethernet or InfiniBand, at most `MAX_IB_DEVS=32` entries. Both filters run *before* `NCCL_IB_HCA`, exactly as NCCL orders them, so the DOWN sibling port these dual-port cards ship cannot be selected into a fail-closed error or drag the index intersection — NCCL never opens it either. An attribute that cannot be read is not evidence of inactivity, so such a port stays a candidate. Truncation at the cap names the ignored members on stderr.

**Per-member validation** — every selected member is validated independently: preferred match IP first, otherwise an IPv4 on the member's *own* netdev (`gid_attrs/ndevs/<i>` + `ip -4 -o addr show`). One shared match IP can no longer silently drop a member that uses another link address.

**Index reconciliation** — members are reconciled by **intersecting their sets of usable RoCE v2 indexes**. A member commonly has more than one usable index, so picking one winner per member and comparing the picks reported a disagreement even when a common global index existed. From the intersection the resolver takes the lowest index, preferring one that some member reached through the match IP.

**Fail closed** — exit 1 when a selected member has no usable RoCE v2 GID, or when the selector matches no candidate (the FATAL then names ports skipped as non-ACTIVE or unsupported-link-layer, so a DOWN port does not look like a typo). Exit 3, with **each member's whole usable set** (`dev:port=i,j`), when the intersection is empty: `NCCL_IB_GID_INDEX` is one global value per rank, so no single manual pin can satisfy a disagreeing selection — the FATAL says to narrow the selector and deliberately does not suggest a pin.

**Literal transport** — selector tokens are injected with `printf %q` and the engine runs under `set -f`, so glob metacharacters and whitespace never expand or split.

Per-member resolutions are logged (`member dev:port -> RoCEv2 gid index N (via match-ip …/own-addr … on <netdev>)`), so a live boot records selected HCA/ports, match addresses, and resolved indexes for both ranks.

A single bare device name — the `.env.dspark.example` default — resolves exactly as before. No defaults change; no cluster-specific value is introduced.

## Tests

`scripts/test-nccl-ib-hca-gid-resolve.sh` (CPU-only, wired into `scripts/ci-validate.sh`) extracts the launcher's own functions, stubs the ssh/bash transport and `ip`(8) (fixture-driven), redirects the sysfs root to fake trees, and runs the launcher's *own* generated lookup script. **60 checks**:

- prefix matching incl. **prefix collisions**, exact lists, comma lists without `=`, **exclusions and exact exclusions** (incl. `^dev:port`);
- **ports 1 and 2** explicitly, **omitted-port multiport** (agreeing → resolves; disagreeing → exit 3);
- **port grammar**: `:08`, `:010`, `:+8`, `: 8`, `:8abc`, `:abc`, `devP:`, `devP::2`, `devP:8:2:0`, `:0000000008`, and the **overflow cases** — `:18446744073709551615` (the value that wraps to the `-1` wildcard), an absurdly wide positive and a wide negative port, all of which must match nothing rather than everything;
- **candidate universe**: a DOWN sibling port dropping out (omitted-port selection still resolves), an explicitly selected DOWN port failing closed, `ACTIVE_DEFER` not counting as `IBV_PORT_ACTIVE`, an unsupported link layer dropping out, a port with unreadable attributes staying in, the skipped-candidate diagnostic, and the **`MAX_IB_DEVS=32` cap** with its truncation notice;
- **index sets**: overlapping sets resolving to the common index, genuinely disjoint sets failing closed with each member's full usable set asserted, and a **discriminating match-IP preference** case — both members usable at `{1,5}`, only `5` reachable through the match IP, so lowest-wins answers `1` and the preference answers `5` (verified by mutation: deleting the preference branch fails exactly that check and nothing else);
- **distinct per-HCA addresses/indexes** resolved via each member's own netdev, mixed match-ip + own-addr members;
- **independent head/worker layouts** in one run (local and ssh transport paths);
- **literal glob/whitespace transport** (run from a directory seeded with glob-trap files);
- **missing devices** (selector matching nothing → exit 1; missing sysfs tree → exit 1; a token naming a missing device alongside a real one keeps the real one, mirroring NCCL);
- the **original exact-list regression**, and a RoCE v1 GID with the same value present to confirm it is skipped.

Regression power: run against upstream `main`'s launcher the suite fails **43 of 60** behaviorally (wrong result / wrong exit code, not an extraction abort); against the previously reviewed head the new cases fail **8 of 60**.

`bash scripts/ci-validate.sh` passes on this head; hosted `validate` is green on this exact SHA.

## Evidence

The original FATAL was hit on our own 2× DGX Spark GB10 pair (TP=2, two direct RoCE links, driver `580.173.02`, Anemll `0.1.1`, official `DeepSeek-V4-Flash-0731`, NCCL `2.30.7`).

**Two live functional boots ran on this exact head** — deployed launcher byte-identical to `f8ea597`'s (`sha256 0ac35563…`), `NCCL_IB_GID_AUTO=1`, no GID pins:

- `NCCL_IB_HCA=roce` (prefix). The previous head could not start with this selector at all: these nodes carry two permanently DOWN `roce*` ports with only link-local GIDs, and the old candidate set failed closed on them. Now both ranks resolve index 3, and NCCL's own `NET/IB : Using [0]rocep1s0f1:1/RoCE [1]roceP2p1s0f1:1/RoCE` confirms it selected exactly the two members the resolver did — the mirrored candidate universe checked against the deployed one, not just against the fake-sysfs harness.
- `NCCL_IB_HCA==rocep1s0f1,roceP2p1s0f1` (the intended multi-HCA selector, now the production configuration). Both ranks resolve index 3, the second member through the own-netdev path the pre-revision resolver silently dropped.

Both boots: `BOOT_EXIT=0`, channels alternating across both physical links, zero `NCCL WARN`/`NCCL ERROR`, launcher readiness gate + built-in smoke passed, and a served request answered with `finish_reason=stop`. Full per-rank capture is in the [boot evidence comment](https://github.com/MiaAI-Lab/DeepSeek-v4-Flash-DSpark-2x-DGX-Spark/pull/95#issuecomment-5371092795).

Not claimed: this is a functional gate, not a performance A/B, and no GDR/DMA-BUF behavior is claimed (this driver reports no DMA-BUF support and transport stayed host-staged).


plotarmordev:
Thanks — this fixes a real multi-HCA launcher problem, but the current resolver can silently choose a wrong GID index. Please keep the PR open and revise it before review:

1. Mirror NCCL selector semantics on the target node, or reject unsupported forms explicitly and fail closed. This includes exclusion (`^` then optional `=`), default prefix matching, exact lists, explicit ports, and omitted/all-port behavior. Do not flatten every selector to a bare port-1 device.
2. Validate every selected HCA/port against its own local address/GID. One shared match IP can silently omit a selected HCA that uses another link address.
3. Fail closed when selected members require different numeric GID indexes. “Warn and use first” knowingly supplies a wrong global index to at least one member; another single manual pin cannot solve that disagreement.
4. Preserve selector tokens literally; avoid shell pathname/glob expansion.
5. Expand the behavioral suite to cover prefix collisions, exclusions/exact exclusions, ports 1 and 2, omitted-port multiport, distinct per-HCA addresses/indexes, disagreement failure, independent head/worker layouts, literal glob/whitespace transport, missing devices, and the original exact-list regression. The main-side regression must fail behaviorally, not only because a new helper cannot be extracted.
6. Correct the PR body, launcher comments, example, tests, and CHANGELOG to match the supported grammar. Exact HCA lists are recommended filtering, not a documented requirement of `NCCL_IB_MERGE_NICS`.
7. Coordinate #94’s stale `NCCL_IB_GID_AUTO=0` guidance and the shared CHANGELOG section.

After CPU gates pass, run one live dual-link boot on the revised exact head with the intended multi-HCA selector and `NCCL_IB_GID_AUTO=1`. Capture both ranks’ selected HCA/ports, match addresses, resolved indexes, NCCL initialization, and a successful request. I will re-review after those changes.

de1tydev:
Revised at `e71f6ff`; PR body rewritten to match. Point by point:

1. **NCCL selector semantics mirrored on the target node.** The flattening helper is gone. The generated lookup script now parses the selector the way `net_ib.cc` does — optional leading `^` (exclusion), then optional `=` (exact match), comma-separated `name[:port]` tokens with `atoi()` port parsing, omitted port = every port, empty list = everything — and applies include = `match XOR exclude` per (device, port) pair enumerated from sysfs. Nothing is flattened to a bare port-1 device.
2. **Every selected member validates against its own address.** Preferred match IP first; otherwise the member's own netdev IPv4 (`gid_attrs/ndevs/<i>` + `ip -4 -o addr`). A member that validates against neither is no longer silently omitted — it fails the resolve (exit 1) with the member named.
3. **Disagreement fails closed.** Members needing different numeric indexes → exit 3 with a `dev:port=index` per-member diagnostic; the launcher's FATAL explains that `NCCL_IB_GID_INDEX` is one global value per rank and says to narrow the selector — it deliberately does not suggest a manual pin, since no single pin can satisfy the disagreement.
4. **Literal token transport.** Inputs are injected with `printf %q` and the engine runs under `set -f`; glob metacharacters and whitespace inside tokens neither expand nor split (tested from a directory seeded with glob-trap files).
5. **Behavioral suite expanded to 30 checks**: prefix collisions, exclusions/exact exclusions (incl. `^dev:port`), ports 1 and 2, omitted-port multiport (agreeing and disagreeing), distinct per-HCA addresses/indexes via own-netdev validation, disagreement failure with the diagnostic asserted, independent head/worker layouts (local + ssh transport paths), literal glob/whitespace transport, missing devices/missing sysfs, and the original exact-list regression. Run against the pre-fix `main` launcher the suite still executes and fails **20/30 behaviorally** — the extraction no longer aborts, so the main-side regression shows up as wrong results/exit codes, not a missing helper.
6. **Grammar corrected everywhere** (PR body, launcher comments, `.env.dspark.example`, test header, CHANGELOG): exact HCA lists are described as *recommended filtering* for deterministic member selection on a multi-HCA node — not a requirement of `NCCL_IB_MERGE_NICS`.
7. **Coordinated with #94**: #94 has been revised in parallel — its stale `NCCL_IB_GID_AUTO=0` guidance is removed entirely and its CHANGELOG entry now defers multi-HCA selector handling to this PR, so the two entries no longer contradict.

CPU gates: `bash scripts/ci-validate.sh` passes on the head (30/30 in the suite).

**Live dual-link boot**: not yet run on this head. Restarting the pair is a production service interruption on our side, so it is scheduled for our next maintenance window; I will post the capture here as a follow-up (both ranks' `member dev:port -> RoCEv2 gid index N (via …)` lines — the resolver now logs exactly the fields you asked for — plus match addresses, NCCL init, and a served request with `NCCL_IB_GID_AUTO=1` and the intended multi-HCA selector).

de1tydev:
Live dual-link boot evidence for the revised exact head, as requested.

**Exact-head accounting**: the deployed `start-deepseek-v4-flash-dspark.sh` is byte-identical to this PR's head `e71f6ff` (`sha256 0c71ad405be562a8731bdf2378a3ab2c44bc4c7bc76ff15bda6fb6e646e0c1aa`, matching `git show e71f6ff:start-deepseek-v4-flash-dspark.sh`). Stack: 2× DGX Spark GB10, TP=2, driver `580.173.02`, Anemll `0.1.1` (digest-pinned), official `DeepSeek-V4-Flash-0731`, deploy tree at upstream `6252303` (the two unrelated local deltas are in compose/prepare, not the launcher).

**Configuration under test** (both nodes): `NCCL_IB_HCA==rocep1s0f1,roceP2p1s0f1` (exact multi-HCA list), `NCCL_IB_GID_AUTO=1`, previous workaround pins (`NCCL_IB_GID_INDEX=3` / `WORKER_NCCL_IB_GID_INDEX=3`) commented out.

**Resolver output at boot** (launcher stdout, verbatim — both ranks' selected members, match addresses, resolved indexes):

```
Resolving RoCEv2 GID indexes from sysfs (head if=enp1s0f1np1 ip=10.100.216.2 selector==rocep1s0f1,roceP2p1s0f1; worker if=enp1s0f1np1 ip=10.100.216.1 selector==rocep1s0f1,roceP2p1s0f1)...
  member rocep1s0f1:1 -> RoCEv2 gid index 3 (via match-ip 10.100.216.2)
  member roceP2p1s0f1:1 -> RoCEv2 gid index 3 (via own-addr 10.100.217.2 on enP2p1s0f1np1)
  member rocep1s0f1:1 -> RoCEv2 gid index 3 (via match-ip 10.100.216.1)
  member roceP2p1s0f1:1 -> RoCEv2 gid index 3 (via own-addr 10.100.217.1 on enP2p1s0f1np1)
RoCEv2 GID index: head=3 (match 10.100.216.2) worker=3 (match 10.100.216.1)
```

Note the second member on each rank resolves through the **own-netdev address path** (`10.100.217.x` on `enP2p1s0f1np1`) — exactly the member the pre-revision resolver silently dropped under a single shared match IP.

**NCCL initialization**: both ranks injected `NCCL_IB_GID_INDEX=3` (auto-resolved, matching what we had hand-pinned before); init completed normally with channels alternating across both physical links — `via NET/IB/0` ×180 / `via NET/IB/1` ×180 in the boot log, no errors, no `/GDRDMA` (expected on this driver).

**Served request** (post-boot, same boot):

```
SMOKE: window boot2 smoke ok  finish= stop
```

Start script exited 0 (its readiness gate + built-in smoke). The stack was booted twice in this maintenance window — once with an unrelated opt-in experiment enabled and once stock; the resolver output was identical in both, and the quotes above are from the final stock boot, which is the configuration now in production (`NCCL_IB_GID_AUTO=1` staying on, pins retired). A pre-window read-only dry-run of the same resolver body against both nodes' live sysfs had predicted exactly these member lines.

plotarmordev:
Thanks — the revision at `e71f6ff835ab5fa7342437ac01085626f5d8f633` genuinely fixes most requested safety defects: target-node prefix/exact/exclusion selection, literal transport, per-selected-member address validation, and the old warn/use-first disagreement path.

Remaining before merge:
1. Rebase onto current `main` `21da90f1d2000661618fdb21dcd70d0afdddc7bf`; the current head is dirty and has no hosted check.
2. Match NCCL port grammar exactly: explicit empty `dev:` is wildcard/absent (`-1`), not port 0; decimal `atoi` forms such as `:08` must not fail Bash arithmetic or change meaning. Add focused cases.
3. Resolve per-member indexes by **intersecting usable index sets**. Choosing one arbitrary own-address winner per member can falsely report disagreement even when a common global index exists. Add a case proving the common-index path.
4. Run the exact-head hosted gate and repeat the final-head live boot/resolver capture after those source changes.

Please keep this contributor-owned; we will re-review the new exact SHA rather than patching the stale branch.

de1tydev:
Revised at `5e8884c573acd086112107c28322d721a21d5321`. Point by point:

1. **Rebased.** On current `main` `5d5a00d81539a2f3708718fbe185a61dd8c387cc` — two commits past the `21da90f1` you cited, so I took the newer tip. The `CHANGELOG.md` conflict is resolved by keeping main's 2026-08-20 `### Changed` and `### Fixed` entries intact and merging this PR's entry into the existing `### Fixed` section rather than emitting a second heading.

2. **Port grammar matched exactly — you were right on both counts.** I checked against `parseStringList` in `src/misc/utils.cc`, which reads `name[:port[:rail[:plane]]]` and comments that "port, rail, and plane will default to -1 if absent or empty":

   - An explicitly empty port field (`devA:`) was being read as port `0`, which matches no real port, instead of `-1` = any port. Fixed; the port field is also now taken *before* any rail/plane field.
   - Non-empty fields went straight into shell arithmetic, where a leading zero means octal. `:08` aborted with `value too great for base`, and — worse, because it is silent — `:010` meant port **8** instead of 10. Parsing is now forced to base 10 so it reads the way `atoi()` does. `:abc` still yields 0 and matches nothing; leading blanks, a sign, and trailing garbage (`:8abc`) behave as `atoi()`.

   New focused cases cover `:08`, `:010`, `:+8`, `: 8`, `:8abc`, `:abc`, `devP:`, `devP::2` and `devP:8:2:0` against a tree with ports 1/8/10 so the leading-zero forms are actually distinguishable.

3. **Usable index sets are intersected.** The old code took each member's first usable index and compared the picks, so two members that each had several usable indexes were reported as disagreeing whenever their arbitrary picks differed — exactly the false disagreement you described. Members' usable sets are now intersected, and only an empty intersection is fatal. The exit-3 diagnostic prints each member's *whole* usable set instead of one pick. From the intersection the resolver takes the lowest index, preferring one that some member reached through the match IP, which preserves the choice this stack was already making.

   `members with overlapping index sets resolve to the common index` proves the path: two members with usable sets `{1,5}` and `{3,5}` now resolve to `5`, where the previous head reported a disagreement. `genuinely disjoint index sets still fail closed (exit 3)` keeps the real failure real.

4. **Hosted gate: green on this exact SHA.** The suite is now 46 checks, all passing. Against the previous head `e71f6ff` the new cases fail **8/46** behaviorally — `:010` resolving to 8, `:08` erroring out, `devP:` matching nothing, and the false disagreement — and against upstream `main` **35/46**. So the new coverage fails for the reasons it is meant to, not by extraction.

**Live boot: not re-run on this head, and I am not claiming it.** Restarting the pair is a production interruption here, so the dual-link boot capture is scheduled for our next maintenance window and I will post it as a follow-up.

What I *can* offer now is a **read-only dry-run of this exact head's resolver body against both nodes' live sysfs** — it only reads `/sys/class/infiniband` and runs `ip -4 -o addr show`, so it needs no restart. Same stack, same `NCCL_IB_HCA==rocep1s0f1,roceP2p1s0f1`, `NCCL_IB_GID_AUTO=1`:

```
head   (match 10.100.216.2) -> 3
  member rocep1s0f1:1   -> RoCEv2 gid index 3 (via match-ip 10.100.216.2)
  member roceP2p1s0f1:1 -> RoCEv2 gid index 3 (via own-addr 10.100.217.2 on enP2p1s0f1np1)
worker (match 10.100.216.1) -> 3
  member rocep1s0f1:1   -> RoCEv2 gid index 3 (via match-ip 10.100.216.1)
  member roceP2p1s0f1:1 -> RoCEv2 gid index 3 (via own-addr 10.100.217.1 on enP2p1s0f1np1)
```

Running the previous head `e71f6ff` against the same live sysfs in the same session gives byte-identical member lines and the same resolved index, so on this hardware the two source changes are behaviour-preserving and the earlier boot capture still describes what this head does. That is a dry-run against live state, not a boot — treat it as such until the window capture lands.

plotarmordev:
Thanks — the current resolver design is substantially improved, and the exact current-main merge preview passes all 46 focused checks plus full CPU CI. A final bounded revision is still needed before the live gate:

1. **Port overflow must fail closed.** A sufficiently large decimal can make Bash arithmetic fail while leaving the initialized `port=-1` wildcard. Clamp/reject before arithmetic and add a focused overflow case.
2. **Make the match-IP preference test discriminating.** Use a common set such as `{1,5}` where own-address supports both and match-IP supports only 5; expect 5. The current singleton intersection cannot detect removal of the preference branch.
3. **Mirror or explicitly document the deployed NCCL candidate universe:** inactive/unsupported-link ports that NCCL skips, and the `MAX_IB_DEVS=32` selector cap. Add one behavior case for the chosen contract.
4. Refresh the PR body to the current 46-check/intersection design and usable-set diagnostics.
5. Rebase onto current `main` (PRs #101 and #94 changed the launcher/adjacent review surfaces), preserve all disjoint changes, and require hosted `validate` on the new exact head.
6. After source/CPU gates are clean, repeat the final-head dual-link functional boot: both ranks' member→index lines, NCCL initialization across both links, exact launcher identity, and one served request.

This is a functional boot gate, not a performance A/B. Please keep the revision contributor-owned; we will re-review the exact new SHA.

de1tydev:
Revised at `f8ea597c3e2545f86bd86c8ce11749b431352651`, and the live gate is done — two functional boots on this exact head. Point by point:

**1. Port overflow fails closed.** You were right that the initialized `port=-1` was reachable, though the mechanism is worse than an arithmetic abort: Bash `$(( ))` does not fail on a wide decimal, it wraps modulo 2^64 — and `18446744073709551615` wraps to exactly `-1`, which *is* the "any port" wildcard. So `NCCL_IB_HCA=devA:18446744073709551615` silently selected every port of `devA` instead of none. The magnitude is now taken after stripping sign and zero padding, and anything wider than 9 digits is clamped to a value no sysfs port can have, so the token matches nothing and the resolve fails closed before any arithmetic. Focused cases: the `-1`-wraparound value, an absurdly wide positive, a wide negative, and `:0000000008` (padding is not width — still port 8). On the previous head the wraparound case resolves as a wildcard and fails behaviorally.

**2. The match-IP preference test is now discriminating.** Exactly your construction: both members usable at `{1,5}`, own-address validation supports both indexes for both members, and only `5` is reachable through the match IP. Lowest-wins answers `1`; the preference answers `5`. Both directions are asserted (`no match-ip hit: the intersection {1,5} takes its lowest index` → 1, `match-ip hit is preferred over a lower own-addr index in the same intersection` → 5). Verified by mutation: deleting the preference branch fails exactly that one check out of 60 — the old singleton-intersection case still passed, as you predicted.

**3. The NCCL candidate universe is mirrored, not just documented.** `ncclIbInit` skips a port whose state is not `IBV_PORT_ACTIVE` and whose link layer is neither Ethernet nor InfiniBand, applies both tests *before* `NCCL_IB_HCA`, and keeps at most `MAX_IB_DEVS=32` entries. The resolver now does all three. An attribute that cannot be read is not evidence of inactivity, so such a port stays a candidate; the "matched nothing" FATAL names ports skipped as non-ACTIVE or unsupported-link-layer, and truncation at the cap names the ignored members. Behavior cases: DOWN sibling port dropping out of an omitted-port selection, explicitly selecting a DOWN port failing closed, `ACTIVE_DEFER` not counting as ACTIVE, unsupported link layer dropping out, unreadable attributes staying in, the skipped-candidate diagnostic, and a 33-device tree proving the cap.

This is not academic on the deployed hardware. Both GB10 nodes expose four `roce*` devices, two of which are permanently DOWN (`rocep1s0f0:1`, `roceP2p1s0f0:1`, `phys_state 3: Disabled`) and carry only link-local GIDs. A plain prefix selector was therefore fatal on the previous head — the previously reviewed resolver body and this one, run read-only against both nodes' live sysfs in the same session:

```
# previous head, selector 'roce'
FATAL: no usable RoCEv2 GID on selected member(s): rocep1s0f0:1 roceP2p1s0f0:1 (no GID matches 10.100.216.2 or an IPv4 on the member's own netdev)
# this head, selector 'roce'
  member rocep1s0f1:1   -> RoCEv2 gid index 3 (via match-ip 10.100.216.2)
  member roceP2p1s0f1:1 -> RoCEv2 gid index 3 (via own-addr 10.100.217.2 on enP2p1s0f1np1)
```

**4. PR body refreshed** to the 60-check suite, the intersection design, usable-set diagnostics, the clamp, and the candidate-universe contract.

**5. Rebased** onto current `main` `6d00e4a` (which now contains #101, #94 and #102), disjoint changes preserved; the `CHANGELOG.md` conflict is resolved by merging into the existing `### Fixed` section rather than emitting a second heading. Hosted `validate` is **green on this exact SHA** (`f8ea597`), and the PR shows mergeable/clean. Locally `bash scripts/ci-validate.sh` passes with 60/60 in the suite. Regression power of the suite: **43/60 fail behaviorally** against upstream `main`'s launcher, **8/60** against the previously reviewed head — wrong results and wrong exit codes, not extraction aborts.

**6. Live functional boots on this exact head — two of them.**

**Exact-head accounting**: the deployed `start-deepseek-v4-flash-dspark.sh` is byte-identical to `f8ea597`'s (`sha256 0ac355639a02152fa4452224093763a94c6bc4b54ac1a0198af1e45bab91b647`, matching `git show f8ea597:start-deepseek-v4-flash-dspark.sh`). Stack: 2× DGX Spark GB10, TP=2, two direct RoCE links, driver `580.173.02`, Anemll `0.1.1` (digest-pinned `sha256:a8394849…`), official `DeepSeek-V4-Flash-0731`, `NCCL_IB_GID_AUTO=1`, no `NCCL_IB_GID_INDEX` pins.

**Boot A — prefix selector `NCCL_IB_HCA=roce`** (22:02–22:08 CST, `BOOT_EXIT=0`). This is the selector the previous head could not start with, so it exercises the new candidate universe end to end.

```
Resolving RoCEv2 GID indexes from sysfs (head if=enp1s0f1np1 ip=10.100.216.2 selector=roce; worker if=enp1s0f1np1 ip=10.100.216.1 selector=roce)...
  member rocep1s0f1:1 -> RoCEv2 gid index 3 (via match-ip 10.100.216.2)
  member roceP2p1s0f1:1 -> RoCEv2 gid index 3 (via own-addr 10.100.217.2 on enP2p1s0f1np1)
  member rocep1s0f1:1 -> RoCEv2 gid index 3 (via match-ip 10.100.216.1)
  member roceP2p1s0f1:1 -> RoCEv2 gid index 3 (via own-addr 10.100.217.1 on enP2p1s0f1np1)
RoCEv2 GID index: head=3 (match 10.100.216.2) worker=3 (match 10.100.216.1)
```

NCCL's own view on the same boot — the check that matters for item 3, because it is NCCL and not our mirror deciding:

```
spark-9f8f:315:315 [0] NCCL INFO NCCL_IB_HCA set to roce
spark-9f8f:315:315 [0] NCCL INFO NET/IB : Using [0]rocep1s0f1:1/RoCE [1]roceP2p1s0f1:1/RoCE [RO]; OOB enp1s0f1np1:10.100.216.2<0>
spark-9f8f:315:401 [0] NCCL INFO NCCL_IB_GID_INDEX set by environment to 3.
spark-506c:250:250 [0] NCCL INFO NCCL_IB_HCA set to roce
spark-506c:250:250 [0] NCCL INFO NET/IB : Using [0]rocep1s0f1:1/RoCE [1]roceP2p1s0f1:1/RoCE [RO]; OOB enp1s0f1np1:10.100.216.1<0>
spark-506c:250:292 [0] NCCL INFO NCCL_IB_GID_INDEX set by environment to 3.
```
(`spark-9f8f` = head / rank 0, `spark-506c` = worker / rank 1; NCCL `2.30.7`.)

NCCL selected exactly the two members the resolver resolved and skipped the two DOWN ports itself — the mirrored contract matches the deployed universe on real hardware, not only in the fake-sysfs harness. Channels alternated across both physical links (`via NET/IB/0` ×181 / `via NET/IB/1` ×180 on rank 0, ×192 / ×192 on rank 1), zero `NCCL WARN`/`NCCL ERROR` lines, and no `/GDRDMA` (expected on this driver). Served request: `6 × 7` → `42`, `finish_reason=stop`.

**Boot B — the intended multi-HCA selector `NCCL_IB_HCA==rocep1s0f1,roceP2p1s0f1`** (22:09–22:15 CST, `BOOT_EXIT=0`). This is the configuration now in production and the boot that answers your gate:

```
Resolving RoCEv2 GID indexes from sysfs (head if=enp1s0f1np1 ip=10.100.216.2 selector==rocep1s0f1,roceP2p1s0f1; worker if=enp1s0f1np1 ip=10.100.216.1 selector==rocep1s0f1,roceP2p1s0f1)...
  member rocep1s0f1:1 -> RoCEv2 gid index 3 (via match-ip 10.100.216.2)
  member roceP2p1s0f1:1 -> RoCEv2 gid index 3 (via own-addr 10.100.217.2 on enP2p1s0f1np1)
  member rocep1s0f1:1 -> RoCEv2 gid index 3 (via match-ip 10.100.216.1)
  member roceP2p1s0f1:1 -> RoCEv2 gid index 3 (via own-addr 10.100.217.1 on enP2p1s0f1np1)
RoCEv2 GID index: head=3 (match 10.100.216.2) worker=3 (match 10.100.216.1)
```

```
spark-9f8f:317:317 [0] NCCL INFO NCCL_IB_HCA set to =rocep1s0f1,roceP2p1s0f1
spark-9f8f:317:317 [0] NCCL INFO NET/IB : Using [0]rocep1s0f1:1/RoCE [1]roceP2p1s0f1:1/RoCE [RO]; OOB enp1s0f1np1:10.100.216.2<0>
spark-9f8f:317:404 [0] NCCL INFO NCCL_IB_GID_INDEX set by environment to 3.
spark-506c:253:253 [0] NCCL INFO NCCL_IB_HCA set to =rocep1s0f1,roceP2p1s0f1
spark-506c:253:253 [0] NCCL INFO NET/IB : Using [0]rocep1s0f1:1/RoCE [1]roceP2p1s0f1:1/RoCE [RO]; OOB enp1s0f1np1:10.100.216.1<0>
spark-506c:253:296 [0] NCCL INFO NCCL_IB_GID_INDEX set by environment to 3.
```

Channels `via NET/IB/0` ×180 / `via NET/IB/1` ×180 (rank 0) and ×192 / ×192 (rank 1). Container env on both ranks: `NCCL_IB_HCA==rocep1s0f1,roceP2p1s0f1 NCCL_IB_GID_INDEX=3`. The launcher's own readiness gate and built-in smoke passed (`Minimal chat request succeeded.`), and a separate post-boot request returned `42` with `finish_reason=stop`.

Not claimed: this is a functional gate only — no performance A/B, and the second member still resolves through the own-netdev path rather than a shared match IP, which is the case the pre-revision resolver dropped. Both boots produced byte-identical member lines and the same resolved index.

sethforprivacy:
Field validation from a production deployment (2x GB10 TP=2, pinned image 3430d661, tree at a462a9e): setting a comma HCA list (`NCCL_IB_HCA=rocep1s0f0,roceP2p1s0f0`) with GID auto-resolution enabled reproduces the FATAL exactly as described. The documented workaround — `NCCL_IB_GID_AUTO=0` with `NCCL_IB_GID_INDEX=3` and `WORKER_NCCL_IB_GID_INDEX=3` (GID 3 verified `RoCE v2` via `gid_attrs/types/3` on all four HCAs, both nodes) — boots clean, passes a 29-check runtime parity gate, NIAH 3/3 at 262144, and has been serving production since. Measured upside of the dual-HCA config that this fix unblocks: +7.6-8.5% end-to-end prefill at ~224K-token prompts, decode and KV pool unchanged (details in PR #104).

Issue #84, opened 2026-08-19
perf: DSpark spec-decode A/B — official 7/greedy vs fork-default 5/probabilistic (+ boot JIT pre-warm)


plj5003:
## Hypothesis
DeepSeek's official `DeepSeek-V4-Flash-0731` HF card, the vLLM recipe (recipes.vllm.ai), and NVIDIA's `nvidia/DeepSeek-V4-Pro-nvfp4-DSpark` card all ship **`num_speculative_tokens:7, draft_sample_method:"greedy"`** for DSpark. We ship the fork default **5/probabilistic**. Community 2xSpark TP=2 measurements show DSpark ~+50% over non-spec decode and 5 > 4 acceptance — 7 is the untested side of the lever with official backing (valid: vLLM requires `num_speculative_tokens >= dspark_block_size(5)`).

Also: the compose previously hardcoded `draft_sample_method`; now env-overridable via `DSPARK_DRAFT_SAMPLE_METHOD` (default `probabilistic` — no behavior change unless set).

## Baseline (2026-08-18, current stack: 5/probabilistic, #48957 + #50298 patches live, MAX_NUM_SEQS=1)
`scripts/benchmark-0731.py` prompt=256, max_tokens=1024:
- c=1: **median 14.03 tok/s**, TTFT 0.46s, prefill ~605 tok/s
- c=4 (serial at MAX_NUM_SEQS=1): median 14.05 tok/s, TTFT queue ~110s

## Test (this PR's .env.dspark)
`MTP_NUM_TOKENS=7` + `DSPARK_DRAFT_SAMPLE_METHOD=greedy`, plus the start-script **boot JIT pre-warm** passes (max_tokens 32/128 with `thinking_token_budget=16`, non-fatal) — issue #65 mitigation that burns the thinking-kernel + MTP-verify JIT before the first user request.

## Method
Same bench cells c=1 and c=4, same box, before/after one restart. Keep 7/greedy if win; if loss, isolate axes (7/probabilistic, 5/greedy) in follow-up. Track decode tok/s + TTFT; acceptance rate if exposed.

plj5003:
Linear tracking: https://linear.app/johnsenhq/issue/JOH-7/ (project spark)

plj5003:
## Verdict: WIN — keeping 7/greedy

[...]

PR #96, opened 2026-08-20
feat(spec): make draft_sample_method configurable (default unchanged)

de1tydev:
## Issue

`docker-compose.dspark.yml` builds the speculative config with one half parameterized and the other half hardcoded:

```
SPECULATIVE_CONFIG="{\"method\":\"dspark\",\"num_speculative_tokens\":$${MTP_NUM_TOKENS:-5},\"draft_sample_method\":\"probabilistic\"}";
```

So `MTP_NUM_TOKENS=7` is a one-line `.env` change, but pairing it with the `greedy` draft sampler the official cards ship means editing compose and keeping that edit alive across every `git pull`. That is the friction in front of #84.

## Change

```
DRAFT_SAMPLE_METHOD: "${DRAFT_SAMPLE_METHOD:-probabilistic}"
```

next to `MTP_NUM_TOKENS`, and the entrypoint line uses `\"$${DRAFT_SAMPLE_METHOD:-probabilistic}\"`. **The default is the value that was hardcoded**, so an unchanged `.env.dspark` renders a byte-identical `--speculative-config`. vLLM still validates the value at engine init, so a typo fails at boot rather than silently.

Plus a commented entry in `.env.dspark.example`, a `docs/ENVS.md` row, and a CHANGELOG entry.

Naming note: #84 refers to this knob as `DSPARK_DRAFT_SAMPLE_METHOD` in a local branch. I used the unprefixed form to match its sibling `MTP_NUM_TOKENS` (both are consumed by the compose command line rather than being vLLM registry keys), but happy to rename if you prefer the `DSPARK_` prefix.

## Why we wanted it: our 7/greedy A/B came out the other way

We ran the #84 comparison on our own 2× DGX Spark GB10 cluster (TP=2, two direct RoCE links, driver `580.173.02`, Anemll `0.1.1`, official `deepseek-ai/DeepSeek-V4-Flash-0731`, deployed from checkout `f752cd0`). Same engine and serving profile across arms; code-generation workload; greedy and temp 0.6 sampling arms.

| Cell | 5 / probabilistic | 7 / greedy |
|---|---:|---:|
| Decode, greedy, 32K | **67.7 tok/s** | 59.0 tok/s (**−14%**) |
| Decode, temp 0.6 | — | flat |
| Per-position acceptance p0 / p1 | 0.90 / 0.79 | 0.85–0.89 / 0.72–0.77 |
| Positions 6–7 acceptance (code) | n/a | 0.35 / 0.21 |
| Positions 6–7 acceptance (draft-hostile text) | n/a | 0.03 / 0.00 |

Tokens per step went up ~9% at k=7, and roughly +5% per-step cost cancelled it. Two additional costs at k=7: a `max_num_scheduled_tokens` squeeze warning (8168 → 8156-class) and a larger cudagraph capture, since capture size is `MAX_NUM_SEQS * (k + 1)`. We kept 5/probabilistic.

That is the argument for the knob rather than for a new default: the lever is workload-dependent enough that people should be able to measure it on their own traffic in a restart, not a patch.

## Not claimed

- We did **not** run this exact branch (rebased on current `main`) live. The A/B numbers are from `f752cd0` with the same parameterization applied locally; only `ci-validate.sh` ran on this branch.
- The default is unchanged, so this PR by itself changes no measured behavior.
- Single-cluster result on one workload family; it does not refute a win on someone else's traffic — that is the point of making it configurable.

## Validation

`bash scripts/ci-validate.sh` on this branch: pass. Rendered `--speculative-config` checked as valid JSON for both `${DRAFT_SAMPLE_METHOD}` unset (`…"draft_sample_method":"probabilistic"`, identical to `main`) and set to `greedy` with `MTP_NUM_TOKENS=7`.

## Related

- #84 — the 7/greedy vs 5/probabilistic A/B this unblocks for other people


plotarmordev:
Thanks — exposing `draft_sample_method` is useful and the default should remain `probabilistic`, but the raw env-to-JSON boundary needs one small correctness fix before merge.

A value outside the documented enum can currently use JSON escapes or duplicate keys to become valid JSON and silently alter known fields. For example, a crafted value can override both `num_speculative_tokens` and `draft_sample_method`; `json.loads` accepts the duplicate keys, so vLLM never sees an invalid config.

Please:

1. Resolve `DRAFT_SAMPLE_METHOD` before building `SPECULATIVE_CONFIG` and accept exactly `greedy` or `probabilistic` (a small shell `case`, matching the nearby `DEFAULT_THINKING` convention, is sufficient). Print a clear error and exit nonzero for everything else.
2. Update `validate-dspark-config.sh` to apply the same contract, report the selected value instead of always printing `probabilistic`, expose the resolved Compose environment value, and reject invalid input.
3. Align the docs with the actual rejection layer.
4. Add focused CPU evidence for unset, empty, explicit `probabilistic`, `greedy`, a normal invalid value, a JSON escape alias, a duplicate-key payload, newline, and shell metacharacters. Valid defaults must preserve the old speculative-config argument exactly; invalid values must fail without execution or silent reconfiguration.

Keep the public name `DRAFT_SAMPLE_METHOD`; it matches `MTP_NUM_TOKENS`. No GPU boot or performance A/B is required for this plumbing change. After the focused matrix and full CPU CI pass, I will re-review for merge.

de1tydev:
Revised at `d6719d6` — thanks for the duplicate-key catch; the raw env-to-JSON boundary was indeed accepting it.

Point by point:

1. **Entrypoint gate**: `DRAFT_SAMPLE_METHOD` is now resolved through a shell `case` (same convention as the adjacent `DEFAULT_THINKING` gate) *before* `SPECULATIVE_CONFIG` is built. Exactly `probabilistic` or `greedy` pass; anything else prints `DRAFT_SAMPLE_METHOD must be one of: probabilistic, greedy (got: …)` and exits 2, so no crafted value reaches the JSON.
2. **`validate-dspark-config.sh`**: applies the same contract right after sourcing the env file (exit 2 on invalid), reports the *resolved* value in the summary line instead of the old hardcoded `probabilistic`, and `DRAFT_SAMPLE_METHOD` was added to the rendered-config grep so the resolved Compose environment value is visible.
3. **Docs**: `.env.dspark.example`, `docs/ENVS.md`, and the CHANGELOG entry now state the actual rejection layer (entrypoint + validate script, fail-fast before vLLM) instead of "vLLM rejects at engine init".
4. **Focused CPU evidence**: new `scripts/test-draft-sample-method-gate.sh` (wired into `ci-validate.sh`) extracts the gate + `SPECULATIVE_CONFIG` assignment *from the compose file itself* (no copy of the logic) and runs the matrix:
   - unset / empty / explicit `probabilistic` → byte-identical to the previously hardcoded JSON `{"method":"dspark","num_speculative_tokens":5,"draft_sample_method":"probabilistic"}`
   - `greedy` → the greedy variant, same shape
   - `random`, `Greedy`, `gree\u0064y` (JSON escape alias for `greedy`), `probabilistic","num_speculative_tokens":9999,"draft_sample_method":"greedy` (duplicate-key payload), embedded newline, shell metacharacters → all exit nonzero with no config built
   - a `$(touch canary)` payload is rejected **and** the canary is verified absent, i.e. nothing embedded ever executes

Result: 12/12 in the focused suite; full `bash scripts/ci-validate.sh` passes on the head (`CI validate passed (CPU recipe gates only)`), including the new gate.

Public name kept as `DRAFT_SAMPLE_METHOD`; default and rendered argument unchanged for valid inputs. No GPU run was done per your note.

plotarmordev:
Thanks — re-review of exact head `d6719d6c978bcaca3310d34966b9b92aac954c4a` confirms the source fixes we requested: strict `probabilistic|greedy` validation precedes JSON construction, injection/duplicate-key inputs are rejected, unset/empty preserve the old default, and Compose/validator/docs are wired consistently.

Remaining before merge:
1. Rebase onto current `main` `21da90f1d2000661618fdb21dcd70d0afdddc7bf`; the current head is dirty with a `CHANGELOG.md` conflict.
2. Replace source-string greps for `validate-dspark-config.sh` with executable behavior coverage of accepted/rejected values.
3. Exercise the actual rendered Compose expansion/entrypoint contract rather than only an extracted fragment, including unset, empty, probabilistic, greedy, invalid enum, quotes, backslashes, and newlines.
4. Require hosted `validate` success on the conflict-resolved exact head.

No GPU result is required to merge a default-preserving configurability knob. The 5/probabilistic vs 7/greedy performance decision remains issue #84's later A/B, not a claim of this PR.

de1tydev:
Revised at `21a07fc94a4845a1d2775d55e60f8d1e3f072838`. Point by point:

1. **Rebased.** On current `main` `5d5a00d81539a2f3708718fbe185a61dd8c387cc` (two commits past the `21da90f1` you cited). The `CHANGELOG.md` conflict is resolved by keeping main's 2026-08-20 `### Changed` and `### Fixed` entries as-is and placing this PR's `### Added` entry above them.

2. **Validator is executed, not grepped.** The two source-string greps are gone. `validate-dspark-config.sh` now runs against stub env files and is asserted on behavior: unset, empty and explicit `probabilistic` all resolve to `probabilistic`; `greedy` is reported as `greedy` (the old hardcoded-summary bug would fail this); invalid enum, case variant and the duplicate-key payload exit 2 with the contract message and print no summary line at all. The `docker compose config` call the validator ends with is not what these cases are about, so they assert on the resolved-value summary — printed before it — and on rc ≠ 2 rather than requiring a full clean run.

3. **The rendered Compose contract is exercised.** A third layer writes a real `.env`, renders the stack with `docker compose config`, pulls the gate out of the *rendered* entrypoint and runs it under the *rendered* environment map, covering unset, empty, `probabilistic`, `greedy`, invalid enum, quotes, backslashes and newlines.

   That layer earned its keep immediately: Compose applies its own escape processing, so a double-quoted `DRAFT_SAMPLE_METHOD="probabilistic\ngreedy"` in `.env` reaches the entrypoint as a genuine two-line value — something the sed-extracted fragment never saw, because the fragment layer is handed the value directly. It is rejected, as are `'greedy'`, `'gre"edy'`, `gree\dy` and the duplicate-key payload. Quoted-but-valid (`"greedy"`) still resolves to `greedy`.

4. **Hosted gate.** `validate` passes on this exact SHA; `bash scripts/ci-validate.sh` passes locally.

**Coverage is not vacuous**: run against a compose whose `case` accepts anything, the suite fails 13/29 — including the duplicate-key input rendering exactly the attack config `{"method":"dspark","num_speculative_tokens":5,"draft_sample_method":"probabilistic","num_speculative_tokens":9999,"draft_sample_method":"greedy"}`. On the shipped gate it is 29/29.

Default and rendered argument unchanged for valid inputs; the 5/probabilistic vs 7/greedy question stays with #84.

plotarmordev:
Thanks @de1tydev — your three PR #96 commits and authorship reached `main` unchanged through the completed maintainer PR #102.

The follow-up only repaired the PR-owned evidence harness after current-main integration: Layer 2/3 now isolate caller auth-free sampler variables, accepted validator cases require exact success through a deterministic stub, the resolved-value diagnostic is factual, and the suite permanently poisons its own parent environment as a negative control. No production, default, docs, or runtime behavior from your implementation was changed.

PR #102 merged normally as `6d00e4a3f149d9121c69bd1da23a9b6353c5164d`; hosted validation passed on both its exact head and merged `main`. GitHub correctly marked this original PR merged when your exact commits reached `main`.

Thank you for the focused, default-preserving configurability work. Any later 5/probabilistic versus 7/greedy performance experiment remains tracked separately in issue #84.

Issue #93, opened 2026-08-20
ci: restore hosted validation workflow referenced by ci-validate.sh


plotarmordev:
## Problem

The repository currently has no `.github/workflows/validate.yml` (and no `.github` tree), but:

- `scripts/ci-validate.sh:2` says it is the same script as `.github/workflows/validate.yml`.
- `CHANGELOG.md` also refers to the GitHub Actions validate workflow.
- PR #91 therefore had no hosted checks. Its exact-head local equivalent, `bash scripts/ci-validate.sh`, passed before merge and was repeated against tree-identical merged content, but unavailable hosted validation must not be treated as green.

## Follow-up obligation

Re-run hosted validation against merged main commit `625230391950f2d80c850f9af8901de99f557760` as soon as a repository workflow exists.

## Suggested resolution

Choose and make one policy true:

1. Add `.github/workflows/validate.yml` for pull requests and pushes to `main`, running `bash scripts/ci-validate.sh` on the supported Python version; or
2. If hosted CI is intentionally unsupported, remove the stale workflow claims and document the local validation requirement explicitly.

## Acceptance

- Repository text and actual CI configuration agree.
- A hosted run validates the PR #91 merged tree, or the repository explicitly records that validation is local-only.
- No paid plan or visibility change is required.

plotarmordev:
Follow-up obligation recorded for PR #97: merged normally as `e4e5056691969ab8deb773306ee39f1da4c3bd02` without hosted checks because this workflow is still absent. Exact reviewed head `dd84f9fe28696f61e070227e31c6d3d82bec502b` passed the full local `scripts/ci-validate.sh` gate and an exact-script two-node warm-cache preparation smoke; the merge tree is byte-identical to the reviewed head tree (`4e39166d01ac7f67806c1a52c21bbaec032fa8d1`). Re-run hosted validation against merged commit `e4e5056` as soon as the workflow is restored.

PR #97, opened 2026-08-20
fix(prepare): pass HF hub timeouts into the download containers

de1tydev:
## Issue

`prepare-dspark-model-cache.sh` runs `snapshot_download` inside a throwaway container and forwards a curated set of `-e HF_*` variables. Two are missing:

- `HF_HUB_DOWNLOAD_TIMEOUT` — per-chunk read timeout on the file transfer
- `HF_HUB_ETAG_TIMEOUT` — timeout on the metadata/HEAD request

`huggingface_hub` defaults both to **10 seconds**. DeepSeek-V4-Flash-0731 is many multi-GB safetensors shards, and on a slow or proxied link a single stall past 10s aborts the transfer, so cache prep dies partway through a shard rather than riding it out. Because the values are not in the `-e` list, setting them in `.env.dspark` (which the script already sources) has no effect — they never reach the container.

## Change

Two lines in each of the two `docker run` blocks, after the existing `-e HF_HUB_DISABLE_XET=…`:

```sh
    -e HF_HUB_DOWNLOAD_TIMEOUT="${HF_HUB_DOWNLOAD_TIMEOUT:-120}" \
    -e HF_HUB_ETAG_TIMEOUT="${HF_HUB_ETAG_TIMEOUT:-30}" \
```

in `run_download` and in `verify_cache`. `verify_cache` runs `local_files_only=True` with `HF_HUB_OFFLINE=1`, so the values are inert there — they are included only so the two blocks stay identical and no one has to reason about which one got the fix.

Defaults are 120s/30s rather than the hub's 10s/10s. These are timeouts, not retries: a healthy link is unaffected, and a genuinely dead link still fails, just not on a transient 10-second stall. Override either in `.env.dspark`.

Plus commented entries in `.env.dspark.example` and a CHANGELOG line. No retry/backoff logic, no other behavior change.

## Evidence

Motivated by cache prep on our own 2× DGX Spark GB10 cluster (Anemll `0.1.1`, official `deepseek-ai/DeepSeek-V4-Flash-0731`, from checkout `f752cd0` with this diff applied locally), where the stock 10s timeouts aborted large-shard downloads over a proxied link and the raised values let the same prep complete.

**Not claimed:** this is not a fix for a broken network, and we did not run this exact branch (rebased on current `main`) — only `ci-validate.sh` ran on the branch.

## Validation

`bash scripts/ci-validate.sh` on this branch: pass (includes `bash -n prepare-dspark-model-cache.sh`).

### DIFF
--- a/.env.dspark.example
+++ b/.env.dspark.example
@@ -64,6 +64,10 @@ WORKER_HF_CACHE=
 HF_HUB_OFFLINE=1
 TRANSFORMERS_OFFLINE=1
 HF_HUB_DISABLE_XET=1
+# Hub HTTP timeouts used by prepare-dspark-model-cache.sh. hf_hub defaults to
+# 10s/10s, which aborts multi-GB shard downloads on a slow or proxied link.
+# HF_HUB_DOWNLOAD_TIMEOUT=120
+# HF_HUB_ETAG_TIMEOUT=30
 # Checkpoint flag: 0 = official 0731, 1 = abliterated (Keys).
 # start- / prepare- resolve DSPARK_MODEL from this — do not set DSPARK_MODEL by hand.
 ABLITERATED=0
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+## 2026-08-20
+
+### Fixed
+
+- **Hub timeouts abort large-shard downloads in `prepare-dspark-model-cache.sh`**: both `docker run` blocks (`run_download` and `verify_cache`) now pass `HF_HUB_DOWNLOAD_TIMEOUT` (default `120`) and `HF_HUB_ETAG_TIMEOUT` (default `30`). `huggingface_hub` defaults both to 10s, which is short enough that a slow or proxied link kills a multi-GB shard mid-transfer rather than riding it out. Override in `.env.dspark`.
+
 ## 2026-08-19
 
 ### Changed
--- a/prepare-dspark-model-cache.sh
+++ b/prepare-dspark-model-cache.sh
@@ -197,6 +197,8 @@ run_download() {
     -e HF_HUB_OFFLINE=0 \
     -e TRANSFORMERS_OFFLINE=0 \
     -e HF_HUB_DISABLE_XET="${HF_HUB_DISABLE_XET:-1}" \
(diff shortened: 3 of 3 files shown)


plotarmordev:
Exact-head maintainer validation completed for `dd84f9fe28696f61e070227e31c6d3d82bec502b`.

**CPU gate — PASS**
- `bash scripts/ci-validate.sh`
- shell syntax: all listed launch/prepare/patch scripts passed
- Python compile: 20 files passed
- all unit suites passed
- DSv4 equality/dormancy gate: 24 passed, 0 failed
- overlay-source and recipe regression guards passed
- final result: `CI validate passed (CPU recipe gates only)`

**Exact-head two-node warm-cache smoke — PASS**
- command path: `./prepare-dspark-model-cache.sh --official`
- isolated temporary profile; production profiles and deployment tree were not changed
- exact script SHA-256 on review worktree, MIDDLE, and BOTTOM: `ac3401192dc2f8e91a3069423fb02b548ab464d99dfb143974e2109d8cd59e34`
- timeout values exercised from `.env.dspark`: download `120`, etag `30`
- MIDDLE: revision `9e165c30e2704aec5d9d593cce3eebd58bbef1cb`, 48 safetensor shards, `missing_shards=0`
- BOTTOM: same revision, 48 safetensor shards, `missing_shards=0`
- optional VL cache preparation remained disabled
- existing serving endpoint remained healthy (`/v1/models` → HTTP 200) after the smoke

**Admission checks — PASS**
- exact diff: 3 files, 14 insertions (`.env.dspark.example`, `CHANGELOG.md`, `prepare-dspark-model-cache.sh`)
- `git diff --check`: clean
- changed range and tracked worktree secret scan: 0 credential/private-key hits
- PR remains OPEN, non-draft, MERGEABLE at the exact reviewed head
- independent DSV4PRO and Sol xhigh reviews found no implementation blocker after these gates

No cold-cache redownload, forced network fault, serving restart, or GPU test was run; none is required for this prepare-only environment passthrough. Hosted checks are absent, so the exact-head local equivalent above is the merge evidence; hosted follow-up remains tracked by #93.

Result: approved for normal merge at the exact reviewed head.

PR #98, opened 2026-08-20
fix(start): harden .env parsing and worker token permissions

hecisaza:
Three fixes found while running this recipe on a 2x DGX Spark (GB10) TP=2 pair. Reproduced against `5d5a00d`, then re-verified with the patch.

## 1. A UTF-8 BOM or CRLF in `.env.dspark` breaks the launch — sometimes silently

`.env.dspark` is `source`d, so anything an editor leaves in the file is executed. Three behaviours, none of which name the cause:

| file | result |
|---|---|
| CRLF **with a blank line** (the shipped example has many) | `rc=127` — `$'\r': command not found` |
| CRLF **without** blank lines | **`rc=0`, silent** — values carry a trailing `\r` |
| UTF-8 BOM | `rc=2` — `syntax error near unexpected token '('` (message varies with the first comment's contents) |

The silent case is the dangerous one:

```console
$ printf 'WORKER_HOST=10.0.0.2\r\nVLLM_PORT=8888\r\n' > .env.dspark
$ ./start-deepseek-v4-flash-dspark.sh
# rc=0, but WORKER_HOST is now $'10.0.0.2\r'
```

`ssh` and the `NCCL_*` interface names then fail far from the env file. (`VLLM_PORT` is caught by the integer guard, but its message — `must be an integer ...: 8888` — reads as if the value were fine, because the `\r` is invisible.)

**Fix:** normalise a temp copy before sourcing. The operator's file is never rewritten. The copy holds the secrets, so the cleanup trap is armed immediately after `mktemp` — before `chmod`, so no window exists — and fires on every exit path including `sed`/`source` failure and signals.

## 2. The env file holds `HF_TOKEN` and reaches the worker world-readable

`cp .env.dspark.example .env.dspark` (per the README) inherits the umask — typically `644` — and the file is copied to the worker on every start.

**Fix:** the worker now receives the **same normalised bytes**, so its `compose --env-file` does not see a BOM/CRLF the head was shielded from. `umask 077` keeps a newly created file unreadable from the start, and an explicit `chmod` enforces `0600` on a destination that **already exists** — umask does not apply to an existing inode, so a stale world-readable worker copy would otherwise survive across restarts. `REMOTE_ENV_FILE` is already `%q`-escaped, so it is used unquoted like the other remote paths. No `|| true`: if `0600` cannot be enforced, `set -e` aborts the launch.

## 3. Banner prints a `max num seqs` default that isn't the launch default

`${MAX_NUM_SEQS:-12}` vs the actual `6` (`docker-compose.dspark.yml`; `:433` already uses `:-6`). With the variable unset the banner reported double the real value.

## Tests

`scripts/test-env-normalisation.py` — 15 CPU-only tests, wired into `scripts/ci-validate.sh`. The launcher block is lifted verbatim via an extractor so the tests cannot drift from the shipped code. Covers BOM, CRLF, combined BOM+CRLF, operator-file immutability, cleanup on source failure and on signal, remote umask, `0600` on a pre-existing destination, normalised bytes reaching the worker, fail-closed remote write, and an escaped remote path.

Verified by mutation: reverting any of the fixes fails the suite with a diagnostic. `bash -n` clean; `scripts/ci-validate.sh` passes.

Tested on: 2x DGX Spark (GB10, sm_121), DGX OS 26.04.1, TP=2 over RoCE, `ghcr.io/anemll/dspark-vllm-gx10:0.1.1`, DeepSeek-V4-Flash-0731.

### DIFF
--- a/scripts/ci-validate.sh
+++ b/scripts/ci-validate.sh
@@ -39,6 +39,7 @@ py_files+=(
   scripts/spec-acceptance.py
   scripts/test-spec-acceptance.py
   scripts/test-ruler-lite-pad.py
+  scripts/test-env-normalisation.py
   scripts/ruler-lite.py
   scripts/verify-dsv4-027-equality-gate.py
 )
@@ -64,6 +65,8 @@ python3 scripts/test-spec-acceptance.py -q
 ok "test-spec-acceptance"
 python3 scripts/test-ruler-lite-pad.py -q
 ok "test-ruler-lite-pad"
+python3 scripts/test-env-normalisation.py -q
+ok "test-env-normalisation"
 python3 tests/test_issue27_inflight_cap.py -q
 ok "test_issue27_inflight_cap"
 python3 scripts/verify-dsv4-027-equality-gate.py
--- a/scripts/test-env-normalisation.py
+++ b/scripts/test-env-normalisation.py
@@ -1,0 +1,273 @@
+#!/usr/bin/env python3
+"""Unit tests for the .env.dspark normalisation and worker-copy hardening in
+start-deepseek-v4-flash-dspark.sh.
+
+CPU-only; no GPU, no network, no real worker. The env-loading block and the
+remote-write line are lifted verbatim from the launcher so the tests exercise
+the shipped code rather than a copy of it.
+
+    python3 scripts/test-env-normalisation.py -q
+"""
+import os
+import re
+import shutil
+import subprocess
+import sys
+import tempfile
+import unittest
+
+ROOT = os.path.dirname(os.path.dirname(os.path.abspath(__file__)))
+LAUNCHER = os.path.join(ROOT, "start-deepseek-v4-flash-dspark.sh")
+
+
+def _extract(start_marker: str, end_marker: str) -> str:
+    """Pull a verbatim block out of the launcher so tests can't drift from it."""
+    with open(LAUNCHER, encoding="utf-8") as fh:
+        src = fh.read()
+    i = src.index(start_marker)
+    j = src.index(end_marker, i) + len(end_marker)
+    return src[i:j]
+
+
+ENV_BLOCK = _extract('_dspark_env_clean="$(mktemp)"', "trap - EXIT HUP INT TERM")
+def _extract_remote_write() -> str:
+    """The worker-copy pipeline: the `sed` that feeds the ssh write.
+
+    Anchored on the ssh invocation (stable) and bounded by the end of its logical
+    line, so changing the *mode* or the redirect target still yields a runnable
+    block and produces a clean test failure rather than an extraction crash.
+    """
+    with open(LAUNCHER, encoding="utf-8") as fh:
+        src = fh.read()
+    j = src.index('| ssh "$WORKER_HOST" "umask 077;')
+    i = src.rindex("sed $'", 0, j)
+    end = src.index("\n", j)
+    return src[i:end]
+
+
+REMOTE_WRITE = _extract_remote_write()
+
+
+def run_env_block(content: bytes, extra: str = "") -> subprocess.CompletedProcess:
+    """Source `content` through the launcher's normalisation block."""
+    d = tempfile.mkdtemp()
+    env_file = os.path.join(d, ".env.dspark")
+    with open(env_file, "wb") as f:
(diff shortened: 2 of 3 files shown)


plotarmordev:
Thanks — the BOM/CRLF transformation and `MAX_NUM_SEQS` banner correction are directionally correct, but exact head `174de4dc0ce0d9302c818d9bbc9bce293c0850ea` is not ready yet.

Required fixes:
1. Make the secret-bearing normalized temp file cleanup failure-safe. It currently survives `sed`/`source` failure and catchable termination paths. Use the existing trap structure (or an equally small mechanism) without masking the original error.
2. Make remote credential permissions fail closed. Remove `|| true`; startup must not continue if mode 0600 cannot be enforced.
3. Fix remote path handling. `REMOTE_ENV_FILE` is already `%q`-escaped; wrapping that escaped token in literal single quotes addresses the wrong path for spaces/quotes.
4. Remove/correct the statement that this `scp` preserves source mode: the command does not use `-p`, and a post-copy chmod leaves a possible exposure window.
5. Add focused CPU/shell coverage for BOM, CRLF, combined BOM+CRLF, operator-file immutability, cleanup on source failure, remote path quoting, and chmod failure. Syntax-only CI cannot prove these behaviors.
6. Rebase if `main` moves and require hosted `validate` success on the exact revised head.

Please keep the correction small and contributor-owned. The launcher overlaps PR #95, so whichever lands second must rebase and preserve both disjoint hunk sets.

hecisaza:
Thanks — all six addressed, rebased onto `5d5a00d`. Head is `5fc02ce`.

1. **Cleanup** — trap armed immediately after `mktemp` (before `chmod`, so there is no window), fires on `EXIT HUP INT TERM`, and is cleared after use so it does not mask `on_error`. Tested against source failure and signal.

2. **Fail closed** — `|| true` gone. Also caught a gap in my first attempt: `umask 077` alone does **not** fix a destination that already exists, since umask does not apply to an existing inode. A stale `0644` worker `.env.dspark` stayed `0644` while the launch continued. Now an explicit `chmod` enforces `0600` in both cases, with a regression test.

3. **Path** — `REMOTE_ENV_FILE` used unquoted, matching the existing `mkdir -p $REMOTE_WORKER_DIR` pattern. Tested with a `%q`-escaped path containing a space.

4. **scp claim** — removed, and the env copy no longer uses `scp` at all. For the record: I measured `scp` without `-p` on two independent hosts and it *did* preserve source mode (644→644, 600→600), so that specific reasoning did not reproduce here — but your conclusion held, since a post-copy `chmod` leaves an exposure window regardless. The `umask 077` + `chmod` pipeline makes it moot either way.

5. **Tests** — 15 CPU-only tests in `scripts/test-env-normalisation.py`, wired into `ci-validate.sh`. The launcher block is extracted verbatim so the tests gate shipped code; mutating any of the fixes fails the suite with a diagnostic rather than passing quietly.

6. **Rebase** — on `5d5a00d`; `scripts/ci-validate.sh` passes.

One in-scope extension worth flagging: the worker now receives the normalised bytes rather than the raw file, so `compose --env-file` on the worker sees what the head sourced. Sister scripts (`stop-`, `smoke-`, `status-`, `prepare-`) still `source "$ENV_FILE"` unnormalised — happy to leave that for a separate PR to keep this one small, or fold it in if you would prefer.

Still disjoint from #95 (that touches the GID resolver); I simulated the merge in both orders and `start.sh` and `ci-validate.sh` auto-merge cleanly.

plotarmordev:
Thanks @hecisaza — your PR identified and initially implemented the BOM/CRLF normalization, worker credential hardening, and `MAX_NUM_SEQS=6` banner correction.

We preserved your exact commit `5fc02ce0c39362578fff75a01301a05168b545f4` and authorship in the completed maintainer PR #101, then added the failure-safe cleanup, single normalized snapshot, private atomic worker publication, and focused regression coverage requested during review.

PR #101 merged normally as `9cce187ff596b8bc90faa751010bf60542f93da0`; hosted validation passed on both the exact PR head and merged `main`. GitHub correctly marked this original PR merged when your unchanged commit reached `main`.

Durable credit is preserved in commit history, the PR #101 body, and CHANGELOG. Thank you for the contribution.

PR #104, opened 2026-08-21
docs(fabric): measured dual-HCA + MTU 9000 guidance for the NCCL passthrough

sethforprivacy:
PR #94 added the NCCL fabric passthrough with deliberately no tuning claim. This fills in the measurement, from two independent 2x GB10 TP=2 pairs (direct QSFP, no switch), so the knobs ship with evidence:

**Method.** ~224K-token fresh-random prompts, 3-4 reps per config, cluster drained from serving during probes, first rep discarded (cold-JIT). Decode checked separately via engine-counter probes (acceptance / tok-per-step / steps-per-s), NIAH at 262144 for correctness after each change.

| config | pair 1 | pair 2 |
|---|---|---|
| MTU 1500, single HCA | 1311 tok/s | — |
| MTU 9000, single HCA | 1343 tok/s | — |
| MTU 9000 + both controllers in NCCL_IB_HCA | **1410 tok/s** | **1422 tok/s** |

**Pair 1 prefill improved 7.6% end-to-end** (1311 → 1410 tok/s). Pair 2 reproduced a **1422 tok/s dual-HCA final rate** but has no same-pair single-HCA baseline, so it is not used for an uplift percentage. Separate engine-counter probes found no material decode change on the measured stack (code 76.0 tok/s @ 73.3% accept, prose 37.3 @ 25.7%, 14-15 steps/s); **KV pool unchanged**. Merged controllers are one logical NIC, unlike unmerged dual-HCA channel splitting, which cost KV buffer memory for ~nothing (+3%, measured 2026-08-06). nccl-tests busbw increased 64% on this stack, but all-reduce is only a slice of prefill step time, so that number is not an end-to-end throughput claim.

Operational prerequisites: use the deterministic exact selector (`NCCL_IB_HCA==dev0,dev1`), verify both controller links are PCIe x4 (early BIOS wires the second controller Gen5 x2), and use MTU 9000 only when both endpoints and every intervening L2 hop support it. These measurements used an explicit index-3 GID pin because the pre-fix multi-HCA resolver FATALed; that deployment field-validates #95's failure mode and older-revision workaround. PR #95 is now merged: current main should keep `NCCL_IB_GID_AUTO=1` and verify the startup member/index audit. The manual pin is fallback guidance for older revisions, not the current default.

All measurements from the pinned image (`3430d661`) at a462a9e.

### DIFF
--- a/.env.dspark.example
+++ b/.env.dspark.example
@@ -314,26 +314,41 @@ NCCL_NVLS_ENABLE=0
 # MEASURED (2026-08-21, two independent 2x GB10 TP=2 pairs, direct QSFP):
 # the GB10 QSFP port enumerates as TWO virtual NICs (2x PCIe Gen5 x4
 # controllers, ~100G each). With only one HCA in NCCL_IB_HCA the link runs
-# at half the port. Listing BOTH controllers (merging is NCCL's default):
+# at half the port. List both controllers with the deterministic exact-selector
+# form (merging is NCCL's default):
 #
-#   NCCL_IB_HCA=rocep1s0f0,roceP2p1s0f0        # map names: ibdev2netdev
+#   NCCL_IB_HCA==rocep1s0f0,roceP2p1s0f0       # map names: ibdev2netdev
 #
-# plus MTU 9000 on both interfaces (host netplan; RoCE derives its IB MTU
-# from the netdev MTU, so the stock 1500 caps frame size) measured, at
-# ~224K-token prompts, prefill 1311 -> 1343 tok/s from MTU alone and
-# -> 1410-1422 tok/s with both HCAs (+7.6-8.5% end-to-end). Decode was
-# unchanged on the same runs (latency-bound all-reduces; no GDR on GB10),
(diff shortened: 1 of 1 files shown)


sethforprivacy:
Concurrency/stability addendum, since the table above is single-stream on a drained pair: we re-ran the same config under load — 8 concurrent ~168K-token prompts spread across two independent 2x pairs on this exact fabric setup, stacked on top of live production traffic (15 additional requests in-window).

Per pair, at `max_num_seqs=4` / `max_num_batched_tokens=8192` / chunked prefill with `LONG_PREFILL_TOKEN_THRESHOLD=1024`: one pair sustained **~3,600 tok/s aggregate prefill** at full 4-way slot occupancy for the duration (the second pair ~2,300 with a lighter share). Every request returned an exact needle retrieval, **zero preemptions on either pair**, and the hottest node stayed at 92.5 °C across ~14 minutes of saturation. Not a controlled benchmark (co-load varies) — offered as evidence the dual-HCA + MTU 9000 config is stable under sustained concurrent long-context load, not just in isolated probes.

PR #106, opened 2026-08-22
fix(scheduler): cache validated partial-prefill cap

### LINKED ISSUE #105: [Bug] Malformed DSPARK_MAX_INFLIGHT_PREFILLS crashes scheduler admission after startup
## Summary

`patches/hotfix-dsv4-issue27-partial-prefill-concurrency.py` injects a bare `int(...)` conversion into the `Scheduler.schedule()` waiting-admission loop. A malformed `DSPARK_MAX_INFLIGHT_PREFILLS` value survives container startup and raises `ValueError` only when a waiting request reaches admission.

Current `main`: `ed9ef7084291e6f496b74a3c1633b6fa527ce54d`.

## Current injected code

```python
_pp_cap = int((__import__('os').environ.get(
    'DSPARK_MAX_INFLIGHT_PREFILLS') or '0') or 0)
```

Compose forwards the variable as arbitrary text and defaults it to `2`.

## Deterministic behavior

| Input | Current result |
|---|---|
| unset / empty | scheduler-config fallback |
| whitespace | `ValueError` |
| `0` / negative | scheduler-config fallback |
| `1`, `2`, `3` | accepted |
| `4` | clamped to `3` |
| `two`, `2.0`, `1x` | `ValueError` |

The malformed value remains in the container environment, so a restart does not repair the configuration. The default `2` is safe; this is an operator-triggered availability defect rather than a default-path failure.

## Expected fix

- Parse and validate once during scheduler construction, not inside `schedule()`.
- Cache the effective integer on the scheduler.
- Preserve the current safe policy: blank/nonpositive values use `SchedulerConfig.max_num_partial_prefills`; values above `3` clamp to `3`.
- Malformed values should warn once and use the conservative scheduler-config fallback rather than crashing service admission.
- Test unset, empty, whitespace, `0`, negative, `1`–`4`, malformed values, and prove the environment is not reread by `schedule()`.

## Existing test gap

`tests/test_issue27_inflight_cap.py` currently checks source strings and patch idempotence only. Its two tests pass while the malformed-value crash remains.

## Provenance and limits

The override was added by #90 on top of the #27 admission hotfix. The failure is source- and CPU-reproduction-confirmed. The live TP=2 service was not deliberately restarted with a malformed value because it is currently in use.

plotarmordev:
Closes #105.

## Problem

The #27 hotfix parsed `DSPARK_MAX_INFLIGHT_PREFILLS` with bare `int(...)` inside every `Scheduler.schedule()` waiting-admission iteration. Malformed values survived startup, then raised `ValueError` when traffic reached the waiting queue. Restarting preserved the malformed environment value.

## Fix

- Parse once during `Scheduler` construction and cache `_dspark_max_inflight_prefills`.
- Preserve existing semantics: blank/nonpositive values use `SchedulerConfig.max_num_partial_prefills`; values above 3 clamp to 3.
- Malformed values warn once and use the scheduler-config fallback.
- Remove all environment reads and conversions from the scheduling hot loop.
- Replace source-string assertions with behavioral parser, fallback, clamp, admission, cache, and idempotence coverage.
- Document the operator-visible fallback.

## Verification

- `python3 tests/test_issue27_inflight_cap.py -q` — 6/6 pass.
- `bash scripts/ci-validate.sh` — pass, including 23/23 atomic-hotfix tests, 72/72 NCCL selector tests, and all CPU recipe gates.
- Exact pinned image `ghcr.io/anemll/dspark-vllm-gx10:0.1.1`: applied Issue #27 then Issue #43 hotfixes, compiled the resulting scheduler, and verified one cached-cap assignment with no environment read/conversion in `schedule()`.
- `git diff --check origin/main` — pass.
- Changed-file secret-pattern scan — no matches.

## Live validation status

Not run yet. The local TP=2 model is currently in use and was not restarted or otherwise touched. Keep this PR draft until a controlled malformed-value boot and restoration can be performed without interrupting that workload.

## Scope

No Compose default or scheduling policy change. Default `DSPARK_MAX_INFLIGHT_PREFILLS=2` remains unchanged. This does not address the separate non-fail-closed semicolon chain for default-on Python patchers.

### DIFF
--- a/.env.dspark.example
+++ b/.env.dspark.example
@@ -195,11 +195,15 @@ LONG_PREFILL_TOKEN_THRESHOLD=1024
 # decode rate #43 reports. NOTE: --max-num-partial-prefills is NOT available
 # on dspark-vllm-gx10:0.1.1 — vLLM's _check_feature_supported() rejects
 # Concurrent Partial Prefill before engine init. Do NOT pass that flag.
-# The #27 hotfix instead reads DSPARK_MAX_INFLIGHT_PREFILLS (clamped 1-3).
-# Default 2: live 32K×c4 decode left the ~8 tok/s floor (~25 tok/s) with
-# #43 floor on and LONG_PREFILL_TOKEN_THRESHOLD=1024. Set 1 to restore the
-# original serialize-one-prefill gate. Image upgrade for real concurrent
-# partial prefill is issue #45.
+# The #27 hotfix instead reads DSPARK_MAX_INFLIGHT_PREFILLS once during
+# Scheduler construction. Values 1-3 are accepted; values above 3 clamp to 3.
+# Blank, whitespace-only, nonpositive, or malformed values fall back to
+# SchedulerConfig.max_num_partial_prefills (stock 1); malformed values warn
+# once instead of crashing admission. Default 2: live 32K×c4 decode left the
+# ~8 tok/s floor (~25 tok/s) with #43 floor on and
+# LONG_PREFILL_TOKEN_THRESHOLD=1024. Set 1 to restore the original
+# serialize-one-prefill gate. Image upgrade for real concurrent partial
+# prefill is issue #45.
 DSPARK_MAX_INFLIGHT_PREFILLS=2
 
 # Issue #43: bounded decode service during mixed prefill steps + scheduler
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,8 @@
 
 ### Fixed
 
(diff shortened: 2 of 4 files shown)


plotarmordev:
Final live validation passed on the exact PR head `f62be5255fdbfa80d01f788b2f93b01fa7d20e66`, composed with merged prerequisite #108, exact O1 head `ebe4ca2145702744b0e3a09b04153eb163dfea41`, and the local operational-only layer.

- Complete combined CPU CI passed.
- TP=2 base model cold-started successfully on the pinned Anemll 0.1.1 image; both ranks joined NCCL and the API became ready.
- Both ranks had `DSPARK_MAX_INFLIGHT_PREFILLS=2` and the cached constructor/admission source installed.
- Live 4-lane cold wave: 22,817 prompt tokens/lane, 91,268 prompt tokens total, 1,024 generated tokens, 0 preemptions, 100% MTP acceptance, worst per-lane decode 19.59 tok/s.
- Six-way chat smoke passed 6/6.

No remaining D2 defect observed. Ready for normal merge.

plotarmordev:
Final live validation complete on the real TP=2 DeepSeek-V4-Flash-0731 deployment (image ghcr.io/anemll/dspark-vllm-gx10:0.1.1). Both ranks received DSPARK_MAX_INFLIGHT_PREFILLS=2 and contain the constructor-cached cap patch. A 4-lane cold concurrent wave processed 91,268 prompt tokens and 1,024 generation tokens with 0 preemptions; worst observed decode rate was 19.59 tok/s and MTP accepted 860/860 drafted tokens. Combined PR #106 + PR #110 local CI passed, and a post-wave API smoke returned the exact requested response. Ready to merge.

PR #110, opened 2026-08-22
fix: remove phantom encoder output tokens

### LINKED ISSUE #109: [Bug] EC producer emits phantom token 0 for encoder-only steps
## Problem

`ModelRunnerOutput.make_empty_encoder_model_runner_output()` says an encoder-only EC producer step generated no tokens, but returns one `[0]` row per request. The scheduler treats each row as a real generated token: it can append token ID 0, advance grammar/stop state, count output, and distort speculative-decoding accounting.

The same pinned/overlay scheduler lineage lacks upstream's paired encoder-only completion branch. Returning honest empty rows without that branch would leave a completed encoder request producing empty steps indefinitely.

The repository overlays and pinned runtime image `ghcr.io/anemll/dspark-vllm-gx10:0.1.1` contain both gaps. Upstream vLLM commit `7ca49fbe4bab019e55d57cdc4b7fd3d55c67c1a6` returns one empty list per request and stops an encoder-only request once its prompt is consumed.

## Expected

- return `[[] for _ in req_ids]`: preserve one row per request without fabricating output
- finish an encoder-only request after all prompt tokens are consumed

## Scope

- fix both source overlays
- add one fail-closed, idempotent pinned-image patcher and Compose wiring
- synchronize the patcher to the worker
- add CPU behavior/drift/idempotence/completion coverage and exact pinned-image composition verification
- document upstream provenance

## Safety

Feature-gated to EC transfer / encoder-only models, but correctness-impacting when enabled. Offline validation only; do not restart or touch the active TP=2 model.

plotarmordev:
Closes #109.

Depends on merged prerequisite #108, which made default-on Python source patchers fail closed before `vllm serve`.

## Built
- return one distinct empty sampled-token row per scheduled request on encoder-only EC producer steps instead of fabricating token ID 0
- add the paired encoder-only completion branch: stop once every prompt token is consumed, avoiding an empty-output loop
- apply both corrections in the source overlays
- add one exact-anchor, idempotent, drift-refusing two-file patcher for pinned image `ghcr.io/anemll/dspark-vllm-gx10:0.1.1`
- mount and invoke the patcher first with `|| exit 1`; synchronize it to the worker
- add CPU behavior, not-early, non-encoder, idempotence, all-or-none drift, overlay parity, and real-Compose failure-injection coverage
- document provenance from vLLM commit `7ca49fbe4bab019e55d57cdc4b7fd3d55c67c1a6`

## Verification
- `python3 scripts/test-empty-encoder-output-hotfix.py -q` — 9/9
- `python3 scripts/test-python-hotfix-failclosed.py -q` — 8/8; the new patcher failure blocks all later patchers and service exec
- exact pinned-image ephemeral composition — first run applies both files, second run is idempotent, empty rows + completion branch present, phantom rows absent
- `docker compose -f docker-compose.dspark.yml config --quiet`
- `bash scripts/ci-validate.sh` on exact commit `ebe4ca2`
- `git diff --check origin/main`
- changed-file secret scan: no matches
- FREN parallel semantic review found the phantom-token consumer effects and supported the empty-row correction; one provisional FREN test note was discarded because it incorrectly proposed pinning the buggy `[0]` contract
- independent Fable round-2 review: PASS, no BLOCKER/HIGH/MEDIUM; its substantive round-1 completion-gap finding was implemented and re-reviewed
- final parent diff review: PASS

## Deviated
- Expanded the initial one-line correction with upstream's paired encoder-only completion branch. Empty rows alone would leave completed encoder requests producing empty steps indefinitely.

## Skipped
- Live EC-transfer/encoder-only serve validation: the shipped DeepSeek profile does not enable EC transfer, and the active TP=2 model was intentionally not restarted or touched.

## Honest gaps
- Draft remains blocked on a controlled encoder-only/EC-transfer runtime validation environment. Hosted feature-branch CI pending.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@
 
 - **Default-on Python source hotfixes now fail closed before `vllm serve` ([Issue #107](https://github.com/MiaAI-Lab/DeepSeek-v4-Flash-DSpark-2x-DGX-Spark/issues/107))**: the encoding copy/reasoning-map rewrite and Issue #21, #55, #27, #43, #26, and suppress-stops patchers were separated by bare semicolons, so a missing file, anchor drift, assertion, or self-check failure could be masked by a later successful command and startup continued with stale runtime code. Every enabled step now propagates failure explicitly with `|| exit 1`, and Issue #21 anchor drift plus a missing suppress-stops target now return nonzero; the existing missing-encoding warning and all enable/skip switches retain their prior behavior. CPU failure injection covers every step, ordering, optional Issue #31, and the suppress-stops skip path.
 
+### Fixed
+
+- **Encoder-only EC producer steps no longer emit phantom token ID 0 or loop after encoding ([Issue #109](https://github.com/MiaAI-Lab/DeepSeek-v4-Flash-DSpark-2x-DGX-Spark/issues/109))**: `make_empty_encoder_model_runner_output` now returns one distinct empty token row per scheduled request, so the scheduler cannot append token 0, advance grammar/stop state, count fake output, or distort speculative-decoding accounting when the encoder producer sampled nothing. The paired scheduler branch stops an encoder-only request once all prompt tokens are consumed, avoiding an empty-output loop. Both source overlays implement the paired semantics from [vLLM upstream commit `7ca49fb`](https://github.com/vllm-project/vllm/commit/7ca49fbe4bab019e55d57cdc4b7fd3d55c67c1a6); one fail-closed, idempotent startup patch applies both corrections to pinned image `ghcr.io/anemll/dspark-vllm-gx10:0.1.1` and is synchronized to the worker.
+
 ## 2026-08-21
 
 ### Added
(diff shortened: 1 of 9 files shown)


plotarmordev:
Final live validation passed on exact PR head `ebe4ca2145702744b0e3a09b04153eb163dfea41`, composed with exact D2 head `f62be5255fdbfa80d01f788b2f93b01fa7d20e66`, merged prerequisite #108, and the local operational-only layer.

- Complete combined CPU CI passed.
- TP=2 base model cold-started successfully on pinned image `ghcr.io/anemll/dspark-vllm-gx10:0.1.1`.
- Startup patcher reported `outputs=applied, scheduler=applied` on both ranks.
- Direct in-container assertions passed on both ranks: empty token rows present, phantom `[0]` rows absent, encoder-only completion flag and prompt-consumed stop branch present.
- `python3 /opt/hotfix-vllm-empty-encoder-output.py --status` reported `APPLIED` on both ranks.
- Live 4-lane 22,817-token cold wave completed with 0 preemptions; six-way chat smoke passed 6/6.

The shipped text model does not activate EC-transfer behavior, so the semantic encoder-only branch remains covered by the regression suite rather than a live encoder model. Startup composition and both installed rank files are proven. Ready for normal merge.

plotarmordev:
Final live validation complete on the real TP=2 DeepSeek-V4-Flash-0731 deployment (image ghcr.io/anemll/dspark-vllm-gx10:0.1.1). Startup applied the empty-encoder-output patch on both ranks; direct assertions against each installed vLLM tree confirmed the per-request empty-row expression, absence of the phantom-token expression, and the paired encoder-only scheduler completion condition. Combined PR #106 + PR #110 local CI passed. The text-only deployment cannot execute the dormant encoder-only runtime path, which is covered by the 9/9 behavior tests and exact pinned-image composition check. Post-load API smoke returned FINAL-SMOKE-OK exactly. Ready to merge.

Issue #82, opened 2026-08-18
[Bug] Multi-turn tool history: repeated tool/reasoning attractor (low effort; isolated replay clean)


MiaAI-Lab:
Split from #37. The original first-turn `high`/`max`+tools DSML leak (Bug A) and the hop-3 official-history leak on that shape (Bug B) **did not reproduce** on `main` at `1969774` for both the lab and the original reporter ([comment](https://github.com/MiaAI-Lab/DeepSeek-v4-Flash-DSpark-2x-DGX-Spark/issues/37#issuecomment-5278470122)).

This ticket tracks the **later, different** failure reported by @Xdd-xund on 2026-08-17:

https://github.com/MiaAI-Lab/DeepSeek-v4-Flash-DSpark-2x-DGX-Spark/issues/37#issuecomment-5316128681

Upstream A/B notes: https://github.com/Anemll/dspark-vllm-gx10/issues/3#issuecomment-5315948654

## What this is not

- Not Bug A (`reasoning_effort=high` first turn). Traffic is **`low`**.
- Not #52 / PR #53: last role is **`user`**, not `assistant`.
- Not #55: repeated calls were complete/parseable JSON, not truncated.
- Not #26 v1: reporter already on **#26 v2** (byte-identical to `main`).
- Not a claim that current `main` is broken: their checkout is `2f180e7` + local backports.

## Shape

- 2× DGX Spark TP=2, Anemll `0.1.1`, official 0731 @ `9e165c30`, `nvfp4_ds_mla`, MTP=5, APC on
- Captured Pi request: 46 messages, 4 tools, official prior `tool_calls` + matching `tool` messages, `chat_template_kwargs={"thinking":true,"reasoning_effort":"low"}`, `thinking_token_budget=2048`
- A Hermes fixture with `low` and **no** budget can show the same family of repeated-reasoning / repeated-tool behavior, so the old #31 Python budget hook cannot be the sole cause

## Observed failure

Once an attractor starts in a long live session, later assistant turns accumulate:

- hundreds/thousands of `count=<n>` lines
- false `<system>` markers
- repeated tool calls / repeated reasoning until the output limit

**Byte-faithful replay of the captured pre-loop request is 3/3 clean.** Trigger appears to include live multi-turn trajectory / warm server state, not a fixed context-length threshold.

[...]

PR #111, opened 2026-08-22
test(agent): add issue 82 live trajectory verifier

plotarmordev:
## Summary

- add a side-effect-free 20-turn Chat Completions trajectory verifier for Issue #82
- seed a 41-message synthetic history approximation and A/B prior assistant reasoning replay versus drop
- accept both `reasoning` and `reasoning_content`, canonicalize replay to `reasoning`, and verify through `/tokenize` + `/detokenize` that model-produced reasoning is present in the server-rendered replay prompt and absent from the drop prompt
- retain immutable, hash-verifiable request snapshots and capture pre-parser output from returned output token IDs
- retain malformed responses, HTTP failures, and detokenization failures as partial failing evidence instead of losing the run
- expose optional `--thinking-token-budget`, `--temperature`, and `--top-p` canary controls while preserving deterministic no-budget defaults
- score stale/invalid calls, parser/DSML mismatches, count markers, false system tags, length exhaustion, and run-wide repetition
- add 19 CPU tests, CI wiring, evaluation guidance, and an explicit diagnostic-only changelog entry

## Live evidence

Earlier official-model matrix on DeepSeek-V4-Flash-0731 revision `9e165c30`, pinned Anemll 0.1.1 image, TP=2, current `main`:

- 3 independent synthetic trajectories, 20 measured turns each
- 41 seeded messages, then alternating required-tool and required-no-tool turns; 91 final messages
- prompt range 43,390 to 45,731 tokens
- each run: 10 required tool calls and 10 required no-tool acknowledgements
- maximum identical reasoning line: 1
- maximum normalized tool-call repetition: 1
- raw pre-parser DSML and parsed tool calls agreed on every measured turn
- shorter reasoning-drop control: 3/3 clean
- shorter reasoning-replay replication: 3/3 clean

One earlier short seeded replay produced `_record_checkpoint` instead of `record_checkpoint` on one of 30 required calls. It was complete, parsed, did not repeat, and did not recur in the next 90 replay calls. It remains a separate transient name-fidelity observation, not Issue #82 evidence.

## Validation

At `172e552`:

- `python3 scripts/test-issue82-tool-history.py -q` — 19/19 pass
- `bash scripts/ci-validate.sh` — pass; CPU-only
- stored request bodies recompute to their recorded SHA-256 and retain original per-turn message counts
- seeded reasoning cannot satisfy the live replay guard
- replay and drop tests independently verify rendered-prompt presence/absence
- malformed-response coverage proves partial evidence survives missing token IDs
- sampling and budget options are present only when selected and reach every request
- live `/tokenize` → `/detokenize` probe against the already-serving abliterated model confirmed that a replayed `reasoning` field appears in the rendered prompt
- live two-turn abliterated drop smoke passed; the replay smoke failed loudly as designed because this serve returned `reasoning=null` on both turns
- no model restart or switch was performed

## Review fixes

- snapshot each request at send time so later history growth cannot mutate retained evidence
- require stored request hashes and original per-turn message counts to verify
- distinguish request-body replay from server-rendered replay with one explicit rendered-prompt proof
- expose the captured budget arm and sampled-trajectory controls
- retain malformed-response and detokenization failures as partial evidence
- use accurate run-wide repetition wording and dated threshold provenance
- normalize seeded/live assistant message shapes and share synthetic tool-result construction
- document the deliberately conservative DSML diagnostic boundary

## Boundary

This is not a byte-faithful reconstruction of the reporter's fixture. The private frozen request had 46 messages, final role `user`, and a free-running 20-turn driver that produced 12 tool-call finishes plus 8 normal stops on the healthy baseline. This public harness instead uses a 41-message synthetic seed and explicitly requires alternating tool/no-tool behavior. The private fixture and driver remain unavailable.

Current `main` therefore did not reproduce the attractor under a synthetic stress approximation. This PR deliberately changes no tokenizer, parser, scheduler, or model behavior and does not claim Issue #82 is fixed. It supplies an auditable diagnostic gate and the missing A/B controls for the next retained-baseline/current-main comparison.


plotarmordev:
Recommendation: **do not merge this PR in its current form. Keep it draft as an investigation artifact.**

The implementation is internally consistent and CI-clean, but its permanent maintenance cost is not proportionate to the evidence:

- 1,137 added lines: 740-line verifier, 378-line test suite, plus CI/docs
- no production serving behavior changes
- no reproduction or fix of Issue #82
- synthetic 41-message, forced alternating tool/no-tool trajectory rather than the reporter's unavailable 46-message final-user fixture and free-running driver
- current evidence is a clean non-reproduction of that approximation, not evidence that #82 is resolved on latest `main`

The branch remains useful: the reporter can run it against the retained baseline, and its artifacts can discriminate generation looping from parser loss. That does not require adding the framework to every checkout yet.

Reconsider merging if one of these occurs:

1. the reporter runs it and reproduces or materially discriminates #82;
2. the sanitized fixture/driver becomes available and the verifier is adapted to the real shape; or
3. maintainers explicitly choose to own a general trajectory-diagnostics framework.

If committed tooling is still desired without those prerequisites, this should first be reduced to the essential loop, replay/drop control, raw-versus-parsed discriminator, repetition gate, and one artifact path. Until then, keep Issue #82 open and do not treat this PR as a resolution.

plotarmordev:
**Maintainer status: retained as DRAFT investigation — the 2026-08-22 recommendation stands.**

Reviewed head: `172e552a2f014d79c46c52b022695741749a34ce` (unchanged; this confirms currency rather than restating the earlier comment).

The branch remains a useful investigation artifact — internally consistent, and its artifacts can discriminate generation looping from parser loss — but the current evidence is a clean non-reproduction of a synthetic approximation of #82, not evidence about the real failure. This stays a draft; it is not being closed, and age alone would not be a reason to close it.

**Evidence needed for reconsideration** (unchanged from the earlier comment): (1) the reporter runs it and reproduces or materially discriminates #82; (2) the real sanitized fixture/driver becomes available and the verifier is adapted to its shape; or (3) maintainers explicitly choose to own a general trajectory-diagnostics framework. Absent one of those, any committed version should first be reduced to the essential loop, replay/drop control, raw-vs-parsed discriminator, repetition gate, and one artifact path.

**What would unblock.** One of the three conditions above; until then, do-not-merge stands and issue #82 stays open.

plotarmordev:
## Hold update

Public head checked: `172e552a2f014d79c46c52b022695741749a34ce`.

**Why it is held:** The verifier is still an investigation tool, not evidence that issue #82 is fixed or that we should maintain the full framework permanently.

**What we tested:** The published verifier ran a 20-turn synthetic trajectory with the loop breaker off and on. Both runs passed: ten tool finishes and ten ordinary stops each, with no repeated-reasoning or repeated-call failure. Tools were simulated, not executed. These are healthy controls; neither run reproduced the reported incident.

**Next action — me, @plotarmordev:** Decide whether to adopt a small diagnostic tool when it can distinguish the real incident, when a sanitized representative fixture/driver is available, or when we explicitly choose to maintain a general framework.

**Author action now:** None. Keep draft and keep issue #82 open. Do not repeat a synthetic passing run just to clear this hold, and do not label it a fix for a failure it did not reproduce.

Issue #65, opened 2026-08-15
infra: engine death — CuTeDSL/TileLang JIT during inference exceeds sample_tokens RPC timeout (VLLM_EXECUTE_MODEL_TIMEOUT_SECONDS=300 default)


plj5003:
## Incident (2026-08-15 ~00:03 UTC)

DeepSeek-V4-Flash-0731 DSpark (vLLM, 2x DGX Spark, :8888) died mid-request:

```
WARNING [jit_monitor.py:129] CuTeDSL JIT compilation during inference: W4A16FusedMoeKernel
EngineCore ERROR TimeoutError: RPC call to sample_tokens timed out.
APIServer ERROR EngineDeadError: EngineCore encountered an issue.
INFO Application shutdown complete.
```

## Timeline
- 22:0x — container start, served fine for ~2h
- 23:58:48 — a request hit an MoE/batch shape not covered by warmup; CuTeDSL began JIT-compiling `W4A16FusedMoeKernel` (+ TileLang `mhc_pre_big_fuse_with_norm`) **during inference**
- 23:59:31 → 00:03:31 — `No available shared memory broadcast block found in 60 seconds` every minute (worker blocked on compile, GPU at 96%)
- 00:03:31 — `sample_tokens` RPC deadline hit (default 300s) → EngineCore treated worker as dead → full engine shutdown
- 00:06 — `restart: unless-stopped` auto-respawned the container; ~12 min cold-start; service recovered without manual intervention

## Root cause (verified in installed vLLM source)
1. NVFP4 MoE + MHC-hybrid kernels require per-shape JIT on Blackwell (sm_121a). New shape mid-inference ⇒ minutes-long CuTeDSL/TileLang compile inside the worker process.
2. `multiproc_executor.sample_tokens()` uses `timeout=envs.VLLM_EXECUTE_MODEL_TIMEOUT_SECONDS` — **default 300s, not overridden in our compose**.
3. vLLM v1 treats an RPC timeout as a hard worker death: `TimeoutError` → `EngineDeadError` → shutdown, instead of waiting for the compile to finish.

Not a fabric/network quirk (zero NCCL/RoCE errors, API pings answered throughout, GPU genuinely busy). Not a weights/hardware fault (clean reload after respawn).

## Fix (this PR)
[...]

PR #112, opened 2026-08-22
feat(compose): add worker-safe healthcheck to vllm-dspark

hecisaza:
## What

Adds a `healthcheck` to the `vllm-dspark` service so Docker reports real serving status instead of only process Up/Down.

## Why

The compose has no healthcheck today — `docker ps` shows `Up` even if vLLM never finished loading or has stopped serving. The only readiness signal is an out-of-band `curl`.

## Worker-safe by design

The same compose runs on both ranks, but only rank 0 serves HTTP; the worker is `HEADLESS=1` with no `:8888` listener. An ungated `curl :8888/health` marks the worker **permanently unhealthy**. The `if [ -n "$HEADLESS" ]` gate skips the probe on headless ranks. `$$HEADLESS` (double) is evaluated in the container at runtime; `${VLLM_PORT:-8888}` is resolved by Compose.

## Validated on real hardware

2× DGX Spark (GB10), image `anemll/dspark-vllm-gx10:0.1.1`, TP=2. Same worker, only the healthcheck differed:

| healthcheck | real worker (rank 1) |
|---|---|
| ungated `curl /health` | **unhealthy** — `Failed to connect to 127.0.0.1 port 8888` |
| gated (this PR) | **healthy**; head reports real status, serves normally |

Head reaches `healthy` ~6 min after start (first `/health` 200); `start_period: 480s` covers it with margin. `docker compose config` parses on both `HEADLESS=` and `HEADLESS=1`, and substitutes `${VLLM_PORT}` (default `8888`, override `9999` confirmed). `ci-validate.sh` passes, including a new assertion that the healthcheck is present and HEADLESS-gated.

## Limitations (stated deliberately)

1. **Worker health is gated, not probed.** A headless rank has no HTTP endpoint, so its check passes whenever the container can run a shell. A green worker status is not an engine probe.
2. **Not self-healing.** `restart: unless-stopped` restarts on process exit, not on `unhealthy` (confirmed: an unhealthy container keeps running). Auto-restart-on-unhealthy would need an autoheal sidecar or external watcher.
3. **Does not detect an alive-but-wedged engine** (#65/#87). `/health` → `check_health()` is a dead/stopped boolean (`errored = engine_dead or not is_running`), not a responsiveness probe.

## Notes

- `start_period: 480s` is tuned for this hardware's ~6 min cold load; operators on slower storage should raise it.
- Touches `docker-compose.dspark.yml`, which open PR #62 also edits — whichever lands second should rebase (disjoint hunks).

### DIFF
--- a/docker-compose.dspark.yml
+++ b/docker-compose.dspark.yml
@@ -11,6 +11,18 @@ services:
     restart: ${DSPARK_RESTART_POLICY:-unless-stopped}
     stop_grace_period: ${DSPARK_STOP_GRACE:-10s}
 
+    # Only rank 0 serves HTTP; the worker runs HEADLESS with no :8888 listener.
+    # Gate on $HEADLESS so the worker is not marked unhealthy forever. $$HEADLESS
+    # (double) is evaluated in the container at runtime; ${VLLM_PORT:-8888} is
+    # resolved by Compose. start_period covers cold model load (~6 min measured;
+    # raise it if your storage is slower).
+    healthcheck:
+      test: ["CMD-SHELL", "if [ -n \"$$HEADLESS\" ]; then exit 0; fi; curl -fsS http://127.0.0.1:${VLLM_PORT:-8888}/health || exit 1"]
+      interval: 30s
+      timeout: 10s
+      retries: 3
+      start_period: 480s
+
     ulimits:
       memlock: -1
       stack: 67108864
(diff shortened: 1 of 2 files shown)


MiaAI-Lab (approved):
Approved and merging.

Worker-safe HEADLESS gate is the right shape for TP=2: rank 0 probes `/health`, rank 1 does not false-unhealthy. Hardware result on Anemll 0.1.1 matches that. Serving path is unchanged; `unless-stopped` still does not restart on unhealthy, and this does not claim a wedged-engine probe (#65/#87).

Non-blocking: the new `ci-validate.sh` assertion sits after the fail/exit block, so it cannot fail the job. Follow-up can move it above that block.

Issue #80, opened 2026-08-18
[Scheduler/Fairness] Dual 400K cold prompts: one decode lane collapses to 0.69 tok/s during peer prefill on 2× DGX Spark TP=2


rickyrvl:
## Summary

We can reliably reproduce a severe mixed prefill/decode fairness problem on 2× NVIDIA DGX Spark with TP=2.

Two independent cold requests, each containing approximately 400K prompt tokens, are submitted simultaneously. Both requests eventually complete successfully, with no OOM, preemption, CUDA, NCCL, or HTTP errors.

However, the request that reaches decode first drops to approximately **0.69 tok/s** while the second 400K prompt is prefilling. Once the second prefill finishes, its decode runs normally at approximately **49 tok/s**.

This does not appear to be physical KV-cache exhaustion. During the failure, KV usage remained low and the waiting reason was `capacity`.

We believe this is the same scheduler/fairness class discussed in #27, #43, PR #44, PR #47, and #45, but we have not found a published reproduction using two prompts of approximately 400K tokens each.

## Environment

- Hardware: 2× NVIDIA DGX Spark / GB10
- Interconnect: ConnectX-7 / 200 Gb/s RoCE
- Tensor parallelism: TP=2
- Pipeline parallelism: 1
- Runtime image: `ghcr.io/anemll/dspark-vllm-gx10:0.1.1`
- vLLM base: `0.25.2.dev0+g752a3a504.d20260714`
- Model: `deepseek-ai/DeepSeek-V4-Flash-0731`
- Model revision: `9e165c30e2704aec5d9d593cce3eebd58bbef1cb`
- Served model: `deepseek-v4-flash-0731`
- KV cache: `nvfp4_ds_mla`
- KV block size: 256
- Maximum model length: 1,048,576
- Maximum sequences: 6
- Maximum batched tokens: 8,192
- Chunked prefill: enabled
- Prefix caching: enabled
- Async scheduling: enabled
- DSpark speculative decoding: enabled
- MTP speculative tokens: 5
- Sampling: temperature 0
- Client path: direct vLLM `/v1/chat/completions`; no gateway or fallback
- No unrelated inference traffic during the test

## Reproduction

1. Restart the two-node service to clear active requests and prefix-cache state.
2. Confirm Running=0, Waiting=0, and KV usage=0.
3. Submit exactly two independent `/v1/chat/completions` requests concurrently.
[...]

PR #114, opened 2026-08-23
fix(sched): cap mixed prefills around active decode

Acermax:
## Summary

Mitigates #80's mixed prefill/decode wall-clock stall on the current pinned Anemll vLLM image.

- keep the global long-prefill threshold at 1,024 when no decode lane is eligible
- cap each RUNNING or newly admitted/resumed WAITING prefill chunk at 256 while an eligible decode lane exists
- preserve the existing #43 decode-service floor and diagnostics
- expose `DSPARK_MIXED_PREFILL_TOKEN_CAP` (`256` default, `0` rollback)
- add fresh/legacy/v2→v3 patch upgrade handling and fail-closed anchors
- reconcile issue43 diagnostics when PRIORITY preemption rolls back a scheduled request

This is an interim post-TTFT QoS mitigation. It does not solve the initial serialized-prefill/TTFT stagger tracked by #45.

## Root-cause evidence

Per-step scheduler diagnostics showed that the affected decode lane was not skipped:

```text
prefill_toks={peer: 1024}
decode_toks={active: 6}
decode_skips=[]
```

Host-side timing on both TP ranks localized the slowdown to the mixed `execute_model` span:

| Shape | Rank 0 execute p50/p95 | Rank 1 execute p50/p95 |
|---|---:|---:|
| decode-only `(6,1,6)` | 3.64 / 5.48 ms | 3.09 / 3.80 ms |
| mixed `(1030,2,1024)` | 577.26 / 620.97 ms | 577.28 / 620.76 ms |

The mean absolute TP-rank difference on 480 mixed steps was 0.40 ms, so this was not a rank/fabric straggler. Sampling was secondary at about 36.5 ms.

## Threshold A/B

Deterministic two-lane workload: Lane A = 256 cold prompt + 8,192 output; Lane B = cold 98,304 prompt + 32 output; TP=2, async scheduling, MTP5, prefill cap=2, zero preemptions/waiting/cache hits.

| Global threshold | Mixed tok/s | Mixed ITL p95 | B TTFT | Pair makespan |
|---:|---:|---:|---:|---:|
| 1024 | 9.74 | 665 ms | 60.18 s | 146.46 s |
| 512 | 15.51 | 515 ms | 75.06 s | 156.66 s |
| 256 | 24.58 | 257 ms | 94.22 s | 161.12 s |
| 128 | 29.34 | ~200 ms runner p95 | 157.59 s | 198.67 s |

256 was the measured knee: 128 gained only about 19% mixed decode but regressed TTFT about 67% and makespan about 23% versus 256.

## Dynamic-cap live validation

Configuration:

```text
LONG_PREFILL_TOKEN_THRESHOLD=1024
DSPARK_MIXED_PREFILL_TOKEN_CAP=256
DSPARK_MAX_INFLIGHT_PREFILLS=2
```

Prefill-only control retained four 1,024-token chunks for a cold 4,096-token prompt.

The first mixed batch, including a newly admitted WAITING request, was exactly:

```text
scheduled_toks=262 requests=2 max_req_toks=256
```

on both ranks. No initial 1,030-token mixed batch occurred.

| Cold peer prompt | Mixed tok/s | Recovery tok/s | Max gap | Verdict |
|---:|---:|---:|---:|---|
| 32K | 22.49 | 87.35 | 2.94 s* | NO_REPRO |
| 64K | 24.41 | 87.41 | 0.304 s | NO_REPRO |
| 98K | 23.74 | 86.88 | 0.361 s | NO_REPRO |
| 131K | 23.53 | 86.72 | 0.389 s | NO_REPRO |

`*` one-time first-shape JIT, symmetric across both ranks.

All cells passed exact access-log, API/Prometheus counter, SSE/usage, sampler freshness, final finite-metric, and drain gates. Final state was Running=0, Waiting=0, KV=0, preemptions=0. MTP acceptance was 27,444 / 27,485 = 99.85%. No CUDA, NCCL, Xid, OOM, or engine errors occurred.

The no-cap 98K baseline reproduced `PASS_DEGRADED_CLASS` and confirmed 3/3 with fresh prefixes. Dynamic v3 returned `NO_REPRO` through 131K.

## Scheduler details

- eligible decode detection scans the full RUNNING list, independent of request order
- finished async sentinels, future-cadence requests, and prefill chunks do not activate the cap
- the cap is applied before Mamba alignment and the existing #43 floor
- the same schedule-start decision caps the first non-async WAITING chunk after RUNNING scheduling
- async remote-KV load-zero remains untouched
- `0` disables the cap; invalid values fail closed at scheduler import
- v3 supports fresh install, old #43 legacy upgrade, v2→v3 upgrade, and current idempotence

## Validation

- 25 focused cap/parser/WAITING/upgrade/preemption tests
- exact pinned-image scheduler extraction and #27→#43/v3 patch application, compile, and idempotence
- issue43 simulator at partial-prefill caps 1 and 2 with zero decode skips
- complete `bash scripts/ci-validate.sh`
- `py_compile`, shell syntax, Compose rendering, and `git diff --check`

## Scope and limitations

- The diagnostic GPUWorker timing patch used to localize the stall is intentionally not included here.
- This does not claim to satisfy #45's whole-request min/max or TTFT-skew acceptance criteria.
- The exact dual-400K workload should remain an upstream acceptance cell; this PR validates cold peers through 131K plus ordinary-traffic captures.
- Rollback requires setting `DSPARK_MIXED_PREFILL_TOKEN_CAP=0` and recreating both containers because the value is parsed at module import.

Addresses #80. Related: #27, #43, #45, #47.

### DIFF
--- a/.env.dspark.example
+++ b/.env.dspark.example
@@ -190,6 +190,11 @@ MAX_NUM_BATCHED_TOKENS=8192
 # from ~46x to ~1.05x. Safe across MAX_NUM_SEQS and prompt lengths.
 LONG_PREFILL_TOKEN_THRESHOLD=1024
 
+# Issue #80: cap each prefill chunk only while an eligible decode lane is
+# active. Prefill-only work retains LONG_PREFILL_TOKEN_THRESHOLD=1024.
+# Default 256; set 0 to disable/roll back. Valid range: integer 0..8192.
+DSPARK_MIXED_PREFILL_TOKEN_CAP=256
+
 # Issue #43: max concurrent in-flight partial prefills. Raising to 2/3
 # overlaps prefills and is the lever that improves the whole-request min/max
 # decode rate #43 reports. NOTE: --max-num-partial-prefills is NOT available
--- a/docker-compose.dspark.yml
+++ b/docker-compose.dspark.yml
@@ -81,6 +81,10 @@ services:
       # (scheduled prefill/decode tokens per request + zero-token decode
       # skips) when set to 1. Off by default (zero overhead).
       DSPARK_ISSUE43_SCHED_DIAG: "${DSPARK_ISSUE43_SCHED_DIAG:-0}"
+      # Issue #80: preserve the global 1024 prefill threshold when no decode is
+      # active, but cap new/resumed and running mixed-prefill chunks at 256.
+      # Set 0 to disable/roll back; parsed once when the scheduler imports.
+      DSPARK_MIXED_PREFILL_TOKEN_CAP: "${DSPARK_MIXED_PREFILL_TOKEN_CAP:-256}"
       # In-flight partial-prefill cap for the #27 hotfix (1-3, default 2).
       # Do NOT pass --max-num-partial-prefills (Anemll 0.1.1 rejects it).
       DSPARK_MAX_INFLIGHT_PREFILLS: "${DSPARK_MAX_INFLIGHT_PREFILLS:-2}"
--- a/patches/hotfix-dsv4-issue43-decode-fairness-and-diag.py
+++ b/patches/hotfix-dsv4-issue43-decode-fairness-and-diag.py
@@ -52,9 +52,30 @@ D. Step summary log (ask #1). When ``DSPARK_ISSUE43_SCHED_DIAG=1`` is set in
    decode_skips=[(rid,pos,computed)..]``. Off by default (zero overhead: the
    diag dict is cheap to build and only the log line is gated).
 
-Idempotent. Anchors are asserted; re-applying is a no-op once all marks are
-present. Safe to apply before or after the issue #27 hotfix (independent
-regions).
+E. Issue #80 mixed-step chunk cap. Parse ``DSPARK_MIXED_PREFILL_TOKEN_CAP``
+   once at module import (default 256; 0 disables; valid range 0..8192).
+   Whenever any eligible decode lane exists anywhere in ``self.running``, cap
+   each prefill chunk before generic Mamba block alignment and then apply the
+   issue #43 floor. Prefill-only steps retain the global
+   ``--long-prefill-token-threshold``.
+
+   The v3 revision also applies that same schedule-start decision to the first
+   chunk of a newly admitted or resumed WAITING request. The WAITING cap lives
+   in the ``load_kv_async == False`` branch, after the global threshold and
+   before budget checks and Mamba alignment; async remote-KV loads still
+   schedule zero new tokens and are untouched.
+
+F. Priority-preemption diagnostic rollback. When stock PRIORITY scheduling
+   removes a request already scheduled in the current step and restores its
+   token budget, remove that request from the prefill/decode diagnostic maps too
+   so their totals continue to match ``num_scheduled_tokens``.
+
+Versioned and idempotent. Current v2 sources receive only the WAITING block and
+v3 marker. A source carrying only the old issue #43 marker receives all issue80
+features, including the WAITING block. Fresh stock receives the complete current
+patch. Every upgrade anchor is asserted and all edits remain in memory until the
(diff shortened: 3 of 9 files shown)


Acermax:
## ⚠️ Real-traffic regression after opening this PR — do not merge yet

A subsequent ordinary-production workload with 3–5 concurrent requests reproduced the same mixed-forward wall-clock slowdown and, subjectively and in aggregate throughput, was worse overall than the pre-PR runtime.

Observed 10-second windows included:

- 2 running requests: 7.0–12.1 generation tok/s
- 3–4 running plus waiting/deferred work: 2.0–7.5 tok/s
- later 3 running: 7.8–18.9 tok/s before recovery to 61–69 tok/s

This was not physical throttling or queue/KV exhaustion:

- decode skips: 0
- preemptions: 0
- KV: ~1.4–11.2%
- both ranks P0 at ~2.17 GHz
- no thermal/power/HW slowdown, CUDA, NCCL, Xid, OOM, or engine errors

Runner/scheduler attribution:

- one capped prefill + decode: `(262, 2, 256)`, execute p50 ~314–335 ms
- two capped prefills + decode: `(518/524, 3–4, 256)`, execute p50 ~487–494 ms
- pure/cadence-ineligible prefill windows still produced 1024/2048-token shapes with ~0.7–1.5 s steps
- real-workload MTP acceptance was ~43–80%, versus 99.85% in the deterministic validation

So the current cap is too narrow: it is per prefill request, allowing two prefills to contribute 512 tokens in one mixed step, and it prolongs the mixed-prefill overlap. The deterministic two-lane result does not generalize to this higher-concurrency workload.

I am rolling the deployment back with `DSPARK_MIXED_PREFILL_TOKEN_CAP=0` after active requests drain. This PR is now draft while the policy is redesigned around a bounded total mixed-prefill budget / higher-concurrency acceptance matrix rather than a per-request cap.

Acermax:
Closing this PR after the production follow-up and a separate successor-policy A/B.

The original per-request 256 cap in this PR was already rolled back after ordinary 3–5 request traffic regressed. I then tested a separate TP-local cadence successor (`mixed -> decode-only -> mixed`, interval 2) from a clean worktree; that code was **not** pushed into this PR.

The warmed five-lane simultaneous A/B used identical contracts (5 cold 32K prompts, 512 output/lane, exact API/Prometheus reconciliation, zero preemptions/errors):

| Metric | cap0 baseline | cadence interval 2 | Change |
|---|---:|---:|---:|
| Makespan | 98.274 s | 121.627 s | **+23.8%** |
| TTFT median | 69.729 s | 86.159 s | **+23.6%** |
| TTFT max | 89.685 s | 113.588 s | **+26.7%** |
| TTFT spread | 55.843 s | 73.616 s | **+31.8%** |
| Lane tok/s minimum | 8.724 | 9.153 | +4.9% |
| Lane tok/s median | 19.524 | 17.057 | **-12.6%** |
| Max stream-event gap | 1.186 s | 2.131 s | **+79.6%** |
| Waiting duration | 68.0 s | 84.5 s | **+24.3%** |
| MTP cycle rate | 4.376 Hz | 3.535 Hz | **-19.2%** |

The candidate measurement was warm: the first-use W4A16/top-k JIT pass was discarded, and the measured pass had no new JIT or runtime error. MTP acceptance remained approximately equal, so this was a scheduling/cycle-rate regression rather than acceptance loss.

The cadence candidate was rejected immediately and the deployment was recreated at the cap0 baseline. Both ranks are healthy/restart0, API 200, Running=Waiting=KV=0, preemptions=0, diagnostics off.

Three local policy directions have now failed the complete objective: global small chunks, this PR's per-request dynamic cap, and periodic decode-only cadence. I do not see a responsible path to merge this PR. The remaining credible direction is architectural: true concurrent partial prefill in a compatible base image, or resource-isolated/disaggregated prefill with additional capacity.

Issue #87, opened 2026-08-19
[Bug] Recurring NCCL collective timeout hangs the TP=2 pair (EngineDeadError) on 2x DGX Spark


cccbleach:
### Summary

The dual-DGX-Spark TP=2 deployment intermittently wedges: the vLLM engine stops generating, while the API surface stays alive. It has now happened **three times** (2026-08-14, 2026-08-18 ~17:00 CST, 2026-08-18 ~22:20 CST). Only a full stop/start of both ranks recovers it. We are following this recipe (repo @ 6f0a061) plus the Anemll GB10 image.

### Environment

- Recipe: MiaAI-Lab/DeepSeek-v4-Flash-DSpark-2x-DGX-Spark @ 6f0a0613c187 (first occurrence on f752cd0)
- Runtime image: ghcr.io/anemll/dspark-vllm-gx10:0.1.1 (config id sha256:3430d6614a8e2925f34d059af6caf05aff42387326db4d05639a60f10f2654d8)
- vLLM: 0.25.2.dev0+g752a3a504.d20260714, NCCL 2.30.7+cuda13.3
- Hardware: 2x NVIDIA DGX Spark (GB10, sm_121), Connector 7 200Gb/s RoCE direct link (~185 Gbps measured, 0 loss), TP=2, PP=1
- Serving profile: MTP-5 DSpark spec decode, nvfp4_ds_mla KV, flashinfer_b12x MoE, FP8 weights, 1M context, max_num_seqs=6
- OS: DGX OS 26.04.1 / Ubuntu 24.04.4, kernel 6.17.0-1029-nvidia, driver 580.173.02
- Models: both lanes affected — official deepseek-ai/DeepSeek-V4-Flash-0731 @ 9e165c30 and abliterated drowzeys/keys-DeepSeekV4-Flash-GA-0731-Dspark-Abliterated-32-32 @ 6c47916f
- Recipe hotfixes enabled: Issue #21/#22/#26v2/#27/#31/#43/#55 + stop suppression + vLLM 0.27 perf backports (defaults per the repo)

### Symptom A (2026-08-18 17:00 CST, official lane)

API server died with EngineDeadError, then NCCL flight-recorder dumps:

    (APIServer pid=1) ERROR [serving.py:851] vllm.v1.engine.exceptions.EngineDeadError: EngineCore encountered an issue. See stack trace (above) for the root cause.
[...]

PR #118, opened 2026-08-23
fix(boot): persist Triton cache + pre-warm spec-decode shape buckets (closes #117 local surface)

### LINKED ISSUE #117: [Bug] Mid-serve Triton JIT (_prepare_dflash_inputs_kernel) stalls a rank past torch's 600s NCCL watchdog — TP=2 pair dies despite the #65 1800s mitigation
## Summary

Two TP=2 pair-loss incidents within 12 hours on the same 2× DGX Spark pair, both in the #65/#87 class, **after** the #65 mitigation (`VLLM_EXECUTE_MODEL_TIMEOUT_SECONDS=1800`) was live. The second one has a complete log chain: a **mid-serve Triton JIT compile** stalled one rank, the peer sat in `_ALLGATHER_BASE` until **PyTorch's ProcessGroupNCCL watchdog killed it at its own 600 s timeout** — a deadline the #65 env does not extend. The 1800 s RPC deadline never got a vote.

## Environment

- 2× DGX Spark (GB10), TP=2, RoCE, Anemll `0.1.1` (vLLM 0.25.2.dev0+g752a3a504)
- recipe checkout `73136d3` (= current `main` + PR #116), official 0731 @ `9e165c30`
- MTP=5/probabilistic, APC on, `FULL_AND_PIECEWISE`, `VLLM_EXECUTE_MODEL_TIMEOUT_SECONDS=1800`, persistent TileLang cache on the HF volume, `jit_monitor_mode=warn`
- Sustained real agent traffic (3 concurrent client connections) at the time of the second incident

## Incident A (freeze variant) — 08-22 13:10 local

External liveness watchdog observed `running=1`, engine iteration counter frozen for 3 × 2-minute samples, and performed its paired restart. No crash, no NCCL timeout surfaced — the engine simply stopped stepping mid-request. (This is the variant #87 describes.)

## Incident B (kill variant) — 08-22 17:06–17:27 UTC, full chain

Head (rank 0), during normal concurrent serving:

```text
17:06:23 (Worker_TP0) WARNING [jit_monitor.py:129] Triton kernel JIT compilation during
         inference: _prepare_dflash_inputs_kernel. This causes a latency spike;
         consider extending warmup to cover this shape/config.
17:07:33 (EngineCore) INFO [shm_broadcast.py:705] No available shared memory broadcast
         block found in 60 seconds. (repeats every 60 s thereafter)
```

Worker (rank 1):

```text
17:17:34 [E ProcessGroupNCCL.cpp:2119] [PG 2 Rank 1] Watchdog caught collective operation
         timeout: WorkNCCL(SeqNum=1485, OpType=_ALLGATHER_BASE, NumelIn=1163520,
         NumelOut=2327040, Timeout(ms)=600000) ran for 600083 milliseconds before timing out.
17:18:40 ERROR [multiproc_executor.py:284] Worker proc VllmWorker-1 died unexpectedly
17:27:12 [F ProcessGroupNCCL.cpp:1653] Terminating the process after attempting to dump debug info
```

Head then dies cleanly: `shm_broadcast acquire_read → RuntimeError: cancelled` → `EngineDeadError` → exit 0. Compose `restart: unless-stopped` brought both containers back (head ×1, worker ×2 restarts) and the pair re-formed and served normally for the next 9+ hours — the self-heal path works; the cost is a ~20-minute silent outage plus whatever was in flight.

Hardware ruled out for this event: both nodes' IB/RoCE error counters (`symbol_error`, `port_rcv_errors`, `link_downed`, `local_link_integrity_errors`) are zero lifetime; SM clocks pin at 2405 MHz under load.

## Gap analysis

1. **The #65 mitigation covers vLLM's RPC deadline, not torch's.** `VLLM_EXECUTE_MODEL_TIMEOUT_SECONDS=1800` was adopted so mid-serve CuTeDSL/TileLang JIT can ride out; but `ProcessGroupNCCL`'s collective watchdog still fires at its default 600 s and *terminates the rank*. Any single-rank stall in the 600–1800 s band kills the pair the #65 change meant to save. The two deadlines should move together (expose the PG timeout, or document the pairing).
2. **Warmup does not cover `_prepare_dflash_inputs_kernel`.** `jit_monitor` says so explicitly. The boot prewarm passes discussed in #84 (johnsen's `520591e`/`bd79b1d`) never landed in this repo's `main`; the shipped smoke request warms one shape only. Under concurrent traffic, new (batch × spec) shapes still JIT mid-serve, and every such compile is a spin of the roulette wheel in (1).

## Asks

1. Extend boot warmup (or land the #84 prewarm work) to cover the spec-decode input-prep kernel shapes reachable at `max_num_seqs=6` — `jit_monitor.py` already names the kernels to chase.
2. Expose the torch process-group collective timeout alongside `VLLM_EXECUTE_MODEL_TIMEOUT_SECONDS` (single knob or documented pair), so operators who opt into 1800 s JIT rides don't keep a hidden 600 s kill switch. External liveness watchdogs (the #87-style paired-restart script both our clusters run) can own true-hang detection at whatever deadline the operator chooses.

Happy to run candidate patches on this pair — we have the traffic profile that reproduces the class (~2 incidents / 12 h under sustained agent load).

de1tydev:
Implements the operator-side half of #117 — the two layers that don't need engine changes:

1. **`TRITON_CACHE_DIR` onto the HF volume** (mirrors the #65 `TILELANG_CACHE_DIR` fix). The in-image `~/.triton/cache` dies on container recreate, so every restart re-JITs already-known shapes mid-serve; #117's incident log shows what one such compile can cost on TP=2 (peer stalled in `_ALLGATHER_BASE` → torch's 600 s ProcessGroupNCCL watchdog kills the rank — a deadline `VLLM_EXECUTE_MODEL_TIMEOUT_SECONDS` does not extend). Cache entries are keyed by Triton version/backend hash, so the shared dir is safe across image updates.

2. **`scripts/boot-shape-warmup.sh`** — launcher post-smoke sweep (gate `DSPARK_BOOT_SHAPE_WARMUP`, default 1, **non-fatal by contract**): concurrency C=1/2/4/6 bursts, a medium and an 8192-boundary-crossing prefill, and a thinking-off arm, all nonce-tagged so prefix caching cannot skip the prefill being warmed. The observed kernel's shape key is `BLOCK_SIZE = min(256, next_pow2(max_tokens_per_req))`, so the reachable buckets are enumerable and this sweep materializes them before traffic does. `jit_monitor` warnings during serve remain the closure signal for anything missed.

Not included (left for upstream per #117 ask 2): exposing the torch process-group timeout — on our own pair we deliberately keep the 600 s backstop, because both observed stalls exceeded 10 minutes and behave like rank desync rather than slow compiles; the layered external watchdog + docker self-heal handled recovery.

## Validation
- `bash scripts/ci-validate.sh` passes; new CPU gate `scripts/test-boot-shape-warmup.sh` (12 behavioral checks via a recording curl stub: arm/tag inventory, single thinking-off arm, chunk-boundary length, per-run nonce rotation, failure and unreachable paths).
- Live validation on both of our 2× DGX Spark TP=2 pairs is running today: first boot populates `/cache/huggingface/triton-cache` and the sweep compiles the buckets; second boot must show cache reuse (fast warmup, no new compile entries) and zero `jit_monitor` warnings under traffic. Will report numbers in this thread.

https://claude.ai/code/session_01Dcfsv9svWFUNJTUGxpYY9v

### DIFF
--- a/.env.dspark.example
+++ b/.env.dspark.example
@@ -317,6 +317,18 @@ TILELANG_CLEANUP_TEMP_FILES=1
 # (stock 300s). TileLang cache lives on the HF volume, not the container layer.
 VLLM_EXECUTE_MODEL_TIMEOUT_SECONDS=1800
 TILELANG_CACHE_DIR=/cache/huggingface/tilelang-cache
+# Issue #117: persist Triton's JIT cache too — the in-image ~/.triton/cache
+# dies on container recreate, so every restart re-JITs known shapes mid-serve,
+# and a compiling rank can stall its TP peer past torch's 600s NCCL watchdog
+# (a deadline VLLM_EXECUTE_MODEL_TIMEOUT_SECONDS does not extend).
+TRITON_CACHE_DIR=/cache/huggingface/triton-cache
+# Issue #117: post-ready boot sweep that burns the spec-decode/prefill Triton
+# shape buckets (concurrency C=1/2/4/6, medium + multi-chunk prefill,
+# thinking-off arm) before real traffic can JIT them mid-serve. Non-fatal:
+# warmup failure degrades to the mid-serve-JIT status quo, never a failed
+# boot. Closure signal: jit_monitor warnings during serve = a shape the sweep
+# missed. Set 0 to skip.
+# DSPARK_BOOT_SHAPE_WARMUP=1
 DG_JIT_USE_NVRTC=0
 DG_JIT_NVCC_COMPILER=/usr/local/cuda/bin/nvcc
 NCCL_NET=IB
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+## 2026-08-23
+
+### Added
+
+- **Boot-time defense against the mid-serve-JIT → TP-pair-loss chain ([Issue #117](https://github.com/MiaAI-Lab/DeepSeek-v4-Flash-DSpark-2x-DGX-Spark/issues/117))**, two independent layers. (1) `TRITON_CACHE_DIR` now defaults to `/cache/huggingface/triton-cache` on the HF volume, mirroring the `TILELANG_CACHE_DIR` fix from issue #65: the in-image `~/.triton/cache` dies on container recreate, so every restart re-JIT-compiled already-known Triton shapes mid-serve — and a rank stalled in compilation leaves its TP peer waiting in a collective until torch's ProcessGroupNCCL watchdog (600 s, a deadline `VLLM_EXECUTE_MODEL_TIMEOUT_SECONDS` does not extend) terminates it; the field incident in #117 recorded exactly this chain from `_prepare_dflash_inputs_kernel`. Cache entries are keyed by Triton version/backend hash, so the shared directory is safe across image updates. (2) `scripts/boot-shape-warmup.sh`, run by the launcher after the smoke request (gate `DSPARK_BOOT_SHAPE_WARMUP`, default `1`): a post-ready sweep — concurrency C=1/2/4/6 bursts, a medium and a multi-chunk (8192-boundary-crossing) prefill, and a thinking-off arm, all nonce-tagged so prefix caching cannot skip the prefill being warmed — that materializes the enumerable spec-decode/prefill shape buckets (`BLOCK_SIZE = min(256, next_pow2(max_tokens_per_req))`) before real traffic can JIT them. Warmup is non-fatal by contract: a failure degrades to the mid-serve-JIT status quo and WARNs, never fails the boot; `jit_monitor` warnings during subsequent serving are the closure signal for shapes the sweep still misses. Documented in `.env.dspark.example` and `docs/ENVS.md`; CPU-gated by `scripts/test-boot-shape-warmup.sh` (recording-curl-stub behavioral matrix: arm/tag inventory, single thinking-off arm, chunk-boundary length, per-run nonce rotation, fail and unreachable paths), wired into `scripts/ci-validate.sh`.
+
 ## 2026-08-22
 
 ### Changed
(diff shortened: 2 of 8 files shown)


de1tydev:
Live validation done on both 2× DGX Spark TP=2 pairs (experimental pair: two consecutive boots; production pair: one boot), all on this branch's content:

**Boot 1 (cold persistent cache), both pairs identical:**
- sweep: 16/16 requests ok in 20 s (arms 0–6 s each)
- `/cache/huggingface/triton-cache` populated with **252 entries / 44 MB on both ranks** — the old in-container cache snapshot had 244, and the +8 delta is exactly the previously-uncovered shape family this sweep exists to burn (the `_prepare_dflash_inputs_kernel` gap from #117 among them)
- zero real `jit_monitor` warnings during boot (the only grep hits are the monitor-activation notice and the config dump)
- #112 healthcheck reports healthy on both ranks (worker HEADLESS-gated as intended)

**Boot 2 (reuse check, experimental pair):**
- sweep again 16/16 in 20 s
- **cache entry count unchanged at 252 → zero new compiles across a container recreate**, which is the property the old `~/.triton/cache` could never give
- perf sentinel unchanged: single-stream decode 67–70 tok/s (same band as pre-change baselines)

Ops note for reviewers running the external #87-style watchdog: we paired this with a pause-file TTL on our side, because the #117 incident only reached torch's 600 s kill precisely while our watchdog was paused for maintenance. Not part of this PR (the watchdog script isn't in this repo), but worth mentioning as deployment context.

plotarmordev:
Maintainer integration is #123. It preserves this PR author attribution and the persistent-cache/default-on warmup design, then applies the independent review blockers: all six live `_prepare_dflash_inputs_kernel` BLOCK keys are now reached through exact-token `/v1/completions` rungs verified by `/tokenize`; C=6 is removed at deployed `MAX_NUM_SEQS=4`; auth propagation and complete outcome accounting are fixed. Cold-cache two-rank validation will be posted on #123 before merge. This contributor PR will remain unmerged and can close after #123 lands.

plotarmordev:
Superseded by maintainer integration #123, which preserves @de1tydev attribution in commit 288e5f0 and merged after mechanism corrections, independent CYBERGLM approval, exact-head hosted CI, and fresh-cache two-rank live validation.

PR #122, opened 2026-08-23
fix(encoder): reopen generation after trailing latest_reminder

### LINKED ISSUE #120: Trailing latest_reminder after assistant leaves encoder without generation header
## Problem

With the pinned DeepSeek-V4 encoder, a transcript ending in an assistant message followed by a trailing `latest_reminder` renders as:

```text
... assistant content <|end of sentence|> <|latest_reminder|> reminder content
```

No new Assistant/thinking generation header follows the reminder. Generation therefore resumes from the same dead state as Issue #52: immediate EOS or raw DSML-like text is possible.

The existing opt-in Issue #52 hotfix only matches `index == len(messages) - 1` when that final message is the assistant, so the appended reminder defeats it.

## Scope boundary

A `latest_reminder` directly after a final user/developer turn is already correct: the checkpoint emits the generation header before that reminder. That rendering must remain byte-identical. Mid-transcript reminders and task precedence must also remain unchanged.

## Reproduction evidence

CPU execution against the real checkpoint `encoding_dsv4.py` confirmed:

- assistant → trailing `latest_reminder`: no terminal generation header (defect)
- user/developer → trailing `latest_reminder`: one existing in-slot header before the reminder (correct)
- mid-transcript reminder: correct existing behavior

## Proposed fix

Extend the opt-in `DSPARK_ENABLE_ASSISTANT_FINAL_HOTFIX=1` patch narrowly: when and only when the final message is `latest_reminder` and its immediate predecessor is assistant, append exactly one fresh Assistant/thinking generation header after the reminder. Keep fail-closed post-write self-checks and behavioral byte-identity tests for all unaffected shapes.

plotarmordev:
Closes #120. Extends the opt-in Issue #52 assistant-final hotfix narrowly.

## Problem

A harness retry can end with `assistant` followed by `latest_reminder`. Stock closes the assistant turn, renders the reminder, and leaves no terminal generation header. The existing final-assistant guard never fires because the reminder is now the final message.

## Change

- append exactly one fresh generation header only when the final message is `latest_reminder` and its immediate predecessor is assistant
- preserve byte-identical rendering for user/developer reminder tails, mid-transcript reminders, task precedence, `wo_eos`, and all existing assistant-final shapes
- extend the patcher's post-write self-check and fail-closed restore path
- add behavior regressions and operator documentation; default remains opt-in/OFF

## Verification

- `bash scripts/ci-validate.sh` — PASS at `b748c03`
- `scripts/test-assistant-final-continuation.py` — 16/16 PASS; pre-extension behavior fails the new regression
- CPU probe against the real checkpoint encoder confirmed the defective and unaffected render shapes
- per-commit secret/source-identifier scan — 0 findings

Live two-rank opt-in render verification will be posted before merge.

### DIFF
--- a/.env.dspark.example
+++ b/.env.dspark.example
@@ -231,14 +231,15 @@ VLLM_PREFIX_CACHE_RETENTION_INTERVAL=4096
 # VLLM_SUPPRESS_STOPS_IN_REASONING=0
 
 # Issue #52 / PR #53 (assistant-final continuation hotfix): when a request's
-# messages array ends with an assistant message, the stock renderer closes the
-# turn with EOS and appends no generation header, so the model generates from
-# a dead state — empty no-op turns / hallucinated DSML markup loops in agent
-# harnesses. Default 0 = stock renderer (patch is present on the worker but
-# never invoked). Set exactly 1 to apply the patch at container boot; the boot
-# then fails (exit 1) if the patch cannot be applied or verified, and an
-# already-patched encoder is re-verified instead of rewritten. See
-# docs/PATCHES.md ("Issue #52") for evidence status and fail-closed details.
+# messages array ends with an assistant message — or with a trailing
+# `latest_reminder` annotation directly after that closed assistant turn — the
+# stock renderer leaves the prompt without a generation header, so the model
+# generates from a dead state — empty no-op turns / hallucinated DSML markup
+# loops in agent harnesses. Default 0 = stock renderer (patch is present on
+# the worker but never invoked). Set exactly 1 to apply the patch at container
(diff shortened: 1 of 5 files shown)


plotarmordev:
Final combined-candidate live evidence (`e3d4bad7450c30ed9b5507a5f56aaf5d52121ec3`): the actual installed encoder hotfix self-check passed on both TP ranks, including assistant-final + trailing `latest_reminder` and byte-intact user→reminder behavior. Combined CPU gate: 16 assistant-final tests PASS; C4 live smoke 4/4. Independent CYBERGPT exact-head review: APPROVE, no required defect.

PR #124, opened 2026-08-23
fix(config): validate numeric knobs (MAX_NUM_SEQS/MTP_NUM_TOKENS) before arithmetic

hecisaza:
## What

Adds integer validation for `MAX_NUM_SEQS`, `MTP_NUM_TOKENS`, and
`MAX_NUM_BATCHED_TOKENS`, matching the existing `VLLM_PORT` treatment.

## Why

`VLLM_PORT` is validated (`start.sh:148`), but `MAX_NUM_SEQS` / `MTP_NUM_TOKENS`
are interpolated into `$(( ))` with no check — at `start.sh:791`,
`validate-dspark-config.sh:117`, and (forwarded unevaluated) the container's own
arithmetic at `docker-compose.dspark.yml:274`. A malformed operator value fails
three ways:

| input | current behaviour |
|---|---|
| `6.5`, `6x`, `08`, CRLF | validator prints a bash arithmetic error **but exits 0** and renders the compose command |
| `010` | read as **octal 8** → cudagraph capture size becomes **48 instead of 60**, silently |
| `eight` | `set -u` "unbound variable" — names neither the knob nor the file |

## Reproduced end-to-end on real hardware

2× DGX Spark (GB10), image `anemll/dspark-vllm-gx10:0.1.1`, TP=2. With
`MAX_NUM_SEQS=6.5` the validator returns **0** (approves the config), but the real
launch **crash-loops the container**: `bash: 6.5 * (5 + 1) : syntax error`,
container stuck `Restarting`. A validator-blessed value takes the cluster down.
Also confirmed in the deployed container's own shell: `010` → 48 (octal), and
overflow `18446744073709551615` → **-6** (negative capture size).

## Fix

Mirror the `VLLM_PORT` pattern in both `start.sh` and `validate-dspark-config.sh`:
a `^[0-9]+$` check plus **`10#` normalisation** — the normalisation is required, not
cosmetic, because a plain regex accepts `010`/`08` which bash then reads as octal.

On the head, the normalised value is also written back into the private env
snapshot, because the **worker** consumes it via `docker compose --env-file` over
ssh and never sees exported shell vars — without this, head and worker would
disagree (60 vs 48) on the same config.

Empty/unset is skipped, so the existing `:-` defaults still apply. Semantic bounds
(e.g. the documented `MTP_NUM_TOKENS` min of 5) are a separate concern, out of scope.

## Tests

`scripts/test-numeric-knob-validation.py` — 10 CPU-only tests, lifted verbatim from
the launcher so they gate shipped code (valid, empty, decimal, alnum, bareword,
CRLF, `010`→60, `08`→48, `MTP_NUM_TOKENS`, `MAX_NUM_BATCHED_TOKENS`), wired into
`ci-validate.sh`. `bash -n` clean; `docker compose config` parses; `ci-validate.sh`
passes with no new failures vs. the clean base. Verified by mutation: reverting the
fix regresses the suite.

### DIFF
--- a/dspark-numeric-knobs.sh
+++ b/dspark-numeric-knobs.sh
@@ -1,0 +1,47 @@
+#!/usr/bin/env bash
+# Shared numeric-knob validation for the DSpark launcher and validator.
+#
+# MAX_NUM_SEQS / MTP_NUM_TOKENS / MAX_NUM_BATCHED_TOKENS are interpolated into
+# bash arithmetic (and forwarded unevaluated into the container's own $(( )) via
+# compose). Validate them the way VLLM_PORT is: reject non-integers, reject
+# out-of-range/oversized values BEFORE any arithmetic (so a huge decimal cannot
+# wrap 64-bit and slip through as a negative), and normalise with 10# (so a
+# leading zero is decimal, never octal).
+#
+# Usage:  dspark_validate_numeric_knobs [ENV_SNAPSHOT]
+#   ENV_SNAPSHOT (optional): a file whose matching KEY= lines are rewritten to the
+#   normalised value. The launcher passes its private snapshot here so the worker,
+#   which consumes it via `docker compose --env-file` over ssh (no exported vars),
+#   sees the same decimal value as the head. Omit it (the validator) to skip.
+# Returns 2 on any invalid value; callers under `set -e` should use `|| exit $?`.
+
+dspark_validate_numeric_knobs() {
+  local snapshot="${1:-}"
+  local knob val
+  # generous upper bounds — well above any real config, purely to bound arithmetic
+  local -A _max=([MAX_NUM_SEQS]=4096 [MTP_NUM_TOKENS]=64 [MAX_NUM_BATCHED_TOKENS]=8388608)
+  local -A _min=([MAX_NUM_SEQS]=1 [MTP_NUM_TOKENS]=0 [MAX_NUM_BATCHED_TOKENS]=1)
+  for knob in MAX_NUM_SEQS MTP_NUM_TOKENS MAX_NUM_BATCHED_TOKENS; do
+    val="${!knob:-}"
+    [ -z "$val" ] && continue
+    if ! [[ "$val" =~ ^[0-9]+$ ]]; then
+      echo "$knob must be a non-negative integer: $val" >&2
+      return 2
+    fi
+    # length guard: >10 digits could overflow 64-bit arithmetic before the bound
+    # check below, so reject as out-of-range without evaluating it.
+    if (( ${#val} > 10 )); then
+      echo "$knob must be between ${_min[$knob]} and ${_max[$knob]}: $val" >&2
+      return 2
+    fi
+    printf -v "$knob" '%d' "$((10#$val))"   # normalise: strip leading zeros, decimal
+    if (( ${!knob} < _min[$knob] || ${!knob} > _max[$knob] )); then
+      echo "$knob must be between ${_min[$knob]} and ${_max[$knob]}: $val" >&2
+      return 2
+    fi
+    export "$knob"
+    if [ -n "$snapshot" ] && [ -f "$snapshot" ] && grep -q "^$knob=" "$snapshot"; then
+      sed -i "s|^$knob=.*|$knob=${!knob}|" "$snapshot"
+    fi
+  done
+}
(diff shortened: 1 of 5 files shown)


plotarmordev (changes requested):
Requesting changes. The octal/garbage handling is directionally correct, but the documented overflow failure still passes the new validator.

1. **High — unbounded decimal input still wraps through Bash arithmetic.** `^[0-9]+$` accepts arbitrary-length values before `$((10#...))`. For the PR body's own example, `MAX_NUM_SEQS=18446744073709551615` normalizes to `-1`, the capture expression remains `-6`, and the block exits 0 while exporting/writing a negative value. `2^63` similarly becomes `-9223372036854775808`; `2^64` becomes `0`. This violates the new “non-negative integer” contract and leaves the reproduced crash class open. Reject oversized input before arithmetic and enforce explicit per-knob bounds, then add boundary/overflow regressions.

2. **Medium — the permanent test does not exercise the load-bearing worker writeback or the validator copy.** The harness forces `_dspark_env_clean=` for every case, so the `grep`/`sed` branch that is supposed to keep rank 1 synchronized never runs. It extracts only the launcher block, so `validate-dspark-config.sh` can drift or lose validation while CI remains green. Use a temporary snapshot and assert its normalized contents, and exercise/extract both shipped blocks (or share one helper rather than duplicating logic).

3. **Low — `test_crlf_rejected` does not contain a carriage return.** Interpolating Python `repr('6\r')` into a shell single-quoted assignment produces the literal characters `\` and `r`, so the test is vacuous for CRLF. Feed an actual U+000D/byte through the subprocess environment or a fixture file.

4. **Low — the validator contains dead snapshot-rewrite code and a misleading launcher-only comment.** `_dspark_env_clean` is never created in `validate-dspark-config.sh`; that branch cannot run there. Remove it from the validator copy or refactor the shared validation/normalization contract.

CI is green, but none of its new cases cover the overflow that blocks merge.

hecisaza:
Thanks — all four addressed in `b0ef570`. You were right on every point, including that green CI was hiding the overflow.

**1. Overflow (High).** `^[0-9]+$` alone let arbitrarily long decimals wrap through `$((10#…))` — `18446744073709551615` → `-1`, capture `-6`, exit 0. Fixed two ways: reject anything over 10 digits **before** any arithmetic (so nothing can 64-bit-wrap into the bound check), then enforce explicit per-knob bounds — `MAX_NUM_SEQS` 1–4096, `MTP_NUM_TOKENS` 0–64, `MAX_NUM_BATCHED_TOKENS` 1–8388608. `2^63`/`2^64` and over-max now exit 2 with a range message. Regression tests added for each.

**2. Tests didn't hit the load-bearing paths (Medium).** Pulled the logic into a shared `dspark-numeric-knobs.sh` sourced by **both** `start.sh` and the validator, so neither can drift. The suite sources that function directly and asserts the **worker env-snapshot writeback** against a real temp snapshot (`010` → the file becomes `MAX_NUM_SEQS=10`), not just exit codes.

**3. CRLF test was vacuous (Low).** It was interpolating Python `repr` → literal backslash-r. Now the value is passed as a real `U+000D` byte through the subprocess environment.

**4. Dead code / misleading comment (Low).** The snapshot-rewrite exists once in the helper, guarded by whether a snapshot path is passed — the launcher passes it, the validator calls with none. No dead branch, no per-copy comment.

Suite is now 19 cases (incl. the overflow boundaries, per-knob bounds, real CR, and snapshot normalisation) and mutation-checked — dropping the guards fails 4 cases. `ci-validate.sh` passes with no new failures vs base.

plotarmordev (approved):
Re-reviewed `b0ef570`. All four requested changes are addressed:

- oversized values are rejected before Bash arithmetic, with explicit per-knob bounds and overflow regressions;
- launcher and validator now share one helper, and the worker snapshot rewrite is exercised against a real temporary file;
- the CRLF case now passes a real carriage-return byte through the environment;
- validator-only dead snapshot code and duplicated validation logic are gone.

The updated hosted `validate` check passes. No remaining blocking finding from my review.

plotarmordev:
Follow-up verification on exact head `b0ef570`:

- `MAX_NUM_SEQS=18446744073709551615` is rejected before arithmetic with exit 2; it no longer wraps to `-1` or reaches the capture expression.
- The focused revised suite passes: 19/19.
- A deliberately long zero-padded value such as `00000000000000000001` is also rejected by the 10-digit representation guard. That is conservative and safe, though stricter than numeric-value normalization; it does not reopen the overflow bug or justify blocking this PR.

Approval stands.

PR #125, opened 2026-08-23
fix(fabric): leave NCCL_IB_GID_INDEX unset after sysfs validation

drowzeys:
## Summary

Follow-up to the dual-HCA GID work ([#11](https://github.com/MiaAI-Lab/DeepSeek-v4-Flash-DSpark-2x-DGX-Spark/pull/11), [#95](https://github.com/MiaAI-Lab/DeepSeek-v4-Flash-DSpark-2x-DGX-Spark/pull/95), [#104](https://github.com/MiaAI-Lab/DeepSeek-v4-Flash-DSpark-2x-DGX-Spark/pull/104)) and the field report in [tonyd2wild#38](https://github.com/tonyd2wild/DeepSeek-v4-Flash-0731-DSpark-1M-NVFP4-KV-2x-DGX-Spark/issues/38).

`NCCL_IB_GID_INDEX` is **one global value per rank**. On dual-HCA GB10 the RoCEv2/IPv4 GID table **drifts** after link/fw events. A pin that was correct at boot (`3` on both HCAs today) later hangs with `ibv_modify_qp 61` after model load. Modern NCCL selects the RoCEv2/IPv4 GID **per device** when this env is absent.

This PR keeps sysfs validation and logging (`NCCL_IB_GID_AUTO=1`) and then **leaves the env unset** in the container.

Empty-string is not the same as unset: NCCL treats `NCCL_IB_GID_INDEX=` as GID 0 (fe80 / link-local). So:

1. After resolve, `NCCL_IB_GID_LEAVE_UNSET=1` (default) clears the shell vars.
2. `remote_nccl_env` omits the key when empty instead of sending `NCCL_IB_GID_INDEX=''`.
3. Compose entrypoint `unset`s a blank interpolation (same pattern as `NCCL_IB_MERGE_NICS`).

`NCCL_IB_GID_LEAVE_UNSET=0` restores pin-the-resolved-index. Explicit `NCCL_IB_GID_AUTO=0` pins still win.

## Field test (Keys 2× GB10, Anemll `dspark-vllm-gx10:0.1.1`, TP=2)

- Both HCAs up (`==rocep1s0f1,roceP2p1s0f1`). Sysfs still reports index 3 on every member; we just do not inject it.
- Boot ~6 min, both ranks healthy, smoke chat OK.
- vLLM process environ: `NCCL_IB_HCA==rocep1s0f1,roceP2p1s0f1`, **no** `NCCL_IB_GID_INDEX`.
- KV pool this boot: 2,592,984 tokens / 2.47× at 1M (util 0.835).
- Real speed bench (`thinking=false`, temp 0) vs the same pair with GID pinned to 3 earlier today: decode **flat** (list C1 82.7 vs 81.1; count 85.7 vs 85.3; essay 33.2 vs 34.1). This is hang-prevention, not a decode claim. The +8% in tonyd2wild#38 was vs a broken/half-merged NIC path.

## Files

- `start-deepseek-v4-flash-dspark.sh` — default-on leave-unset after validate; omit empty GID on the worker env.
- `docker-compose.dspark.yml` — unset blank `NCCL_IB_GID_INDEX` before `exec vllm`.
- `.env.dspark.example`, `CHANGELOG.md`, `docs/SETUP.md`.

Opt-out: `NCCL_IB_GID_LEAVE_UNSET=0`.


plotarmordev (changes requested):
Requesting changes. The head-rank empty→unset chain is correct for the main service, but the worker and VL-sidecar paths do not implement the same contract.

1. **High — omitting the remote key resurrects a stale worker pin from `.env.dspark`.** With `AUTO=1` and default `LEAVE_UNSET=1`, `remote_nccl_env()` omits `NCCL_IB_GID_INDEX` when the resolved worker value is blank. Remote Compose still runs with `--env-file .env.dspark`; an absent process variable falls through to that file. Therefore an upgrader's existing `NCCL_IB_GID_INDEX=3` is masked to empty on rank 0 but reappears on rank 1. The launcher logs “Leaving ... unset” while the TP ranks diverge and the worker retains the exact stale pin this PR is intended to remove. Keep emitting `NCCL_IB_GID_INDEX=''` to the remote Compose process so it masks the env file; the new container entrypoint normalization can then convert defined-empty to truly unset.

2. **High — `docker-compose.vl-sidecar.yml` receives defined-empty GID index without unsetting it.** The sidecar maps `NCCL_IB_GID_INDEX` but its command lacks the blank→unset line added only to `docker-compose.dspark.yml`. Under the new default, the head sidecar receives empty (which the PR states NCCL interprets as GID 0/fe80), while the worker can receive the stale env-file pin from finding 1. Apply the normalization to the sidecar path or remove the map entry there.

3. **Medium — the pin-era common-index intersection still blocks the new per-device mode.** `resolve_nccl_gid_indexes()` exits before the leave-unset block when selected HCAs have usable but disjoint RoCEv2 index sets. With the variable absent, NCCL is supposed to choose per HCA, so a single common index is no longer required. The motivating drift state can therefore still fail launch. In leave-unset mode validate that every selected member has a usable RoCEv2/IPv4 GID without requiring one shared index; retain the intersection gate only for pinning modes.

4. **Medium — the new default-on fabric behavior has no permanent regression tests.** Existing passthrough extraction ends at the fourth unset before the new line, and resolver orchestration tests exit before the blanking block. Add tests for legacy pins in the synced env file, both ranks' rendered/container environments, VL sidecar normalization, `LEAVE_UNSET=0`, and `AUTO=0` precedence. No hosted checks are reported for this head.

5. **Medium — `NCCL_IB_GID_LEAVE_UNSET` is not validated.** Anything other than literal `1` silently opts back into the stale-pin behavior. Accept exactly `0|1` and fail with a named configuration error on typos.

The CHANGELOG's unconditional “pin no longer survives into the container” claim should be held until findings 1 and 2 are fixed.

plotarmordev:
P1 recheck: the current head `227ccd2c3f7367ee61f440cf3c826530dcc9ff5d` remains **changes required** and must not merge.

The prior requested changes remain live:
1. AUTO=1 still computes a common GID-index intersection and can exit 3 even when every selected HCA has its own valid RoCEv2/IPv4 index.
2. Omitting the empty worker key allows a stale worker `.env.dspark` pin to reappear during Compose interpolation.
3. `NCCL_IB_GID_LEAVE_UNSET` adds an unnecessary second mode switch.

Required clean contract:
- `NCCL_IB_GID_AUTO=1`: validate every selected member independently, permit disjoint usable index sets, then clear both rank values; always emit the empty remote assignment so it masks stale env-file state; use one shared Compose entrypoint to turn defined-empty into true process absence.
- `NCCL_IB_GID_AUTO=0`: preserve explicit head/worker pins unchanged.
- Remove `NCCL_IB_GID_LEAVE_UNSET` and the obsolete common-index selection path; add resolver, stale-env precedence, and entrypoint regressions.

Validation evidence for that contract: resolver suite 75/75, fabric/Compose suite 8/8, and complete `scripts/ci-validate.sh` passed (27 shell syntax gates, 48 Python compiles, 323 unit tests with 10 declared skips). Live TP=2 AUTO=1 cold boot deliberately supplied stale pins 77/88; both serving processes had `NCCL_IB_GID_INDEX` truly absent, passed text/image/concurrent smokes, and had zero restarts/OOM/NCCL/RDMA errors. AUTO=0 cold boot preserved explicit 3/3 independently on both ranks and passed smoke. The untouched original deployment was restored and re-smoked.

Please rebase/update this PR head to the contract above. Repository policy prevents publishing code or patches from the deployment checkout; this comment contains review/validation evidence only.

plotarmordev:
@drowzeys The current exact head is still `227ccd2c3f7367ee61f440cf3c826530dcc9ff5d`. Please update/rebase this PR to the reviewed acceptance contract above; I can immediately re-run exact-head review and validation when the branch changes.

PR #128, opened 2026-08-24
perf(stage-c): add drafter-private CUDA graph sizes

plotarmordev:
## Summary
- remove the byte-identical, unused proposer copy and stale launcher validation/worker copy
- add Stage-C drafter-private request-count CUDA graph buckets behind strict `VLLM_DSPARK_DRAFT_CAPTURE_SIZES=0|1`
- keep target graph sizes immutable and capture missing drafter buckets largest-first
- register and forward the switch through the Stage-C-only environment surface

This deliberately does not patch the default Anemll 0.1.1 runtime. That image has no legacy `dspark_proposer.py`; its private DFlash graph manager already gives C1 an exact graph, so Tony's single-stream result does not transfer directly.

## Verification
- `./scripts/ci-validate.sh` at exact head `1f35d91`
- merged base + Stage-C compose config validates
- exact-head overlay image `sha256:9ed2b7976a4ae3a2e6cf42c3b21b25ec3a0b5121c31480ebd939d8c6cdf9af03`
- in-image mapping, env registration, bucket resolution, and proposer import assertions passed
- in-image dummy capture orchestration exercised enabled and disabled paths
- six post-implementation FREN reviews: no blocking defect; HELP and CYBERGLM independently approved the Stage-C-only, default-off design

## Gate before ready-for-review
A two-node Stage-C A/B must measure C1/C2/C4/C6 throughput, acceptance, graph memory, KV capacity, startup allocation failures, prefill, and stable quality probes. Extra drafter graph memory is outside target profiling. Leave the switch off unless that gate clears.

## Hosted CI follow-up
If hosted Actions is unavailable at merge time, re-run `.github/workflows/validate.yml` on the merged branch as soon as capacity returns and record the result.


plotarmordev:
Superseded by #129. GitHub closed this stacked PR when its temporary base branch was deleted after #127 merged; the unchanged head is now targeted directly at main.

PR #129, opened 2026-08-24
perf(stage-c): add drafter-private CUDA graph sizes

plotarmordev:
## Summary
- remove the byte-identical, unused proposer copy and stale launcher validation/worker copy
- add Stage-C drafter-private request-count CUDA graph buckets behind strict `VLLM_DSPARK_DRAFT_CAPTURE_SIZES=0|1`
- keep target graph sizes immutable and capture missing drafter buckets largest-first
- register and forward the switch through the Stage-C-only environment surface

This deliberately does not patch the default Anemll 0.1.1 runtime. That image has no legacy `dspark_proposer.py`; its private DFlash graph manager already gives C1 an exact graph, so Tony's single-stream result does not transfer directly.

## Verification
- `./scripts/ci-validate.sh` at exact head `1f35d91`
- merged base + Stage-C compose config validates
- exact-head overlay image `sha256:9ed2b7976a4ae3a2e6cf42c3b21b25ec3a0b5121c31480ebd939d8c6cdf9af03`
- in-image mapping, env registration, bucket resolution, and proposer import assertions passed
- in-image dummy capture orchestration exercised enabled and disabled paths
- six post-implementation FREN reviews: no blocking defect; HELP and CYBERGLM independently approved the Stage-C-only, default-off design
- hosted `validate` succeeded on the exact head before the stacked base merged

## Gate before ready-for-review
A two-node Stage-C A/B must measure C1/C2/C4/C6 throughput, acceptance, graph memory, KV capacity, startup allocation failures, prefill, and stable quality probes. Extra drafter graph memory is outside target profiling. Leave the switch off unless that gate clears.

PR #128 was automatically closed when its temporary stacked base was deleted after merge; this PR retargets the unchanged exact head to `main`.


plotarmordev:
Attempted the required two-node Stage-C A/B today on the maintainer cluster (2× GB10). **It cannot boot on current main** — the entrypoint/launcher has drifted Anemll-only since the Vision-Exp era. Blocker chain (each fixed, next appears):

1. Launcher k-guard rejects the 0731 lane's k=5 (`MTP_NUM_TOKENS >= 5 and divisible by 3`, unconditional — not keyed to the model's actual `n_predict`).
2. #117 SHM-ring preflight targets Anemll-layout paths (`/usr/local/lib/python3.12/dist-packages/...` missing on the Stage-C venv layout).
3. Encoder install `cp`s to the hardcoded Anemll path (`vllm/tokenizers/` exists at `/opt/env/lib/...` on Stage-C); also `ENCODING_SOURCE` env is clobbered by `DSPARK_ENCODING_FILE` at entrypoint start.
4. Ungated bare-`python3` issue55 hotfix call — Stage-C image has python only at `/opt/env/bin`.
5. Hardcoded Anemll-path model-file existence check (`.../models/deepseek_v4/nvidia/model.py` FATAL).
6. Final `exec /usr/local/bin/vllm serve` — wrong binary path on the venv image (Stage-C CMD is `sleep infinity`, no entrypoint).

I prototyped venv-aware entrypoint deltas (PATH + issue55 guard + exec selection) to get further; each fix exposes the next assumption. Conclusion: **the PR's own gate ("two-node Stage-C A/B must measure...") is unreachable on current main** — the lane needs a lane-aware entrypoint (or a documented Stage-C boot path with its own compose command) before the A/B can produce numbers.

Build side worked: base `ghcr.io/bjk110/vllm-spark:unholy-fusion-prod-ready` pulled (plus a docker-save/load transfer between nodes when WAN flaked), the 3-stage overlay build completes in ~20s on cached layers, and both rank images self-test OK with identical `dspark.py` bytes (sha256 b989d6c6…).

Suggest keeping this draft and adding a `stage-c-entrypoint` compat layer (or pinning the last lane-compatible launcher). Happy to contribute the prototype deltas if useful.

plotarmordev:
Closing as not-pursued (author decision, 2026-09-04). Rationale: the drafter-private bucket problem is Stage-C-only — the pinned Anemll lane's DFlash graph manager already gives C1 exact graphs, so the feature has no effect on the maintained lane. The Stage-C lane itself (vLLM 0.21.1 recipe) no longer boots on current main's entrypoint (five verified Anemll-layout divergences, documented in the comment above: k-guard, #117 preflight, encoder path, ungated python3 hotfix, model-path/exec paths). Reviving it means rebuilding the lane's entrypoint contract — a separate project, not a PR fix. The build-side work (base pull, node transfer, overlay build, self-test, image parity) is recorded above for anyone who resurrects the lane. Thank you — the analysis above remains available for reference.

PR #130, opened 2026-08-24
Warm sampler compile variants during boot

de1tydev:
## Why

The boot-shape sweep covered prefill/spec-decode kernels but never dispatched Triton's `_topk_topp_kernel`: every existing arm used greedy sampling (`temperature: 0`), which drops the top-k/top-p tensors. A first real sampled request could therefore compile on both TP ranks during serving.

## What

Add explicit sampled warmup arms for the three enumerable compile-time branches:

- top-k only;
- top-p only;
- top-k + top-p.

Each branch gets one C=1 request plus one C=3 redundancy burst. The C=3 burst is dispatch redundancy: cold-cache validation showed C=1 alone did not reliably materialize all three entries. It is not another compile-key axis.

After the arms, the warmup inspects `_topk_topp_kernel.ttir` entries in the head rank's persistent Triton cache and reports the postcondition. `BATCH_SIZE` is a plain runtime TTIR argument in the pinned image; it is not a compile key and no second batch-size class is required.

The host-side postcondition is explicitly a head-cache proxy. Both TP ranks execute the sampler, but this script can directly inspect only the head node's host cache. Acceptance evidence therefore inspects both rank-local caches independently.

## Exact-head TP=2 evidence

Head `fd9b10eb24503370b45f3ee790da429ff464f2a7`:

- cold boot: all three constexpr combinations present on both ranks (24 sampler files per rank);
- 600.649-second C=1/C=3 matrix: 1,996/1,996 HTTP 200 responses, zero failures (`k=668`, `p=664`, `k+p=664`);
- each rank's full sampler manifest was identical before and after the matrix;
- same-cache recreate produced fresh head and worker containers, all three combinations remained present on both ranks, and both full sampler manifests remained byte/metadata-identical;
- six-way smoke passed 6/6 before and after recreate.

The earlier odd-batch indexer arms are not part of this PR. Issue #135 closes that pointer-alignment specialization engine-side.

No merge is requested by this update.


de1tydev:
Live validation on the experimental 2× DGX Spark pair (TP=2, deployed `MAX_NUM_SEQS=6`), three consecutive paired restarts on this branch's content:

- Sweep: **52/52 ok** each boot (34s / 32s / 24s); all new arms fired and completed (samp-k/p/kp, samp-c3, short-c3, short-c5, ragged-c3 visible in the arm log with their profiles).
- Persistent Triton cache: stable at 258 entries across all three boots — the new arms created **zero new compile keys on this pair**, because the keys they target are exactly the ones that already leaked mid-serve on 2026-08-23 (that's the honest reading: on a warm volume the arms are no-ops; their value is deterministic boot-time coverage on a cold cache / new image / new Triton hash, where these combos would otherwise compile mid-serve again).
- Post-sweep serve traffic + decode sentinel: zero new jit_monitor warnings beyond the known per-process `_prepare_dflash_inputs_kernel` first-touch line (which fires on disk-cache hits too — worth knowing when using jit_monitor as a closure signal); single-stream decode 64.5–72.1 tok/s, spec acceptance 0.69–0.71, same band as pre-change baselines.

One observation for the engine-side follow-up: TTIR attrs confirm the `_compute_global_topk_indices_and_lens_kernel` family keys purely on per-step buffer alignment, so request-side arms can only diversify, not enumerate — aligned/padded allocation of `token_to_req_indices`/`is_valid_token` would collapse the family to O(1) keys. Happy to file that as its own issue if wanted.

https://claude.ai/code/session_01HXtETnbPgS5yPWvWethDHN

plotarmordev (changes requested):
Exact-head Nightshift admission is blocked (`REWORK`) at `ea093f55fd017e24c08e52fe7d1c43d97a1fa5cd`. The head is based on `d1b7625` while current main is `d5ae65c`, is merge-dirty, and still carries indexer warmup arms superseded by merged #135. Please rebase onto current main, remove the now-obsolete indexer arms, add non-one-batch cases for top-k-only and top-p-only (not only k+p), and make the expected `_topk_topp_kernel` cache signatures an observable postcondition so three concurrent curls cannot falsely pass as three BATCH_SIZE=1 steps. After that, the exact-head A/B should use isolated cold Triton caches on both ranks and prove all six sampler signatures are created during warmup with zero target-key/hash delta during a 10-minute B1/B3 matrix and same-cache recreate. No merge action taken.

de1tydev:
Reworked at `d8751cc`, addressing all four review points:

1. **Rebased onto current main** (merge-clean).
2. **Indexer arms removed** — the odd-batch/ragged bursts existed only to chase `_compute_global_topk_indices_and_lens_kernel`'s pointer-alignment keys, which #135's `do_not_specialize_on_alignment` fix closes engine-side (the deterministic closure those arms could only approximate). What remains is purely the sampler family.
3. **top-k-only and top-p-only at a non-one batch** — the C=3 sampler bursts now run all three constexpr combos (`samp-k-c3`, `samp-p-c3`, `samp-kp-c3`), not just k+p; gated on the resolved `MAX_NUM_SEQS` (≥3) like every other arm. Honest accounting: 47 requests at the shipped C=6 profile, 35 at C=4, 18 at C=2.
4. **Observable cache-signature postcondition** — the sweep no longer assumes three concurrent curls formed one `BATCH_SIZE=3` step. After the arms it classifies the node's persistent Triton-cache TTIR for `_topk_topp_kernel`: the `TOPK/TOPP_ENABLED` constexprs are folded out of the signature, so combos are observed through argument *use* (`%K`/`%P` referenced beyond declaration, with a boundary guard so `%P` never matches `%PERCENTILE_TO_STD_TABLE`), and the `BATCH_SIZE` class is read literally off the declared argument (absent/attributed/plain). Requirement: every combo present in two distinct `BATCH_SIZE` classes (one below C=3). On a miss the C=3 bursts are retried (bounded ×2, accounting-exact — retries overwrite the same per-arm result files), and a still-missing class fails the non-fatal sweep with a precise diagnostic. The launcher hands the sweep the host-side cache path derived from `HF_CACHE` for the compose-default layout; custom `TRITON_CACHE_DIR` layouts skip the check with an explicit note.

CPU coverage went from 63 to 72 assertions: sampler-arm placement and non-leak into other arms, C=3 gating at lower profiles, and the postcondition's MET / retry-converges / retry-exhausts / skip paths against fixture TTIR caches (including tally exactness under retries). `bash scripts/ci-validate.sh` passes on the new head.

The exact-head A/B per the review — isolated cold Triton caches on both ranks, all six sampler signatures created during warmup, zero target-key/hash delta across a 10-minute B1/B3 matrix, then a same-cache recreate — is staged on the experimental pair; results will follow in a separate comment once its maintenance window runs.

de1tydev:
Exact-head A/B complete on the experimental pair at `6917a53` — and first, thanks for the pushed fix: our own cold-cache run on the prior head reached the same conclusion independently before we saw your commits. On that run the postcondition did its job in the most useful way possible: it honestly refused to pass — `unmet — k-only:1/2 p-only:1/2 k+p:1/2` through all bounded retries — because the "six signatures" model this PR previously asserted **does not exist on the shipped MTP-5 config**. Ground truth from that cold cache: exactly **three** `_topk_topp_kernel` entries (one per constexpr combo), every one carrying a plain `%BATCH_SIZE: i32` argument (no ==1 fold, no divisibility attribute), and two of the three hashes byte-identical to the wild mid-serve entries from 2026-08-23 production. Under spec decode BATCH_SIZE is the step's sampled-row total (≥ 1+num_spec per request), so it is never 1, and C=1 vs C=3 land in the same class; the divisibility-16 residue seen in the wild comes only from traffic-dependent row totals. Your combos-existence postcondition with C=3 as dispatch redundancy is the right semantics; we adopted the pushed head as-is.

A/B per the review protocol, exact head `6917a53`, isolated **fresh** `TRITON_CACHE_DIR` on both ranks:

- **Cold boot**: sweep 47/47 ok in 32 s; `sampler-cache postcondition: MET` with combos {k-only, p-only, k+p}; 254 cache entries per rank, rank listings equal.
- **10-minute B1/B3 matrix**: 729 rounds (~1,460 requests) alternating single-request and 3-concurrent rounds while rotating greedy / top-k-only / top-p-only / k+p profiles, zero request failures — and **zero cache-key delta on both ranks**, measured over the entire cache listing, not just the sampler family.
- **Same-cache recreate**: sweep 47/47 in 31 s, postcondition MET, zero new keys on both ranks.
- **Production landing** (deploy tree = main + this head + #131 + #132): postcondition MET against the long-lived production cache — and the plain k-only entry (`MUYH…`), which production traffic had *never* materialized, compiled once inside the `samp-k` arm: precisely the boot-window closure this PR exists to provide. Decode sentinel after landing: 62.6–70.3 tok/s, spec acceptance 0.68–0.70 (healthy band).

Engine-side follow-up worth its own issue, analogous to #135's alignment fix: `do_not_specialize` on `BATCH_SIZE` would collapse the remaining traffic-dependent residue axis (plain vs div-16) to one key per combo and make the family fully deterministic. Happy to file it if wanted.

plotarmordev:
## Maintainer exact-head admission evidence

No merge or review-dismissal action was taken.

**Exact head:** `fd9b10eb24503370b45f3ee790da429ff464f2a7` (PR author/head owner remains @de1tydev)

### Verification

- CPU changed-contract test: `bash scripts/test-boot-shape-warmup.sh -q` → `67 ok, 0 failed`.
- Hosted `validate` completed successfully at this exact head: run `32836484588`, job `97766379463`.
- Cold TP=2 boot materialized the three intended `_topk_topp_kernel` TTIR combinations on **both** ranks: `k-only`, `p-only`, and `k+p` (24 sampler-cache files per rank).
- Exact-head C=1/C=3 stress ran for `600.649 s`: **1,996/1,996 HTTP 200**, zero failures (`k=668`, `p=664`, `k+p=664`; max recorded latency `0.975 s`).
- Separate head and worker sampler manifests were identical before and after the stress run: paths, sizes, hashes, inodes, and mtimes did not change.
- Same-cache recreate produced fresh head and worker containers. Both ranks still held all three combinations; both full manifests remained identical to the pre-recreate manifests.
- Six-way smoke passed `6/6` before and after recreate.

The source now states the actual boundary: every TP rank executes the sampler, while the shipped host-side postcondition can inspect only the head cache and is therefore a proxy. The retained acceptance evidence independently inspects both rank-local caches. `BATCH_SIZE` is a runtime TTIR argument, not a compile-key axis; C=3 remains dispatch redundancy only.

### Independent review

- Final CYBERGPT rereview: **READY** at this head; both earlier technical blockers closed.
- FREN3 fallback cross-PR audit: **READY** for per-PR technical admission; no material contradiction or attribution loss found.

### Remaining gates / limits

- GitHub still reports `CHANGES_REQUESTED` from an older pre-fix review. It must be explicitly superseded/dismissed and replaced with a current-head approval before merge.
- The CHANGELOG deliberately underclaims closure (`pending exact-head A/B`); the evidence above now supplies it.
- Custom Triton-cache layouts that cannot be mapped to the host HF cache retain the documented diagnostic skip.
- This proves the known `TOPK_ENABLED × TOPP_ENABLED` family, not every possible future specialization.
- All three open PRs (#130/#131/#132) insert at the same top-of-file CHANGELOG location. After any one merges, the others will require conflict resolution/rebase, new exact heads, validation, and renewed review.

**Verdict:** technically ready at the exact head above; **not merged**.

plotarmordev (approved):
Approved at exact merged head 2881406ae7f1f17902aba37d20e889e2981c7d81 against main 3a049db08a9443a2fcc9524d988c1e55324bb8a5. Local ci-validate and hosted validate pass; exact-head review is READY; scope is four intended files; secret-pattern scan is clean; the additive CHANGELOG resolution preserves PR #131 and de1tydev authorship. Residual limitation remains documented: cache attestation covers the mapped head-rank cache and skips unmappable custom layouts; this PR is mitigation, not issue #117 closure.

PR #134, opened 2026-08-24
Fix global top-k Triton recompilation family

### LINKED ISSUE #133: [Enhancement] Align/pad per-step token buffers so _compute_global_topk_indices_and_lens_kernel stops re-specializing on pointer alignment (issue #117 family)
## Problem

`_compute_global_topk_indices_and_lens_kernel` (deepseek_v4 `common/ops/cache_utils.py`) keeps growing new Triton compile keys mid-serve — two leaked on a fully-warmed pair on 2026-08-23 and stalled traffic behind a mid-serve compile (reported on #117). Warmup can't close this family deterministically: PR #130 adds odd-batch/ragged arms, but they diversify rather than enumerate, because the key axis isn't a shape bucket at all.

## Evidence

TTIR function-signature diff across all six live persistent-cache entries for this kernel (2× DGX Spark pair, image `0.1.1`, vLLM 0.25.2.dev): every entry is identical **except** the `tt.divisibility = 16` attribute on

- `token_to_req_indices_ptr`
- `is_valid_token_ptr`
- `block_size` (int residue)

i.e. Triton re-specializes purely on whether those per-scheduler-step buffers happen to land 16-byte aligned, which depends on step token composition. The theoretical family is bounded (2³ × block_size variants) but which members appear is traffic-dependent, so any sweep is best-effort by construction — and each first appearance is a mid-serve JIT with the TP-desync hazard #117 documents.

## Proposed fix (engine-side, deterministic)

In the wrapper that launches this kernel (and its sibling per-step buffers):

1. Allocate `token_to_req_indices` / `is_valid_token` (or the parent workspace they're sliced from) with 16-byte-aligned starts per slice — e.g. round the per-step slice offsets up to a multiple of 16 bytes, or allocate fresh contiguous tensors instead of offset views. Masked loads already tolerate the padding tail.
2. Pass `block_size` as `tl.constexpr` (it's config-constant per deployment), removing the int-residue axis.

Result: the compile-key family collapses to O(1) members, all materialized by the existing boot sweep, and this kernel exits the mid-serve-JIT population entirely.

## Trade-offs

- Slight allocation padding (bytes per step).
- If the buffers are views into a shared workspace, alignment rounding changes offset arithmetic in one place; no kernel body changes needed (divisibility hints only get *better*).

Happy to draft the patch if the approach sounds right — flagging the mechanism first since the allocation site is maintainer territory.

Related: #117 (hang chain), #130 (request-side best-effort arms + TTIR extraction method), #131/#132 (cache persistence / forensics).

https://claude.ai/code/session_01HXtETnbPgS5yPWvWethDHN

plotarmordev:
## Summary

- disable Triton alignment specialization for the sliced `token_to_req_indices_ptr` and `is_valid_token_ptr` metadata pointers
- port the stable `tl.constexpr` annotations from merged vLLM PR [#51967](https://github.com/vllm-project/vllm/pull/51967)
- add a driverless Triton 3.6 regression test for cache-key closure and SM121 compilation, plus an optional CUDA numerical case

Closes #133.

## Why this fixes the six-entry family

Mixed prefill slices the `int32` token-to-request map and the `bool` validity mask at the same token offset `k`.

- the `int32` pointer is 16-byte aligned when `k % 4 == 0`
- the `bool` pointer is 16-byte aligned when `k % 16 == 0`

Because both slices share `k`, only three pointer states are possible: both aligned, only the `int32` pointer aligned, or neither aligned. The observed effective block sizes 64 and 2 form the other compile-time axis, producing the six persistent-cache entries from #133:

`3 pointer states × 2 block sizes = 6 variants`

Both affected pointers feed scalar loads only. Their 16-byte alignment does not enable vectorized memory access, so specializing on it creates traffic-dependent compile keys without improving this kernel. Current vLLM uses the same narrow decorator for sliced scalar-loaded metadata in its [MiniMax M3 index kernel](https://github.com/vllm-project/vllm/blob/79bb395eea64dbfef99a55f010d2854db71f8571/vllm/models/minimax_m3/common/ops/index_topk.py#L88-L92).

The patch therefore keeps the existing buffers and callers intact:

```python
@triton.jit(
    do_not_specialize_on_alignment=[
        "token_to_req_indices_ptr",
        "is_valid_token_ptr",
    ]
)
```

Stable strides, `topk`, and `block_size` become `tl.constexpr`, matching upstream vLLM #51967. That leaves two deterministic compiled variants, one for each effective block size, instead of padding fixed-address CUDA-graph metadata or copying slices per scheduler step.

## Verification

Exact commit `88c4c4e`:

- `docker run --rm -v "$PWD:/work:ro" --entrypoint python3 ghcr.io/anemll/dspark-vllm-gx10:0.1.1 /work/tests/test_issue133_triton_specialization.py -v`
  - Triton 3.6 specialization-key test: passed
  - SM121 driverless compile for block sizes 64 and 2: passed
  - CUDA numerical test: skipped because this validation intentionally did not attach a GPU
- `python3 -m unittest discover -s tests -v`: 13 passed/skipped as expected; no failures
- `ruff check tests/test_issue133_triton_specialization.py`: passed
- `ruff check --ignore I001,RUF100 recipe/overlay/vllm/models/deepseek_v4/common/ops/cache_utils.py`: passed; the ignored findings predate this patch
- pushed-commit secret scan: no matches

The driverless test constructs same-slice offsets `0`, `4`, `1`, and `16`, covering both aligned, `int32`-only aligned, and neither aligned. It proves those offsets share one specialization for each block size and that `block_size` is the only remaining key difference.

## Live gates before merge

This draft deliberately does not claim live serving closure yet. Before merge on a spare GPU/pair:

1. run the included CUDA numerical case for offsets `0`, `1`, `4`, and `16` with block sizes 2 and 64;
2. start from an isolated empty Triton cache and confirm all former combinations produce exactly two entries whose mtimes remain stable;
3. run a fresh-cache TP=2 warmup plus mixed prefill/decode soak and require zero post-warmup compilation of this kernel, pair loss, or NCCL watchdog termination.

This closes the dynamic-JIT family proven in #133. It does not claim to resolve every silent-stall mechanism tracked by #117.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+## 2026-08-24
+
+### Fixed
+
+- **Bounded `_compute_global_topk_indices_and_lens_kernel` to two deterministic Triton compile variants ([Issue #133](https://github.com/MiaAI-Lab/DeepSeek-v4-Flash-DSpark-2x-DGX-Spark/issues/133))**: mixed prefill slices the `int32` token-to-request map and `bool` validity mask at the same traffic-dependent token offset, so their pointer-alignment specialization produced three cache-key states; the observed effective block sizes 64 and 2 doubled that family to six. The kernel now disables alignment specialization only for those two scalar-loaded pointers and treats its stable strides, top-k width, and block size as compile-time constants, matching upstream vLLM's kernel optimization without padding fixed-address CUDA-graph metadata or copying slices per step. A driverless Triton 3.6 regression test proves the three alignment classes share a key, only block size remains as a two-member axis, and both variants compile for SM121; an optional CUDA case checks bit-exact outputs across all former alignment classes when a device is available.
+
 ## 2026-08-23
 
 ### Added
--- a/recipe/overlay/vllm/models/deepseek_v4/common/ops/cache_utils.py
+++ b/recipe/overlay/vllm/models/deepseek_v4/common/ops/cache_utils.py
@@ -430,18 +430,23 @@ def compute_global_topk_indices_and_lens(
     return global_topk_indices, topk_lens
 
 
-@triton.jit
+@triton.jit(
+    do_not_specialize_on_alignment=[
+        "token_to_req_indices_ptr",
+        "is_valid_token_ptr",
+    ]
+)
 def _compute_global_topk_indices_and_lens_kernel(
     global_topk_indices_ptr,
-    global_topk_indices_stride,
+    global_topk_indices_stride: tl.constexpr,
     topk_lens_ptr,
     topk_indices_ptr,
-    topk_indices_stride,
-    topk,
+    topk_indices_stride: tl.constexpr,
+    topk: tl.constexpr,
     token_to_req_indices_ptr,
     block_table_ptr,
-    block_table_stride,
-    block_size,
+    block_table_stride: tl.constexpr,
+    block_size: tl.constexpr,
     is_valid_token_ptr,
     num_blocks,
     TRITON_BLOCK_SIZE: tl.constexpr,
--- a/tests/test_issue133_triton_specialization.py
+++ b/tests/test_issue133_triton_specialization.py
@@ -1,0 +1,236 @@
+#!/usr/bin/env python3
+"""Driverless regression test for issue #133's Triton specialization family."""
+
+from __future__ import annotations
+
+import importlib.util
+import sys
+import types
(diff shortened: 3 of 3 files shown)


MiaAI-Lab:
Superseded by #135 (overlay + Anemll startup hotfix + live TP=2 soak).

PR #137, opened 2026-08-25
Add ablation support

mrexodia:
Only tested briefly, but appears to work on previously-refused requests.


plotarmordev:
Validation status for exact head `adc97c8b7b0711cef15bdf226503839c75d2ea4b`:

- `python3 scripts/test-runtime-ablation.py`: 10/10 passed.
- `bash scripts/selftest-runtime-ablation.sh` against the pinned image: passed; patch applied one attention hook, direction SHA matched, and the numerical gate reported `fp32_error=1.91e-06`, `bf16_error=4.09e-03`.
- Independent source review also confirmed fail-closed anchor/site-count handling, dual-node SHA-verified direction staging, off-mode stale-patch rejection, and launcher/config validation rejecting `ABLATE=1` with `ABLITERATED=1`.

This remains P3 and is not admitted to live TP=2 testing yet. The capability changes trained refusal behavior and redistributes a direction artifact, so owner policy approval is required before engineering admission. If approved, the remaining gate is a fresh-cache TP=2 behavioral/throughput/JIT A/B under the #117 stability standard. The unconditional boot-path hotfix dependency (even at `ABLATE=0`) and direct container-env bypass of the documented layer cap should also be resolved or explicitly accepted. `ok-to-test` is intentionally absent; higher-priority P1 candidates are being handled first. No merge action taken.

mrexodia:
Loving the clanker gates 😆

plotarmordev:
yep clanker written, human approved after review

plotarmordev:
Thank you for the runtime-ablation work. Closing as the earlier duplicate of #182: the runtime patcher and `direction_r1.pt` blobs are byte-identical between the two proposals, and #182 has already received the final disposition.

This approach is distinct from main's supported `ABLITERATED=1` alternate-checkpoint path: it avoids the second checkpoint download by projecting decoder activations while running the official checkpoint. We are not maintaining that additional path without evidence that its download/storage benefit outweighs the per-layer FP32 hot-path cost, pinned model-source patching, direction-artifact provenance/versioning, and extra support surface.

If avoiding the checkpoint download becomes a required constraint, a new current-main proposal should include reproducible direction provenance and hashes plus live A/B evidence for TTFT/decode, memory, long context, tools, images, and refusal reduction.

PR #142, opened 2026-08-26
diagnostics: add opt-in issue #117 TP dispatch trace

plotarmordev:
## Purpose

This is opt-in diagnostic instrumentation for #117, **not a root-cause fix**. It is intended to distinguish a missing TP dispatch from a worker/device stall or a lost response during the controlled clean-launch versus worker-only dirty-restart experiment.

## What changes

- adds `patches/hotfix-vllm-issue117-dispatch-trace.py`
- logs driver dispatch/response and worker receive/call/response phases with existing FIFO sequence and shared-memory ring positions
- wires `DSPARK_ISSUE117_DISPATCH_TRACE=1` through compose with fail-closed startup
- syncs the patch to the worker and documents the experiment contract
- adds focused behavior, idempotency, drift-refusal, and gated-boot coverage

## Safety contract

- default is `DSPARK_ISSUE117_DISPATCH_TRACE=0`; stock executor source remains untouched
- no RPC payload or synchronization protocol changes
- no prompts, request bodies, tensors, or secrets are logged
- the patch is source-locked to **vLLM 0.25.2.dev0** in **`ghcr.io/anemll/dspark-vllm-gx10:0.1.1`**
- on any other build, anchor drift is detected and container startup aborts by design rather than applying partial instrumentation

## Controlled experiment

Enable the flag on both members, archive both ranks' logs, and run identical traffic after:

1. a clean paired launch
2. a worker-only dirty restart

Interpretation:

- `disp` without matching `recv`: dispatch/ring delivery path
- `recv` without `done` or `fail`: worker/device execution
- `done` without `resp`: async device-output wait
- `resp` without driver response progress: response transport

## Verification

- `python3 tests/test_issue117_dispatch_trace.py -q` — 3 passed
- `python3 scripts/test-python-hotfix-failclosed.py -q` — 10 passed
- pinned-image apply → `--status` → compile → import smoke — passed
- `bash scripts/ci-validate.sh` — passed
- changed-file and pushed-range secret signature scan — no findings

## Known gap

The live TP=2 clean-versus-dirty-restart experiment has not been run; this draft exists to make that evidence collection reviewable before asking the affected operator to run it.

Refs #117


plotarmordev:
Maintainer review at exact head `9b7e8c626071d6e2589eac79f91dffb8e60e8ce0`: the scope is appropriate for a **draft diagnostic**, not a fix. The trace is opt-in, source-locked to the pinned Anemll vLLM build, and does not intentionally change RPC payloads or synchronization.

Keep this draft until it has paired live evidence for both:
1. a clean paired launch; and
2. the worker-only dirty-restart sequence described in #117.

Acceptance should prove event correlation across dispatch, worker receive, callable completion, response enqueue, and driver response without logging prompts/tensors. Source drift must continue to abort rather than partially instrument. No merge approval at this stage.

plotarmordev:
Closing this diagnostic draft as superseded by the issue #117 resolution now on `main`.

The merged source-exact SHM reader recovery bounds authoritative-flag rechecks and releases read slots on consumer exceptions; the associated issue is closed after validation. The proposed dispatch trace was valuable for localizing the failure before that fix, but retaining an 827-line source-patched diagnostic path now would add maintenance and boot-failure surface without a current experiment that needs it.

If the same dispatch-loss signature recurs on current main, please open a fresh, narrowly scoped diagnostic proposal against the then-current pinned image and include the new incident evidence.

Issue #143, opened 2026-08-26
[Bug] MAX_NUM_SEQS=16: first c=16 burst reliably kills the engine on 70a7cc4 (2/2 boots) — N=8 and N=32 identical bursts are clean


sethforprivacy:
## Summary

On current main (`70a7cc4`), a 2x DGX Spark TP=2 pair configured with **`MAX_NUM_SEQS=16` + `MAX_NUM_BATCHED_TOKENS=12288`** dies on its first c=16 burst — reproduced on **2/2 independent boots**. The same burst is clean at `MAX_NUM_SEQS=32` and `MAX_NUM_SEQS=8` on the same tree, same nodes, same everything else. This matters because `.env.dspark.example` itself suggests `MAX_NUM_SEQS=16` as the higher-throughput profile.

## Environment

- 2x DGX Spark / GB10, TP=2 over CX-7 RoCE (MTU 9000, GID 3)
- Recipe `70a7cc4`, image `ghcr.io/anemll/dspark-vllm-gx10:0.1.1` (`3430d661`), vLLM `0.25.2.dev0+g752a3a504`
- `nvfp4_ds_mla` KV, MTP dspark k=6 probabilistic, `LONG_PREFILL_TOKEN_THRESHOLD=1024`, boot-shape warmup completed before every probe
- Probes: 16 concurrent short-prompt chat completions, max_tokens=500, thinking=low

## The matrix (all on 70a7cc4, bt=12288, k=6)

| MAX_NUM_SEQS | burst | verify rows | result |
|---|---|---|---|
| 16 | c=8 | 56 | **healthy** — 154.3 tok/s aggregate |
| **16** | **c=16** | 112 | **ENGINE DEATH — 2/2 boots** (0/16 complete) |
| 32 | c=16 | 112 | healthy — 213.0 tok/s |
| 32 | c=32 | 224 | healthy — 325.5 tok/s |
| 8 | c=8 / c=16 | 56 | healthy — 149.4 / 147.7 tok/s |

Notes on what this rules out: it is **not** verify-batch size (112 rows dies at N=16 but is clean at N=32; 224 rows is clean), **not** a full-batch condition (N=32 c=32 and N=8 c=8 are both full and clean), and **not** load per se (N=32 sustains far more). The failing cell is specifically `MAX_NUM_SEQS=16`.

## Failure signature (second reproduction, logs preserved)

[...]

Issue #141, opened 2026-08-25
[Bug] TP=2 pair deadlocks in sparse-MLA decode once the DSpark verify batch exceeds flashinfer's _DECODE_MAX_TOKENS (64 rows)


kosztyua:
## Summary

On 2× DGX Spark (GB10, sm_121a), TP=2, DSpark k=5, the pair deadlocks under
sustained concurrent **long** generations. PyTorch's 600 s `ProcessGroupNCCL`
watchdog kills both ranks, `restart: unless-stopped` brings them back, so the
visible symptom is a ~20-minute silent outage with `RuntimeError: cancelled`.

The hang happens inside FlashInfer's SM120 sparse-MLA decode, on the **prefill
fallback** path, in the DSpark verify pass. `flashinfer/mla/_sparse_mla_sm120.py`:

```python
_DECODE_MAX_TOKENS = 64
# Decode/prefill cutoff: num_tokens > _DECODE_MAX_TOKENS routes to the ... prefill path
```

`num_tokens` there is the verify batch, `MAX_NUM_SEQS × (k+1)`. At k=5 **any
deployment above 10 concurrent sequences** exceeds 64 rows and leaves the
specialised `decode_dsv4` kernel for the generic prefill fallback.

Being on that path is **necessary but not sufficient**:

- **Necessary** — chunking every call to 64 rows survives 3,145,728 tokens;
  chunking to **65**, one row over, dies in round 1 with zero requests
  completed. Reproduced at two concurrency levels.
- **Not sufficient** — called directly with the same shapes the fallback never
  hangs, at any size we tried, including with 47 of 48 SMs held by a spin kernel.

So it is an interaction with something in the live TP=2, 43-layer, dual-stream
context. **We localised the trigger but did not isolate the mechanism.**

This is **not** #117. PRs #130/#131 fixed that one for us (16-stream
long-generation load went from dying 3/3 to surviving 393,216 tokens). Here there
is no `jit_monitor` warning near the hang — the last compiles finish 13 minutes
before it.

## Environment

2× DGX Spark (GB10, sm_121a), TP=2 over ConnectX-7 RoCE, driver 580.173.02,
CUDA 13.0, kernel 6.17.0-1029-nvidia. `ghcr.io/anemll/dspark-vllm-gx10:0.1.1`
(vLLM 0.25.2.dev0), this recipe at `f104c39` + PRs #130/#131 applied.
[...]

PR #145, opened 2026-08-26
docs(env): document the admission/batched-token measurements as performance data, with an explicit #141 stochastic-stall caveat

sethforprivacy:
## What

Docs/comments only — no behavior change. Rewritten under @plotarmordev's HOLD
(comment 5440225676) after the later repeated-burst evidence in #141
invalidated this PR's original safety framing.

**Removed** — every claim that any `MAX_NUM_SEQS` value is safe:

- the `WARNING: do NOT use MAX_NUM_SEQS=16 ... 8 and 32 are both verified
  clean` block, and
- the #143 admission-cap mechanism it rested on, which I have retracted in
  #141: every cell in that matrix was n=1.

## Why the original framing was wrong

- **The "safe" cell is not safe.** N=32 / c=32 (224 verify rows at k=6) ran
  clean for three bursts and **died on burst 4** on my pair; on @kosztyua's
  pair the same cell died on burst 2 and again on a fresh-boot first burst.
- **The proposed clean bracket is not universal.** N=16 / k=5 (96 rows) survived
  3x393,216 tokens on their pair and **died at ~237K tokens** (16/32 completed)
  on mine.
- **Wall times cluster regardless of config** — 706, 727, 740, 746, 759, 781,
  897, 899 s — with an identical `shm_broadcast.py` `acquire_read` →
  `RuntimeError("cancelled")` signature, no NCCL error, no JIT warning, no OOM.

A single clean burst cannot distinguish a safe configuration from a lucky one.
The failure is stochastic per burst and sensitive to stream lifetime, so a
configuration change alters incidence at best.

## What the block says now

- The `MAX_NUM_SEQS=32` + `MAX_NUM_BATCHED_TOKENS=12288` numbers stay, clearly
  labeled **performance data from bursts that completed**: c=1 47–54 tok/s,
  c=8 154, c=16 213, c=32 325 aggregate (2x DGX Spark TP=2 @ `70a7cc4`, k=6,
  `nvfp4_ds_mla`, short prompts, thinking=low, 500-token generations) vs ~108
  tok/s at c=8 with shipped-class admission; single-stream decode is recovered
  by the batched-tokens raise (the draft-slot squeeze the engine already warns
  about at boot).
- **The KV-pool correction stands on its own** and is the one durable finding
  here: the pool is nearly `MAX_NUM_SEQS`-invariant (~2.6x → ~2.1x concurrent
  full-1M requests from N=4 to N=32), so the old "drop `MAX_MODEL_LEN` to 200K
  to get concurrency" suggestion overstated the context cost regardless of how
  #141 resolves. That is what this PR was originally for.
- A `STABILITY` block stating plainly that **no tested `MAX_NUM_SEQS` is
  established as generally safe**, with the observed axes (verify rows
  `MAX_NUM_SEQS * (k+1)` × stream lifetime × possibly cap saturation), the
  chunk-64-survives / chunk-65-dies pair as the only clean kernel-path
  localizer, and the severity spectrum whose mild end is **silent stream
  truncation** (`finish_reason` absent, zero preemptions, zero queue depth — no
  existing metric fires; detection needs an absent-`finish_reason` check).
- Context, not endorsement: my production pair runs `MAX_NUM_SEQS=13` (91 rows)
  as an interim with no claim it is proven safe, and #151's opt-in
  `DSPARK_ENABLE_ISSUE141_SPARSE_MLA_CHUNK` is described as under review,
  default off, and a mitigation rather than root-cause closure.

Also retained from the original PR: the capture-size comment fix (the engine
clamps `max_cudagraph_capture_size` to 24, so the documented 36 at k=5/seqs=6
was never in effect) and the note that concurrent heavy prefills remain
scheduler-bound (#80/#140).

The CHANGELOG entry is retitled and redated to match the new content.

Happy to re-run any cell — with repeats this time, since n=1 is what produced
the retracted claim.


ashhart:
Amazing work Seth!!

plotarmordev:
Maintainer review at exact head `11b9e7e8d3270a8cb7595f503509fbcb1cbc5ab8`: **HOLD pending a #141-informed rewrite.**

The later repeated-burst evidence in #141 invalidates the document's safety framing: the N=32/c=32 cell failed on burst 4, the proposed N=16/96-row clean bracket failed on another configuration, and the prior #143 admission-cap mechanism was explicitly retracted. These are stochastic/lifetime-sensitive failures, not a verified N=16-bad/N=32-safe boundary.

Please remove the known-false safety recommendation and document the evidence conservatively: configuration changes may alter incidence, but no tested `MAX_NUM_SEQS` value is established as generally safe. This should not merge in its current form.

plotarmordev (changes requested):
Re-review at 839e476: the requested #141-informed rewrite is substantively correct and removes the false N=32-safe recommendation. Two current-main integration blockers remain. (1) The PR now conflicts in CHANGELOG.md; rebase current main and preserve the newer 2026-08-28 Issue #154 entry plus the 2026-08-27 Added/Fixed entries, then add this docs change without replacing them. (2) Refresh two stale status claims introduced by the old base: .env.dspark.example still calls N=4 the shipped default, while current main ships MAX_NUM_SEQS=6; and both the env block and changelog call PR #151 'under review/proposed', but #151 merged as a5403b77. Describe N=4 as the measured comparison baseline and #151 as merged, default-off, and still a mitigation rather than root-cause closure. This is docs-only, so no live gate is needed after those corrections; exact-current-main CI is sufficient.

plotarmordev (approved):
Maintainer integration at 828b963 resolves the current-main changelog conflict, preserves the Issue #154 safe prefill default and all newer entries, updates the N=4 result to a measurement baseline rather than the shipped default, and records PR #151 as merged/default-off. The requested #141 rewrite now states the evidence conservatively and retracts the unsupported safety claim. Exact-head local and hosted scripts/ci-validate.sh both pass. Approved.

PR #147, opened 2026-08-26
fix: ship pi example with thinking_token_budget capability off

hirocaster:
## What

`pi-models.dspark.example.json` shipped `compat.supportsThinkingTokenBudget: true` while the server default after #66 is `DSPARK_ENABLE_ISSUE31_GPU_HOTFIX=0`. On the default server, vLLM's stock V2 runner rejects every request carrying `thinking_token_budget` with `thinking_token_budget is not yet supported by the V2 model runner`, and the error's own advice (`VLLM_USE_V2_MODEL_RUNNER=0`) points at a knob this recipe does not expose.

A fresh clone that followed the README on both sides — `.env.dspark.example` for the server, `pi-models.dspark.example.json` → `~/.pi/agent/models.json` for the client — therefore failed **every** pi request in which thinking was enabled.

This PR:
- ships the example with `supportsThinkingTokenBudget: false` to match the server default (`DSPARK_ENABLE_ISSUE31_GPU_HOTFIX=0`),
- rewrites the README's pi section to describe the mechanism pi actually implements: the boot flag **and** the client capability are both required, and pi sizes the budget from per-level defaults overridable through the pi `thinkingBudgets` setting — there is no per-model `thinkingTokenBudget` field (the previous docs referenced a mechanism pi does not implement, and `omitWhenUnset`/`$var: "model.thinkingTokenBudget"` are not pi keys),
- documents the failure mode and the reverse direction,
- adds a CHANGELOG entry.

## Verification

- Example JSON stays valid; `supportsThinkingTokenBudget` ships `false`.
- Statements cross-checked against pi internals (`@earendil-works/pi-ai`):
  - pi attaches a top-level `thinking_token_budget` when `supportsThinkingTokenBudget` is true and thinking is enabled,
  - default budgets: `minimal` 1024 / `low` 2048 / `medium` 8192 / `high` 16384, clamped so the answer always keeps room; overridable via the `thinkingBudgets` setting,
  - `omitWhenUnset` and `$var: "model.thinkingTokenBudget"` do not exist in pi.
- The quoted error string matches `patches/hotfix-dsv4-issue31-v2-thinking-budget-gpu.py`.

## Not verified live

The "on" direction (boot flag `1` + capability `true`) was checked against the pi code paths cited above but not re-benchmarked on live hardware in this PR. The README states the budgets pi will use by default; if you run with the capability on, please report back so the docs can be confirmed end-to-end.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+## 2026-08-27
+
+### Fixed
+
+- **`pi-models.dspark.example.json` no longer advertises `thinking_token_budget` against a server that rejects it**: issue #66 made the GPU budget hotfix opt-in (`DSPARK_ENABLE_ISSUE31_GPU_HOTFIX`, default `0`) on the premise that fresh clones omit the field, but the shipped pi model entry kept `compat.supportsThinkingTokenBudget: true`. A clone that followed the README on both sides — `.env.dspark.example` for the server, `pi-models.dspark.example.json` into `~/.pi/agent/models.json` for the client — therefore failed every pi request with `thinking_token_budget is not yet supported by the V2 model runner`, and the error's own advice (`VLLM_USE_V2_MODEL_RUNNER=0`) points at a knob this recipe does not expose. The example now ships `false`, matching the server default; the README's pi section states that the budget needs the boot flag *and* the client capability, quotes the error produced by an out-of-step pair, and documents the reverse direction (turn the capability off, keep `DEFAULT_THINKING` and the `reasoning_effort` mapping, leave `maxTokens` generous enough for reasoning to finish).
+
 ## 2026-08-25
 
 ### Added
--- a/README.md
+++ b/README.md
@@ -347,34 +347,34 @@ curl :8888/v1/chat/completions -H 'Content-Type: application/json' -d '{
 }'
 ```
 
-**pi** — set a reasoning budget in the model entry, and pi sends the field
-for you on every request where you enable thinking. Copy
-[`pi-models.dspark.example.json`](pi-models.dspark.example.json) to
-`~/.pi/agent/models.json`, then add a `thinkingTokenBudget` (in tokens) to the
-`deepseek-v4-flash-0731` model:
+**pi** — the budget needs the boot flag **and** the pi model entry, so
+[`pi-models.dspark.example.json`](pi-models.dspark.example.json) ships
+`supportsThinkingTokenBudget: false` to match the server default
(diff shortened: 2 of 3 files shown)


plotarmordev (approved):
Exact head 42d5bb7 validated locally: JSON and cross-file default/hotfix gating are consistent, and the docs/config-only cutover matches the server default. Hosted validate is green at the same head. Approved for owner merge; not merged here.

plotarmordev:
## Exact-head problem proof and regression evidence

Validated head: `42d5bb7787e570b0269915376b6f27b3ac5e7f41`.

**Reported/default-path problem:** the shipped server example sets `DSPARK_ENABLE_ISSUE31_GPU_HOTFIX=0`, where pinned stock V2 rejects `thinking_token_budget`, while the shipped pi model example advertised `supportsThinkingTokenBudget: true`. Pi's documented/source behavior maps that capability to the top-level vLLM `thinking_token_budget` field whenever thinking is enabled. A fresh clone following both examples therefore enabled a client field that its default server explicitly rejects.

**Fix proof:** the exact-head JSON parses, the `deepseek-v4-flash-0731` entry now asserts `supportsThinkingTokenBudget == false`, and the server example still asserts `DSPARK_ENABLE_ISSUE31_GPU_HOTFIX=0`. README instructions accurately describe the two-part opt-in and pi's `thinkingBudgets`; the documented built-in values and 1,024-token answer-room clamp match current pi source. No runtime/Compose/launcher code changes are present.

**Regression evidence:** complete repository CPU recipe CI passed at this exact head, including 11 GPU-thinking-budget tests and every existing shell/Python/recipe gate; hosted `validate` is also green. The enabled direction remains documented rather than re-benchmarked here, which is appropriate because this PR only makes the shipped client default match the shipped server default.

Conclusion: docs/config-only correction is approved for normal merge at this exact head.

plotarmordev:
## Post-merge gate closure

Retrospective checks for the already merged exact range `70a7cc4...42d5bb7` are complete:

- commit-range secret scan found no private-key, GitHub-token, AWS-key, or `sk-` credential patterns;
- independent CYBERGLM review: **PASS**, no blockers, explicit **NO REVERT** recommendation;
- CYBERGLM independently confirmed pi 0.84.3 request semantics, budget defaults, answer-room clamp, real schema keys, pinned vLLM error text, and docs/config-only scope;
- merge commit `346a898115e0ba58a91cb1a746bcae993ce8c481` is byte-equivalent to the approved PR diff;
- post-merge `main` validation passed: https://github.com/MiaAI-Lab/DeepSeek-v4-Flash-DSpark-2x-DGX-Spark/actions/runs/33132283112

No revert is indicated. Minor wording boundary: “every pi request fails” means every thinking-enabled pi request.

PR #148, opened 2026-08-27
Experimental opt-in: LMCache KV offload (107K-token reload in ~1.9s vs ~65s re-prefill)

sethforprivacy:
## What

An **opt-in, off-by-default** `lmcache/` option directory wiring [LMCache](https://github.com/LMCache/LMCache) (MP mode) into this recipe: per-node `lmcache server` processes (CPU L1 + NVMe filesystem L2) hold KV across engine restarts, connected via `LMCacheMPConnector`. Nothing in the stock path changes — the generated compose overlay boots identically with `DSPARK_ENABLE_LMCACHE` unset.

## Why

These pairs ship with 4TB NVMe that the recipe barely touches. With this enabled, a ~107K-token context that costs **~65s to re-prefill reloads in ~1.83–1.92s** (measured on a GB10 pair, TP=2, `nvfp4_ds_mla` KV, this recipe's pinned image + `pip install lmcache==0.5.4` and friends in a derived layer). Same numbers across repeated trials; warm-hit serving path verified end-to-end (boot → store → warm hits → engine restart → reload).

## What's in the directory

- `README.md` — full guide: derived-image deps (including why `cupy-cuda13x` is load-bearing and how to source-build `cuda_ops` for sm_121a), non-negotiable config, lifecycle rules, verification steps
- `run-lmcache-server.sh` — per-node server launcher (`--restart no` is deliberate — see below)
- `patch-compose-lmcache.py` — generates `docker-compose.lmcache.yml` from `docker-compose.dspark.yml`; injects the connector config gated on `DSPARK_ENABLE_LMCACHE=1`, pins `PYTHONHASHSEED=0`, clears `PYTORCH_CUDA_ALLOC_CONF` (vLLM rejects KV connectors with `expandable_segments:True`). Anchor-asserted so it fails loudly if the compose layout drifts. Validated against current `main`'s compose with `docker compose config`.

## Why "experimental"

The happy path is solid; the *failure* paths upstream are still being hardened, and the README is blunt about them:

- Restarting an `lmcache server` loses its GPU contexts, and current upstream then parks every lookup-hit request forever (scheduler heartbeat never starts + lookup errors don't propagate). I filed the field report as LMCache#4759 and the heartbeat fix as LMCache#4764; LMCache#4754 adds a defensive event-IPC selector for driver 580/CUDA 13 platforms.
- Mitigation until those land (documented as hard rules): treat pair+servers as one lifecycle unit, run servers with `--restart no` so death is visible, restart the whole pair on any `No GPU context found`.
- `PYTHONHASHSEED=0` is mandatory on every process — chunk keys use Python's randomized `hash()`, so an unpinned restart silently invalidates the entire cache.
- On 128GB unified-memory nodes the server can be OOM-pressured during engine weight load; monitor it.

## Testing

- Overlay generation validated against current `main` compose (`docker compose config` passes; stock boot unchanged with the flag off)
- Serving path: repeated store/reload trials on a production-config 2× pair — 3/3 instrumented trials + 2 untraced trials, reload TTFT 1.83–1.92s for the 107K-token context
- Failure modes exercised deliberately (server restart mid-lifecycle, engine restart against live servers) — that's where the lifecycle rules and upstream issues above come from

Happy to iterate on placement/naming if you'd rather this live elsewhere (e.g. `docs/` as a guide rather than an option directory).

### DIFF
--- a/lmcache/README.md
+++ b/lmcache/README.md
@@ -51,7 +51,9 @@ crash and does not fall back to re-prefill, so an operator watching only for
 container exits sees a healthy pair serving nothing. The only reliable signal
 is the server's own exit plus `No GPU context found` in its log — which is why
 the launcher uses `--restart no` (a dead server must stay dead and visible)
-and refuses to replace a running server unless `LMCACHE_FORCE_REPLACE=1`.
+and will not recreate the server at all while a `vllm-dspark` model container
+is running: the fresh-server-under-live-engine case IS this wedge, so the only
+supported recovery is a full-pair restart (servers first, then the engine).
 Once LMCache PR #4764 lands the parks become graceful degradation; until then,
 **the blast radius of a dead cache server is the model service.**
 
@@ -73,7 +75,8 @@ the upstream heartbeat fix lands.
   and `unset PYTORCH_CUDA_ALLOC_CONF` (vLLM rejects KV connectors alongside
   `expandable_segments:True`). The stock `PYTORCH_CUDA_ALLOC_CONF` service
   env entry is left untouched and no `PYTHONHASHSEED` env entry is added, so
-  with the flag off the engine process gets stock env and stock argv.
+  with the flag off the engine keeps stock allocator/hash behaviour and stock
+  argv; the config still differs from stock by the two inert deltas below.
 - Launch with
   `COMPOSE_FILE=$PWD/docker-compose.lmcache.yml ./start-deepseek-v4-flash-dspark.sh`.
 
@@ -146,8 +149,13 @@ COMPOSE_FILE=$PWD/docker-compose.lmcache.yml ./start-deepseek-v4-flash-dspark.sh
 
 Knobs on `run-lmcache-server.sh`: `LMCACHE_DISK_DIR`, `LMCACHE_L1_GB`,
 `LMCACHE_PORT`, `LMCACHE_OOM_SCORE_ADJ` (default `0`), and
-`LMCACHE_FORCE_REPLACE=1` to override the guard that refuses to replace a
-server that is currently running.
+`LMCACHE_FORCE_REPLACE=1` to override the guard against re-creating a
+currently-running server. It applies only when the pair is actually down:
+while any model container is live, server re-creation is refused
+unconditionally (see Operational risk 3). "Model container" means either a
+compose container labelled `com.docker.compose.service=vllm-dspark` or any
+container whose name contains `vllm-dspark` (deliberately over-eager —
+refusing too hard is the safe side of the wedge).
 
 With `DSPARK_ENABLE_LMCACHE` unset, `0`, or anything other than exactly `1`,
 the generated compose boots the stock recipe unchanged — every LMCache change
@@ -161,9 +169,12 @@ diff <(docker compose -f docker-compose.dspark.yml config) \
 
 with the flag unset, the only deltas are inert: an added
 `DSPARK_ENABLE_LMCACHE: "0"` pass-through, the `KVT_ARGS=""` gate, and an
-empty `$${KVT_ARGS}` expansion in the serve argv. `PYTORCH_CUDA_ALLOC_CONF`
-still renders as `expandable_segments:True` and `PYTHONHASHSEED` is absent,
-exactly as on stock.
+empty `$${KVT_ARGS}` expansion in the serve argv (an empty unquoted expansion
+adds no argument, so the exec'd argv is unchanged). The env is therefore not
+literally byte-identical to stock — the inert `DSPARK_ENABLE_LMCACHE=0`
+variable is added — but no stock behaviour depends on it.
+`PYTORCH_CUDA_ALLOC_CONF` still renders as `expandable_segments:True` and
+`PYTHONHASHSEED` is absent from the service env, exactly as on stock.
 
 To verify it's working: the head engine log shows `LMCacheMPConnector`
 at startup, and a repeated long-context request logs a lookup hit with TTFT
--- a/lmcache/README.md
+++ b/lmcache/README.md
@@ -7,18 +7,75 @@ GB10 pair, `nvfp4_ds_mla`). KV is held by per-node `lmcache server` processes
 
 **Status: experimental.** Serving-path verified (boot, store, warm hits,
 reload) across repeated trials; the *failure paths* upstream are still being
-hardened — see Known issues. Do not enable on a pair you cannot restart.
+hardened — see Operational risk and Known issues. Do not enable on a pair you
+cannot restart.
+
+## Operational risk — owner decision required
+
+Enabling this is an explicit operational trade-off, not a free win. The
+failure modes below are known and **not** fully mitigated in this option
+directory; an owner has to accept them before turning the flag on.
+
+**1. The cache servers live outside Compose.** `run-lmcache-server.sh` starts
+one plain `docker run` container per node. Compose does not know about it, so
+it is not started, stopped, restarted, health-checked or supervised with the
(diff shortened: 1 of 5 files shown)


sethforprivacy:
Small evidence update: the "server can exit under engine-boot memory pressure" caveat in the README is now confirmed root-caused — `dmesg` on the head node shows a kernel global OOM kill of the `lmcache` server process during an engine cold boot (unified-memory weight-load spike; details posted to LMCache#4759). The documented mitigation stands unchanged: `--restart no` + alert on server exit, full-pair restart after any server death. A negative `oom_score_adj` on the server container is a reasonable extra belt on 128 GB boards.

plotarmordev:
Maintainer review at exact head `b98de48b42c8b804de1b272499d086ddae3c122f`: **HOLD; the opt-in boundary is not yet clean.**

Blocking items:
- with the LMCache flag unset, rendered runtime behavior still changes through unconditional `PYTORCH_CUDA_ALLOC_CONF` and `PYTHONHASHSEED` injections, contradicting the claim that stock boot is identical;
- the README names `start-deepseek-v4-flash.sh`, but this repository's launcher is `start-deepseek-v4-flash-dspark.sh`;
- server lifecycle remains outside Compose, and the confirmed unified-memory cold-boot OOM plus unresolved LMCache heartbeat/failure modes can leave the model service without the cache server.

Please gate every LMCache-specific environment change, fix the launcher command, and either resolve the failure-path dependencies or make the operational risk an explicit owner decision. CI being green does not close these runtime blockers.

plotarmordev (changes requested):
Re-review at exact head `db1300690bd13f8f9320a9967df384be346295b8`: the three requested changes are present. The exact-`1` branch now contains the connector argument, `PYTHONHASHSEED=0`, and allocator unset; the launcher name is fixed; and the known OOM/heartbeat/server-lifecycle risk is an explicit owner decision. The generator also applies cleanly to current `main`, its Python and shell compile checks pass, and the generated Compose config renders.

**HOLD; three remaining safety/verification blockers:**

1. `patch-compose-lmcache.py` accepts shell substitutions in `server-urls`, then embeds the value inside a double-quoted shell assignment. Reproducer accepted `tcp://192.0.2.10:6667$(touch${IFS}/tmp/pr148-injection)`. Do not execute that output. Strictly parse a comma-separated `tcp://host:port` grammar, reject control characters, `$`, backticks, and shell punctuation, and validate numeric port range.
2. The new generator and lifecycle script have no committed behavioral tests and are not registered in `scripts/ci-validate.sh`; the hosted green check therefore does not exercise them. Add hermetic tests covering unset/`0`/`true`/`2` versus exact `1`, source-anchor drift refusal, malicious URL rejection, and generated-Compose rendering. The disabled cases must preserve stock allocator/hash behavior; exact `1` must add all three LMCache changes together.
3. `run-lmcache-server.sh` refuses replacement only while `lmcache-server` itself is running. In the documented failure mode the server is dead while the model engine remains live/wedged, so the script proceeds to recreate an empty server even though the documented recovery is a full-pair restart. Refuse to remove/recreate the server while the model container is running unless an explicit recovery override proves the pair is already down.

Also correct the literal claim that the engine environment is byte-identical when disabled: the inert `DSPARK_ENABLE_LMCACHE=0` variable is added. The behavioral stock-path claim is supported; byte identity is not.

LMCache PRs #4764 and #4754 are still open. That is not a new blocker because the README now makes the risk an explicit opt-in owner decision, as requested. With the three items above fixed, this remains a useful P2 experimental option worth reconsidering; it is not merge-ready at this head.

sethforprivacy:
Re-review follow-up at head `de51c16e` — all three blockers plus the wording correction are addressed.

**1. `server-urls` injection (`patch-compose-lmcache.py`).** The value is now produced by a strict whitelist parser instead of the quote/space denylist: comma-split, per-entry `fullmatch` of `tcp://<IPv4 dotted-quad or RFC1123 hostname>:<port>` — port `[1-9][0-9]{0,4}` with a `<= 65535` check and leading zeros refused; digits+dots hosts that are not valid dotted quads (`999.999.999.999`, `01.2.3.4`) are refused rather than deferred to connect time; control characters and embedded newlines cannot pass `fullmatch`; host length capped. Anything outside the grammar exits non-zero before anything is written. The exact reproducer `tcp://192.0.2.10:6667$(touch${IFS}/tmp/pr148-injection)` now exits 1, writes no overlay, and no marker is created. IPv6 literals are deliberately refused and documented (the fabric preflight in the launcher is IPv4-only).

**2. Committed behavioral tests + CI registration.** New `scripts/test-lmcache-compose-gate.sh` (registered in `scripts/ci-validate.sh`: shell-syntax list, `py_compile` of the generator, and the functional step). It extracts the *shipped* gate branch out of the *generated* compose (not a copy) and runs it under bash for unset/`0`/`true`/`2`/`FALSE`/`yes` vs exactly `1`: disabled cases preserve stock allocator (`expandable_segments:True`), absent `PYTHONHASHSEED`, and empty `KVT_ARGS`; exactly `1` flips all three changes atomically and emits a valid JSON `--kv-transfer-config` carrying the urls verbatim. Also covered: docker-free post-generation invariants (stock allocator entry intact, no service-env `PYTHONHASHSEED`, single gate + one pass-through), anchor-drift refusal on all four anchors, 33 malicious/malformed url inputs, and a rendered-config layer (skips where `docker compose` is absent) asserting the flag-off render is a pure word-addition over stock confined to the gate/pass-through tokens, plus lifecycle-guard tests for item 3 under a stateful docker stub.

**3. Server lifecycle vs live engine.** `run-lmcache-server.sh` now refuses — with no `LMCACHE_FORCE_REPLACE` bypass — to remove/recreate the server while any model container is running (compose-service-label check plus a deliberately over-eager name-substring check for label-less hand-run containers). `LMCACHE_FORCE_REPLACE` only governs the stale-server-pair-down case, matching its documented contract. The three guard `docker ps` probes fail **closed**: a docker failure is now a refusal, not a false "nothing is running". Section 5 of the new test exercises the real launcher under a docker stub: refusal under a live model with both `force=0/1`, live-server refusal without the override (server not removed), forced re-create allowed with the pair down, clean-boot path proceeds, and fail-closed on a broken `docker ps`.

**Wording.** The byte-identical claim is gone from the generator docstring and the README: the rendered config always carries the inert `DSPARK_ENABLE_LMCACHE: "0"` pass-through and an empty `$${KVT_ARGS}` expansion (no argv delta); stock *behaviour* is what the flag-off path preserves, and the docs now say exactly that.

Panel note: the fix set was additionally reviewed by a 4-model independent panel (unanimous ship verdict; 8 follow-up items — fail-closed guards, guard-coverage tests, fake-IPv4 rejection, docker-free invariant checks, reporting bug in the test itself — all fixed and confirmed by the reporting reviewers). Local verification: new gate 16/16 green incl. the rendered-config layer; full `ci-validate.sh` green except three pre-existing macOS-environment failures (bash 3.2 / BSD stat) that reproduce identically on the untouched base.

plotarmordev (approved):
Re-review at de51c16 against current main 2a4dd0b: the three blockers are closed. The URL parser rejects the command-substitution reproducer and malformed endpoints before writing; committed tests cover exact-1 gating, stock-path preservation, anchor drift, rendered Compose, and fail-closed lifecycle guards; server replacement is unconditionally refused while a model container runs. Local exact-current-main merge completed cleanly and full scripts/ci-validate.sh passed, including test-lmcache-compose-gate (16/16). The PR adds only the opt-in lmcache directory plus CI/test registration; no shipped runtime file changes. Approved as experimental/default-off with the documented upstream OOM/heartbeat risks.

Issue #136, opened 2026-08-25
MTP + xgrammar termination can livelock sample_tokens on Anemll 0.1.1 (vLLM #52805)


data-angel:
## Summary

The Anemll `0.1.1` lane can hit the vLLM MTP + xgrammar termination livelock fixed upstream by vLLM PR #52805:

- https://github.com/vllm-project/vllm/pull/52805
- merge commit: `12f64b39d29282437e35be9aa5db432fb2a1a6e6`

This is distinct from the mid-serve JIT timeout in #65 and the collective failure in #87. Increasing `VLLM_EXECUTE_MODEL_TIMEOUT_SECONDS` postpones engine death but does not resolve this livelock.

## Affected public configuration

- 2x DGX Spark / GB10, TP=2
- Anemll image `ghcr.io/anemll/dspark-vllm-gx10:0.1.1`
- vLLM `0.25.2.dev0+g752a3a504.d20260714`
- DeepSeek V4 Flash 0731 with DSpark speculative decoding
- Async scheduling and `structural_tag` / xgrammar structured output

## Incident

A standard structured-output request entered the failure state. Immediately before progress stopped, xgrammar logged that its matcher had terminated after the stop token and then attempted to accept additional speculative tokens. Generation throughput dropped to zero while the request remained running. Both ranks subsequently logged `No available shared memory broadcast block found in 60 seconds` once per minute.

After exactly 1,800 seconds:

```text
TimeoutError: RPC call to sample_tokens timed out
EngineDeadError: EngineCore encountered an issue
```

There was no JIT activity during the stall, OOM, or preceding NCCL/RoCE error. This did not match the concurrent long-prefill pressure case.

## Root cause and fix

The symptoms match vLLM #52805: after the xgrammar matcher terminates, the MTP path can continue advancing speculative tokens and livelock sampling.

We applied a compatibility-gated backport of #52805 to the pinned image. Repeated concurrent structured tool-call canaries then passed with no further matcher-after-termination warning, `sample_tokens` timeout, broadcast stall, or engine death.

## Suggested recipe action

If the next runtime image updates vLLM past merge commit `12f64b39d29282437e35be9aa5db432fb2a1a6e6`, no recipe patch should be necessary. If the Anemll `0.1.1` image remains supported, a small fail-closed backport would protect that pinned lane in the interim.

I can provide the idempotent backport and its fixture test if useful.

[...]

PR #151, opened 2026-08-27
fix(runtime): bound oversized sparse-MLA decode batches

### LINKED ISSUE #141: [Bug] TP=2 pair deadlocks in sparse-MLA decode once the DSpark verify batch exceeds flashinfer's _DECODE_MAX_TOKENS (64 rows)

plotarmordev:
## Summary

Mitigates #141 on the pinned Anemll vLLM/FlashInfer lane.

The sparse-MLA decode wrapper has a fixed 64-row FlashInfer limit. This change adds a default-off, source-locked hotpatch that keeps calls at or below 64 rows on the original path and chunks only oversized calls into fixed 64-row slices.

## Safety boundary

Only row-coupled tensors are sliced: query, sparse indices, output, SWA top-k lengths, and optional extra sparse indices/lengths. KV caches, workspace, scale, and per-head sinks remain shared. The patch is atomic, idempotent, exact-source locked, and fail-closed before engine execution on either rank.

Enable explicitly with `DSPARK_ENABLE_ISSUE141_SPARSE_MLA_CHUNK=1`. Default behavior remains unchanged.

## Validation

- 14 focused CPU/source-lock tests passed;
- full CPU recipe CI passed with CUDA hidden;
- Compose rendering passed;
- status and application passed against `ghcr.io/anemll/dspark-vllm-gx10:0.1.1`;
- live TP=2 recreation/environment proof passed;
- both ranks reported the patch applied and verified;
- concurrency-16 live generation completed with 128 output tokens per request;
- post-run smoke passed with zero restarts and zero OOMs;
- the original deployment was restored and verified afterward.

## Limitation

This is an opt-in mitigation, not a claim that the stochastic root cause is closed. The live campaign did not include the repeated long-duration burst/soak required to close #141, so the issue should remain open while this PR is reviewed.


plotarmordev:
Independent Fable exact-head review of `890d9de934327b34ef9a65ac2101c4b35a0f9586`: **PASS on runtime correctness/security as an opt-in mitigation; no code blocker found.**

Regression evidence at this exact head:
- 14 focused source-lock/chunking tests passed;
- full CPU recipe CI passed with CUDA hidden;
- hosted `validate` passed;
- Compose rendering and pinned-image apply/status passed;
- the <=64-row path remains the original single call;
- live TP=2 recreation/environment proof passed;
- both ranks reported the patch applied and verified;
- concurrency-16 generation completed with 128 output tokens per request;
- smoke passed with zero restarts/OOMs and the baseline deployment was restored.

The 1,623 added lines decompose to 674 production, 808 tests, 113 docs, 18 CI, and 10 config lines. The actual runtime behavioral delta is about 45 lines; most volume is exact-source/transactional armor and tests. Fable identified roughly 250-320 lines that could be reduced without changing safety.

One documentation correction is required before final merge recommendation: `docs/PATCHES.md` says no live cluster was touched, which contradicts the completed TP=2 validation recorded in this PR. Correct that sentence, then rerun focused/full CI at the new exact head. This PR remains a mitigation and must not close #141 as root-caused.

This is an evidence record, not a self-approval: the PR author cannot provide an independent GitHub approval. An owner or another reviewer should approve after the documentation correction; no merge was performed.

sethforprivacy:
Validated this PR at exact head `890d9de` on a second failing pair — full results posted in https://github.com/MiaAI-Lab/DeepSeek-v4-Flash-DSpark-2x-DGX-Spark/issues/141 (2× DGX Spark TP=2, recipe `70a7cc4`, `nvfp4_ds_mla`, k=5 and k=6). Summary:

- Both of our previously-fatal cells ran clean with the flag on: N=16/k=5 (96 verify rows, 393,216 tok, died at ~237K unpatched) and N=32/k=6/c=32 (224 rows, 5/5 repeated bursts, died on burst 4 unpatched). Zero silent truncation (per-request `finish_reason` checked), engine alive after every burst.
- The opt-in flow worked as documented end to end: default-off boot is stock (no patch lines), exact-1 enables it, both ranks logged `applied and verified` on every boot, the launcher synced the patch to the worker, and rollback via the paired stop/start restored stock bytes on both ranks (verified zero `issue141` log lines and a clean generation afterward).
- Observed cost on our pair at the 224-row cell: 249–283 tok/s aggregate vs 325–337 unchunked in completed bursts (~15–20%, un-retuned).

This is n=1/n=5 against a stochastic failure, so it does not close #141 — but the mitigation now has clean results on both known-failing pairs and the operational boundary behaved exactly as claimed. Works for us as the interim; we are adopting it (opt-in) on our production pairs.

plotarmordev:
## Cleaned exact-head validation — `4cc616119cee1622ece4c886ae4905015b2b8890`

The Fable cleanup reduced this branch from `1623+/4-` to `1351+/4-` versus `origin/main` (272 lines removed, 16.8%) while retaining the whole-method source lock, five load-bearing FlashInfer guards, transactional publication/rollback, row-boundary behavior, and startup wiring. CYBERGLM independently reviewed this exact head: **PASS**, no blockers; the remaining estimated redundancy is cosmetic only.

### Exact-head local validation

- `python3 scripts/test-issue141-sparse-mla-decode-chunk.py`: **14/14 passed**.
- `python3 scripts/test-python-hotfix-failclosed.py`: **11/11 passed**.
- `CUDA_VISIBLE_DEVICES= bash scripts/ci-validate.sh`: **passed** (complete repository CPU recipe gate).
- `docker compose -f docker-compose.dspark.yml config -q`: **passed**.
- Pinned image `ghcr.io/anemll/dspark-vllm-gx10:0.1.1`: `--status` reported **NOT APPLIED (compatible)**; an ephemeral apply reported **applied and verified**.
- Secret scan: no private-key, GitHub-token, AWS-key, or `sk-` credential patterns in the worktree or `origin/main...HEAD` commit range.

### Nightshift validation

- Exact branch diff routing selected heavy gates **26+52+55+66** and `manual-required` (10 paths: 6 static, 2 serving, 1 manual, 1 unknown).
- Full Nightshift regression suite passed: tier routing/shared-boot/recreation/warmup; 19 prefill-sentinel cases; degradation, chain-verdict, chain-queue, AA-skip, boot-stats, and probe-timeout gates; 8 prefix32k cases; 10 deadstate A/B cases; 2 truncation cases; all pinned probe checksums.
- Exact-head TP=2 live campaign (`4cc6161`): recreation gate **PASS**; patch status **APPLIED + VERIFIED on both ranks**; smoke `hang=0 fail=0`; concurrency 16 produced 128 output tokens for every request (`median_decode=23.6 tok/s`, aggregate `236.8 tok/s`, TTFT `2800 ms`); both containers remained running with `RestartCount=0` and `OOMKilled=false`; restore completed `rc=0`, followed by clean smoke/verify.

### Hosted validation

- GitHub `validate` completed **SUCCESS** on this exact head: https://github.com/MiaAI-Lab/DeepSeek-v4-Flash-DSpark-2x-DGX-Spark/actions/runs/33120626532/job/98686310266
- GitHub currently reports the PR mergeable.

### Evidence boundary

This is a bounded exact-head validation, not a blanket no-regression claim. It covers one short concurrency-16 TP=2 burst. Repeated stochastic soaks, disposable SM121a numerical/CUDA-graph testing, the full OFF/ON/drift boot comparison, and the second failing pair's 64/65 comparison remain outstanding.

plotarmordev:
## Reported-problem proof and bounded regression evidence

Exact head: `4cc616119cee1622ece4c886ae4905015b2b8890`.

**Proof for reported issue #141:** the incident was sampled in FlashInfer's sparse-MLA paged fallback, and a reporting pair showed the relevant 64/65-row distinction. A second failing pair reported its previously fatal 96-row and 224-row cells completing with the opt-in chunking enabled, including five repeated bursts for the latter. Our cleaned exact-head TP=2 campaign then recreated both containers, confirmed `APPLIED + VERIFIED` on both ranks, and completed concurrency-16 generation with 128 output tokens for every request.

**Regression boundary exercised:** the focused 14-case suite proves the ≤64 path remains one unchanged call; >64 rows are covered in ascending ≤64 slices while preserving all six row-coupled tensors, shared caches/workspace/sinks/scale, optional arguments, source locks, idempotence, and transactional rollback. The exact head also passed 11 shared fail-closed tests, complete CPU recipe CI, Compose rendering, pinned-image status/apply, full Nightshift regressions and probe checksums, and hosted `validate`. Both ranks stayed running with zero restarts/OOMs; restoration and post-restore smoke passed.

Fable cleanup and CYBERGLM exact-head review both passed with no blockers. This remains a bounded mitigation, not proof of root cause or a blanket no-regression claim: the long stochastic soak, disposable SM121a numerical/CUDA-graph campaign, full OFF/ON/drift comparison, and second pair's direct 64/65 comparison remain open.

plotarmordev:
## Current-main integration gate — `7fc042f231b6362bf2edd4b3cf36ed208ece18d7`

Merged current `main@2bf690c` normally into the feature branch; no rebase or history rewrite. Conflict resolutions retained both merged #147/#153 content and #151 content exactly once across changelog, Compose, env/patch docs, CI registration, and launcher preflights.

- Issue #141 patcher and focused suite are byte-identical to previously approved/live-tested head `4cc6161`.
- Every issue #136 runtime/test artifact is byte-identical to current main; the two patches target disjoint vLLM files.
- Prior separate live evidence carries. The combined exact-1 startup wiring is defended by the merged fail-closed suite; dual-enable was not live-tested and remains a bounded caveat.
- Exact integration-head validation: 14/14 issue141, 27/27 issue136, 14/14 shared fail-closed, complete CPU recipe CI, and Compose rendering passed.
- Exact diff still routes through Nightshift gates 26+52+55+66 plus manual review.
- Worktree and `origin/main...HEAD` secret scans found no credential patterns.
- CYBERGLM exact integration review: **PASS**, no blockers; normal merge commit recommended.
- Hosted `validate`: **SUCCESS** — https://github.com/MiaAI-Lab/DeepSeek-v4-Flash-DSpark-2x-DGX-Spark/actions/runs/33134881571

Ready for normal merge at this exact head. Existing #141 stochastic-evidence limitations remain unchanged.

plotarmordev:
Merged normally at exact reviewed head `7fc042f231b6362bf2edd4b3cf36ed208ece18d7` as merge commit `a5403b77daaa819b5c74795fc6f3bdfc7b29b36b`. Post-merge `main` validation completed **SUCCESS**: https://github.com/MiaAI-Lab/DeepSeek-v4-Flash-DSpark-2x-DGX-Spark/actions/runs/33134934905

PR #152, opened 2026-08-27
Add bounded loop-breaker hotfix for the #82 announce-loop (detokenizer repeated-line detector)

jdecker76:
## Bounded loop-breaker for the #82 announce-loop — default OFF

Companion to the original [issue #82 report](https://github.com/MiaAI-Lab/DeepSeek-v4-Flash-DSpark-2x-DGX-Spark/issues/82#issuecomment-5445734497). This is a mitigation, not a root-cause fix or closure of #82.

### Current contract
Only exact `DSPARK_LOOP_BREAKER=1` applies and arms the detokenizer repeated-line detector. Every other value leaves a fresh detokenizer untouched and parses no numeric knob; a previously patched module remains inert when disabled at import. `DSPARK_SKIP_LOOP_BREAKER_HOTFIX=1` skips application/preflight, but `--status` always inspects bytes.

The detector returns `loop-breaker` through the stop-string path when eligible normalized lines reach bounded thresholds: prose repeats default 6, short sentence repeats default 15, minimum output tokens default 64. Repeat bounds are 2–1024 and minimum-token bounds 0–1000000. Fenced/indented code, structural lines and DSML-marked lines are excluded; the last-400-character DSML tail heuristic holds fire. This is not general tool-block parsing or calibrated false-positive proof.

The patcher requires complete unique compiling injected blocks for applied status, uses atomic replacement with mode preservation, and attempts restoration after failed postwrite verification including read/decode errors. Restore failure is explicit and nonzero. Both numeric validators bound normalized decimal magnitude before conversion.

The launcher forwards all five resolved controls and canonical worker patch paths, syncs the selected override bytes to both workers, and rejects an enabled/unskipped missing, non-file or symlinked selected patcher before remote operations.

### Latest published source and CPU follow-up
Current published head: `cdcbda8167b4ea3738011618dc8b4b00f601ab96`. Test-only cleanup removed incidental environment/path/wording assertions while retaining selected-source consumption, fail-closed behavior and rank outcomes. Main reran the final candidate: 43 loop-breaker tests and 32 hotfix-chain cases passed.

[Hosted validate 34978322676](https://github.com/MiaAI-Lab/DeepSeek-v4-Flash-DSpark-2x-DGX-Spark/actions/runs/34978322676) succeeded with all twelve late guards individually verified. It checked out `bcc2302160aabc9f4e79f06d8be75e49691a39af`. The checked tree exactly matches this published head. This does not qualify the pinned image, real Compose/ranks or live loop quality; the HOLD and unresolved #82 status remain.

### Earlier production-source review CPU evidence
Earlier review head: `2b5a4f394ee77e92c921d85d171e99f274433a96`; normal integration anchor: `e369ce6e506ab09199e0f2e29e7f04e62bb9efbe`.
Main ran 43 loop-breaker tests (including six consumer cases) and 32 hotfix-chain cases successfully at this head. Consumer tests execute the real patcher against synthetic detokenizer anchors using inert local transport and a modeled Compose mount/environment resolver. They observe apply, disabled/skip inertness, malformed-knob failure, selected bytes and missing-artifact refusal, not merely echoed command fields.

Limitations: the Compose resolver is a test representation, boot simulation invokes the patcher rather than the full image entrypoint, and launcher slices are not a whole-launcher or real-rank test. [Hosted validate34973012397](https://github.com/MiaAI-Lab/DeepSeek-v4-Flash-DSpark-2x-DGX-Spark/actions/runs/34973012397) passed `bash scripts/ci-validate.sh`, including all43 loop-breaker cases. It checked out synthetic merge `a0bcd20719e8add544dfdc3635d74385d23b6e3a`; its tree `2b8bcbd451ba134b65afb28a9bec8a121864c4ad` exactly matches the published head. The CPU suite reports image/Torch-dependent skips; success is not pinned-image or live qualification.

### Historical contributor observations — not qualification of this head
The original author reported a 388K-token reproducer, runaway trials ending at 425–493 tokens on the sixth repeat, a 12-case calibration, and a production day without observed loops or false stops on a two-Spark pair. The original report also discusses sampling-parameter observations. Those default-on-era live results were not reproduced for this reworked default-OFF head and are retained as historical attribution only; see the linked report for methodology.

### Maintainer-owned HOLD — do not merge or infer live qualification
Owner: maintainer/Main for integration order, hardware access and runtime qualification scope; the designated merge owner retains the hold.
Reason: actual pinned-image entrypoint/detokenizer compatibility, remote Compose and multi-rank startup, and live detector calibration remain unqualified.
Unblock: pass required cumulative exact-head CPU checks (refresh canonical-main integration normally if needed), select and authorize runtime qualification, review the resulting exact-head evidence, then explicitly lift the hold. Source-only approval and publication do not lift it; this is not a missing-author-evidence classification.

### DIFF
--- a/.env.dspark.example
+++ b/.env.dspark.example
@@ -230,6 +230,17 @@ VLLM_PREFIX_CACHE_RETENTION_INTERVAL=4096
 # DSPARK_SUPPRESS_STOPS_IN_REASONING=1
 # VLLM_SUPPRESS_STOPS_IN_REASONING=0
 
+# Issue #82 bounded loop-breaker: finish a generation cleanly
+# (stop_reason=loop-breaker) when the same normalized line repeats >=6 times
+# (>=15 for short "Now."-style sentence lines). Never counts code/JSON/DSML
+# lines; holds fire inside a DSML tool block; min 64 output tokens. Bounds the
+# runaway announce-loop (4,096 tokens of babble -> ~450-token clean stop).
+# Default on. DSPARK_SKIP_LOOP_BREAKER_HOTFIX=1 skips the patch.
+# DSPARK_LOOP_BREAKER=0
+# DSPARK_LOOP_BREAKER_REPEATS=6
+# DSPARK_LOOP_BREAKER_SHORT_REPEATS=15
+# DSPARK_LOOP_BREAKER_MIN_TOKENS=64
+
 # Issue #52 / PR #53 (assistant-final continuation hotfix): when a request's
 # messages array ends with an assistant message — or with a trailing
 # `latest_reminder` annotation directly after that closed assistant turn — the
--- a/docker-compose.dspark.yml
+++ b/docker-compose.dspark.yml
@@ -57,6 +57,8 @@ services:
       - ${DSPARK_ISSUE133_HOTFIX:-./patches/hotfix-dsv4-issue133-triton-specialization.py}:/opt/hotfix-dsv4-issue133-triton-specialization.py:ro
       # Client stop strings stay dormant until </think> (Tony/Capicua25x Patch 5).
       - ${DSPARK_SUPPRESS_STOPS_HOTFIX:-./patches/hotfix-dsv4-suppress-stops-in-reasoning.py}:/opt/hotfix-dsv4-suppress-stops-in-reasoning.py:ro
+      # Issue #82 bounded loop-breaker (repeated-line detector in the detokenizer).
+      - ${DSPARK_LOOP_BREAKER_HOTFIX:-./patches/hotfix-dsv4-loop-breaker.py}:/opt/hotfix-dsv4-loop-breaker.py:ro
       # A request ending with an assistant message must still get a generation
       # header, or the model generates from a dead state (empty turns /
       # hallucinated DSML markup loop). Upstream issue #52, PR #53.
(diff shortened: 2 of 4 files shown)


plotarmordev (changes requested):
HOLD at `59d00c2`; useful mitigation idea, but not merge-ready.

1. The documented controls do not reach the serving process. Compose never passes `DSPARK_SKIP_LOOP_BREAKER_HOTFIX`, `DSPARK_LOOP_BREAKER`, or any threshold variable into the service. Rendering the submitted Compose shows zero `LOOP_BREAKER` environment entries, so the patch is always applied and runtime/default/threshold overrides are inert. Add explicit pass-through and tests against rendered service env.

2. Full repository CI currently fails: `scripts/test-python-hotfix-failclosed.py` has two failures because the new default-on patch chain was not integrated into its ordering/skip contracts. The claimed 12-case detector calibration is not committed, and there is no CI coverage for apply/idempotence/drift/status, stock and suppress-stops header states, runtime gates, malformed knobs, finish/stop reason, or false positives.

3. The claimed code safety is not implemented. `_lb_check` only excludes lines containing braces, quotes, `DSML`, or a fullwidth bar; it has no code-fence/indented-code/YAML/markdown state. Six occurrences of an ordinary code line such as `return value` after 64 output tokens satisfy the default detector. Because the patch is default-on for every request, this can silently stop healthy coding output. Make this exact-1 opt-in until a broad committed false-positive suite supports default-on, and either narrow the detector or correct the contract.

4. Publication/idempotence is below the repository's current hotfix standard: direct `write_text` can expose a truncated/partial target, rollback is non-atomic, `MARK` + `_lb_check` accepts partial post-state as idempotent, and `--status` returns success for a marker-only/NOT APPLIED target. Use a mode-preserving atomic same-directory replace, full postcondition verification, and nonzero status for anything except the exact applied image.

5. Validate the integer knobs once with positive bounded ranges. Current bare `int(...)` imports crash on malformed values; zero/negative repeat thresholds fire on the first eligible line. Also bound per-request detector state or replace the unbounded dict with the minimal tail state.

Finally merge current main into the contributor branch (no history rewrite): the PR conflicts in Compose and its old head predates the restored `DSPARK_MAX_INFLIGHT_PREFILLS=1` default. Do not live-test the stale head. Once the items above and exact-current-main CI are green, this is reasonable as a default-off experimental P2 mitigation; it should not close #82 as a root-cause fix.

plotarmordev:
**Maintainer status: CHANGES REQUESTED / HOLD — the existing review at this head stands unresolved.**

Reviewed head: `59d00c2497b8dcf52756902bd18335543eb03d16`

This references the maintainer CHANGES_REQUESTED review of 2026-08-29 at this exact head, which has had no author response. A fresh static pass re-confirms two of its items against the diff:

1. **Runtime knobs never reach the serving process.** The compose diff adds no `environment:` pass-through for `DSPARK_SKIP_LOOP_BREAKER_HOTFIX` / `DSPARK_LOOP_BREAKER*`, while sibling knobs on main are passed via `environment:`. The in-container gate therefore always reads the default and the documented kill switch is inert — while the breaker is default-on. Default-on plus an inert kill switch is the worst combination.
2. **False-stop risk on legitimate code output.** `_lb_check` excludes only lines containing braces/quotes/DSML markers and tracks no code-fence or indented-code state, with counts accumulating over the whole generation. As few as six repeated `return value`-style lines past the token floor satisfy the default threshold, so a long legitimate structured generation can be silently truncated — the same class of harm this mitigates.

**Required behavior.** The documented controls must actually reach the serving process and gate the hotfix at runtime; and the breaker must not fire on fenced/indented code or other legitimate structured output (code-state awareness, and repetition counted within a bounded window rather than across the whole generation).

**Required validation.** Rendered-env tests proving the knobs propagate into the container; committed false-positive calibration over representative legitimate outputs (code-heavy included); the repo CI suite re-run at the reworked head; then live agent-lane validation.

**What would unblock.** All four items from the 2026-08-29 review addressed, with the validation above. The mitigation idea remains valuable; this hold is about the current implementation, not the goal.

plotarmordev (approved):
Source-only approval at 2b5a4f394ee77e92c921d85d171e99f274433a96. Complete-image classification, skip-independent status, postwrite restore handling and bounded numeric conversion corrections are accepted. Five-control forwarding, canonical selected bytes on both workers, and early enabled/unskipped selected-file admission address rank parity and the formerly missing override edge.

The requested recorder correction is accepted: tests now execute the mounted real patcher against synthetic anchors and assert consumer apply/inert/failure outcomes; old source/default-only pin tests were removed. The fake Compose resolver and extracted launcher slices are not real Compose, full image-entrypoint or live-rank proof. Main supplied 43 loop-breaker and 32 chain passes at this exact head. Historical contributor live claims are not fresh qualification; default remains OFF and issue82 is not resolved.

Maintainer/Main owns the unlifted HOLD for cumulative integration checks and selected runtime/hardware qualification. Do not merge until those exact-head gates pass and the maintainer explicitly lifts HOLD; this approval is source-only. Outgoing historical content and candidate provenance were reviewed; supplemental signature scan found no credential or owner-local deployment hits.

plotarmordev (changes requested):
## Hold update

Public head checked: `cdcbda8167b4ea3738011618dc8b4b00f601ab96`.

**Changes still needed for integration:** Route the four loop-breaker file-sync calls through the `dssh`/`dscp` wrappers from #249. Update `scripts/test-loop-breaker.py` so its launcher fixture includes the transport and port-guard pieces it now uses. Without that fixture update, the combined CPU suite had six failures with `dssh: command not found`.

**What passed:** After local integration fixes, the CPU checks and actual serving-process checks on both ranks passed. The bounded repeated-line case reached its output limit with the breaker off and stopped with `loop-breaker` with it on. The fenced-code control completed normally in both modes. The synthetic 20-turn trajectory also passed in both modes: useful controls, but not a reproduction or resolution of issue #82.

**Next action — me, @plotarmordev:** Publish and review the prepared transport/fixture corrections. The author does not need to redo our local fixes or repeat the live run now.

**Before ready:** Run `python3 scripts/test-loop-breaker.py -q` and `bash scripts/ci-validate.sh` on the published integrated head. Keep the breaker off by default. Preserve the real-consumer on/off, repeated-output and legitimate-code controls; do not present this mitigation as a root-cause fix for #82.

PR #157, opened 2026-08-29
fix(boot): make Spark recovery durable (DGX-8)

jayefel:
## Summary

- add idempotent head/worker recovery installation and verification commands for DGX Spark
- recover from the observed MT7925 kernel-oops failure mode with a wired-route guard, kernel panic recovery, and the system watchdog
- make Docker lifecycle deterministic with an init process, bounded shutdown grace, and a configurable worker orphan guard

Linear: [DGX-8](https://linear.app/ticketfi/issue/DGX-8/make-dgx-spark-boot-and-inference-recovery-durable)

## Why

The two-node MIA deployment repeatedly became unreachable after the stock MT7925 reset path oopsed and wedged RTNL. A physical reboot restored it, but the same boot/Docker/orphan-rank conditions could recur. These controls were installed and verified on both affected Sparks before being rebased onto current upstream `main`.

## Test strategy

```test-strategy
bash -n ops/dgx-recovery/install.sh ops/dgx-recovery/verify.sh ops/dgx-recovery/mia-worker-orphan-guard scripts/ci-validate.sh
docker compose -f docker-compose.dspark.yml config --quiet
ops/dgx-recovery/install.sh invalid  # exits 2
ops/dgx-recovery/verify.sh invalid   # exits 2
git diff --check
```

The repository's full CPU validation passes the changed shell/config gates on macOS, then hits an existing GNU/BSD `sed` incompatibility in `scripts/test-numeric-knob-validation.py`; the required GitHub Linux workflow remains the authoritative full gate.

## Runtime evidence

- recovery verifier passed on both installed Sparks after reboot
- DeepSeek V4 Flash `/v1/models` and inference recovered
- two-node NCCL `all_reduce_perf` remained correct (`#wrong=0`), peaking at 23.68 GB/s bus bandwidth

## Safety

- install fails closed when the default route is Wi-Fi
- no secrets or machine environment files are committed
- host/container/API addresses remain overridable for another Spark pair

### DIFF
--- a/docker-compose.dspark.yml
+++ b/docker-compose.dspark.yml
@@ -2,6 +2,7 @@ services:
   vllm-dspark:
     image: ${DSPARK_VLLM_IMAGE:-ghcr.io/anemll/dspark-vllm-gx10:0.1.1}
     pull_policy: if_not_present
+    init: true
     entrypoint: []
     network_mode: host
     ipc: host
@@ -9,7 +10,7 @@ services:
     # Issue #38/#39: bring a crashed worker back. Keep grace short — vLLM
     # ignores SIGTERM while loading and a long period hangs `compose stop`.
     restart: ${DSPARK_RESTART_POLICY:-unless-stopped}
-    stop_grace_period: ${DSPARK_STOP_GRACE:-10s}
+    stop_grace_period: ${DSPARK_STOP_GRACE:-30s}
 
     # Only rank 0 serves HTTP; the worker runs HEADLESS with no :8888 listener.
     # Gate on $HEADLESS so the worker is not marked unhealthy forever. $$HEADLESS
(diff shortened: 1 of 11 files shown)


jayefel:
Post-reboot hardware proof on both DGX Sparks is green for this exact patch:

- wired default route on both nodes
- MT7925 module absent after boot
- panic-on-oops and hardware watchdog active
- Docker enabled and active
- both MIA containers running with init enabled
- worker orphan guard enabled and active
- `/v1/models` ready immediately
- real DeepSeek V4 request returned `DGX_READY` (`chatcmpl-ba38340d93205326`)

The remaining gate is GitHub workflow run 33266128777, which is still `action_required` and needs repository-admin approval.

Issue #79, opened 2026-08-17
vllm found wasting core cpu on dgx spark on tp=2


fche:
https://nacyot.github.io/artifacts/vllm-spin-wait-gb10-en/  offers analysis and a one-line patch for vllm .../shm_broadcast.py to significantly lower temps / waste.

No known vllm-project/vllm issue filed yet.

lilianmoraru:
Had this while the Sparks were idling (just finished a task) - seemed random, sometimes the CPU is pegged after some work and it overheats, sometimes it is not.

richardbiely:
Adressed here:
https://github.com/MiaAI-Lab/DeepSeek-v4-Flash-DSpark-2x-DGX-Spark/pull/74

Deep-AI-Evo:
**English** (中文见下) — *updated 2026-08-19 with controlled A/B data, replacing the earlier quick-test numbers*

Independent reproduction on the same deployment shape as the original analysis: 2× DGX Spark (GB10), TP=2 `mp` backend, DeepSeek-V4-Flash-0731 FP8, vLLM 0.25.2 (`ghcr.io/anemll/dspark-vllm-gx10:0.1.1`).

**Symptom before the patch**: aggregate CPU looked harmless (<20%), but `mpstat -P ALL` showed 3-4 performance cores pinned at ~100% during decode; under sustained real-world load the SoC peaked at **95 °C** (OS force-shutdown is 104.8 °C). The GPU itself stayed much cooler.

**Fix applied**: one-line patch `busy_loop_s: float = 1` → `0.002` in `vllm/distributed/device_communicators/shm_broadcast.py`, built as a thin derived image layer, deployed on both nodes.

**Controlled A/B measurement** (same prompt, max_tokens=4000, 120 s of sustained load per run, both nodes sampled every 30 s, cooled to <56 °C between runs, full image swap + service restart between arms). Peak temperatures within each 120 s window:

| Concurrency | Version | head SoC peak | worker SoC peak | Total decode |
|---|---|---|---|---|
| 1 | original (1s) | 91.2 °C | 84.1 °C | 40.5 tok/s |
| 1 | spinfix (2ms) | **78.5 °C** | **81.5 °C** | 39.8 tok/s |
| 2 | original (1s) | 93.1 °C | 86.9 °C | 59.3 tok/s |
[...]

PR #158, opened 2026-08-30
fix(shm): bounded dispatch-ring reader re-check (backport, issue #117)

zerokeeper:
Backports upstream vLLM SHM_READER_RECHECK_INTERVAL_MS with a 1000 ms
ceiling so a lost PUB/SUB notify on the EngineCore->local-worker ring
self-heals instead of stranding the reader in poll(None) forever
(rank0 quiescent TP-pair loss; dispatch-trace: disp 27590/27591 idx
8->9->0 with no recv on rank 0; 600 s NCCL kill on rank 1).


de1tydev:
## Production follow-up: corrected #45224 backport deployed on the affected TP=2 pair

We deployed a hardened variant of this idea on the production 2x DGX Spark pair that produced captures #4/#5. The bounded reader re-check matches the mid-serve quiescent-rank failure very well, but I recommend revising the current PR head before merge.

### Why we expedited it

Before the rollout the pair froze four times in about 91 minutes:

| onset (UTC) | frozen iteration | NCCL RAS discriminator |
|---|---:|---|
| 06:44 | 5575 | rank 1 ahead by 59 custom AllReduces |
| 07:15 | 2191 | rank 1 ahead by 59 custom AllReduces |
| 07:53 | 3650 | rank 1 ahead by 59 custom AllReduces |
| 08:15 | 954 | rank 1 ahead by one AllGather (1283 vs 1282) |

The last capture was the clearest: EngineCore waited for a response, the head-local `Worker_TP0` was indefinitely parked in `shm_broadcast.acquire_read/dequeue`, and remote `Worker_TP1` had already returned to socket receive. Both py-spy samples were byte-identical. There was no OOM, Xid, fatal AER, RoCE error, traceback, new serve-time JIT, or #117-class NCCL timeout before the external watchdog recovered the pair.

### What we actually deployed

We used the merged upstream implementation as the source of truth: vLLM PR [#45224](https://github.com/vllm-project/vllm/pull/45224), merge `10c75477b07c2f1a361f54b7357af1019bba5fd8`.

Key differences from the current `0cdde02` head:

1. **Upstream fidelity:** keep the tested `SHM_READER_RECHECK_INTERVAL_MS = 5000`, rather than changing it to 1000 without wakeup/CPU measurements.
2. **Backport both upstream hunks:** bounded idle-reader re-check **and** the `finally`-based read-slot release when a consumer raises.
3. **Separate rollback:** a dedicated `DSPARK_SKIP_ISSUE117_RECHECK_HOTFIX` leaves the existing #79 `busy_loop_s=0.002` CPU/heat fix untouched.
4. **Exact state classification:** accept only pinned `vllm==0.25.2.dev0+g752a3a504.d20260714` plus complete stock or complete post-patch identities for the constant, timeout method, and read-slot release regions. Mixed/partial/marker-only state, symlinks, and version drift are incompatible.
5. **No production `assert` source locks:** all compatibility checks are unconditional. The candidate is built and compiled in memory, staged beside the target, fsynced, atomically renamed, re-read and verified; any post-publication failure restores and verifies the original bytes and metadata.
6. **Cluster-wide gate:** worker and head run non-mutating `--check` before either service starts; each container then runs apply followed by `--status` before `exec vllm`.
7. **Focused tests:** indefinite no-warning timeout returns 5000; `finally` release is present; apply/idempotence; each of the three independent anchor drifts; partial post-image; symlink; version lock; mode/metadata preservation. Full recipe CPU CI also passes.

Two concrete fail-closed problems reproduced against the current PR implementation:

- Keeping `[shm-reader-recheck]` but changing the constant from 1000 to 999 still makes a rerun exit 0 as "already applied".
- With `PYTHONOPTIMIZE=1`, removing the constant anchor disables the Python `assert`; the patcher exits 0 and writes a timeout method that references an undefined `SHM_READER_RECHECK_INTERVAL_MS`.

So marker presence alone should not be treated as an applied-state proof, and `assert` should not guard production compatibility.

### Production result

The corrected backport started on both ranks at 08:34:57/08:35:00 UTC. Both runtime sources verify as `patched` with identical SHA-256 `cb9923071483fdb5affd224a0a55bc2495bb61d69f6ea88c25c19fd5e04b6527`; the live constant and `timeout_ms(None, False)` both resolve to 5000.

Admission gates:

- minimal generation passed
- boot-shape warmup 47/47 in 25 s; sampler postcondition MET
- issue #136 live matrix 145/145, health 200 before/after
- greedy 8K: TTFT 3.463 s, 65.88 tok/s
- greedy 32K: TTFT 13.582 s, 64.32 tok/s
- initial real-traffic observation: 17 one-minute samples, iteration 1917 -> 5896, max running 4 / waiting 3, zero restart/OOM/frozen event

Latest readback at 13:27 UTC, about 4 h 52 min after container start:

- both containers healthy, `RestartCount=0`, `OOMKilled=false`
- iteration 56941, active real traffic
- zero post-cutover watchdog frozen samples or restart thresholds
- zero `No available shared memory broadcast block`, EngineDead, traceback, NCCL warning/timeout, runtime JIT, or CUDA error after warmup
- 1362 successful Chat Completions HTTP 200 recorded by the head access log

This is strong short-soak evidence for the mid-serve lost-notify variant, not proof of long-horizon closure.

### Scope boundary

The new [orphaned `/dev/shm/psm_*` startup-stall capture](https://github.com/MiaAI-Lab/DeepSeek-v4-Flash-DSpark-2x-DGX-Spark/issues/117#issuecomment-5468189535) should remain a separate lifecycle/cleanup track. The bounded reader re-check prevents an already-written local dispatch from being stranded after a lost notify; it does not delete orphaned POSIX SHM objects or prove that every startup stall has the same cause.

Suggested disposition for this PR: preserve the bounded re-check, revise it to the fail-closed contract above, add the upstream lost-notify regression tests, quantify the soak, and describe it specifically as the mid-serve quiescent-reader fix. I can provide a clean follow-up patch if the maintainers want the exact implementation rather than this design summary.

plotarmordev (changes requested):
P1 disposition: the current head `0cdde02031968e03acf41284b4241baf7a6da116` remains **changes required** and must not merge.

Evidence against this head:
- It couples issue #117 rollback to the existing issue #79 spin-wait switch.
- It uses a 1,000 ms interval instead of upstream vLLM's 5,000 ms and omits the upstream `finally` read-slot release.
- Marker-based/idempotence checks admit partial states; the compatibility checks are not safe under `PYTHONOPTIMIZE=1`.
- It has no exact installed-source identity gate, two-rank non-mutating preflight, or permanent regression suite.

Acceptance contract independently reviewed and validated from primary bytes:
- exact upstream source pin `752a3a504`: 39,864 bytes, SHA-256 `7ff67c2ef6b8a33a13b11aa3cb202da7887d1d44eed27c6a02d817ea24807d61`;
- exact upstream merge `10c75477`: 40,312 bytes, SHA-256 `911e0dd65e0a0c6346e4f8f2120d2417fefe431ce5f2618ba9e9c9e1986faf23`;
- exact issue-79-only installed state: 39,868 bytes / `423234a203429b4d74aa48021a3ea02f3811be7d6a1938369feed298254fd51f`;
- exact combined state: 40,316 bytes / `30d8b62817adab4fabde8ddc6ce9a0f4b71899b80f70e8a50c688f5e63a46b0f`;
- complete four-state classification, atomic publish/rollback, no `assert` fail-open, worker-then-head `--check` before either rank starts, and an issue-117-only rollback switch.

Validation evidence: focused suite 22/22 under normal Python and `-O`; complete `scripts/ci-validate.sh` passed. Live TP=2 canary passed exact patch/status on both ranks, text/image/concurrent smokes, ~8K and ~32K prompts, and a 10.6-minute soak with zero restarts/OOM/errors. A separate cold boot with issue #117 skipped retained issue #79 on both ranks and stayed healthy. The untouched original deployment was restored and re-smoked.

Please rebase this PR on current `main` and update its head to meet the acceptance contract above. Repository policy prevents publishing code or patches from the deployment checkout; this comment contains review/validation evidence only.

plotarmordev:
@zerokeeper The current exact head is still `0cdde02031968e03acf41284b4241baf7a6da116`. Please update/rebase this PR to the reviewed acceptance contract above; I can immediately re-run exact-head review and validation when the branch changes.

PR #161, opened 2026-08-31
Drop digest from DSPARK_VLLM_IMAGE example

mrsage-AI:
The example env file pins the image by digest, but build-dspark-vllm-runtime.sh reuses the same variable as the output tag for the stage-c build, so a fresh copy of .env.dspark.example fails with:

  ERROR: failed to build: failed to solve: refusing to create a tag with a digest reference

Docker can pull by digest but cannot create a tag with one. This drops the @sha256 suffix so the value is a plain name:tag. Alternatively the pinned pull and output tag could be split into two variables.


plotarmordev:
Thank you for reporting the digest-as-build-tag failure. Closing as superseded by merged #190 (`22bdac7`).

Main now keeps the immutable digest pin for the prebuilt runtime image while `build-dspark-vllm-runtime.sh` writes a separate valid local Stage-C tag instead of passing the digest reference to `docker build -t`. That fixes the fresh-example build failure without weakening the image pin. Dropping the digest from `.env.dspark.example` would now undo the intentional supply-chain guarantee, so this patch should not be merged.

PR #163, opened 2026-08-31
fix(issue138): let `--status` recognise a source another patch has extended

msdjzmst:
> Stacked on #162 — please review that one first. This branch contains only the
> `--status` change; the diff against #162 is two files.

`_source_state()` decides between stock and applied by matching the whole
`input_item_parsing` method verbatim. That is the right contract for `apply()`
— it must refuse to patch a source it does not recognise — but it also means
`--status` breaks as soon as any later patch touches the same method, even when
both patches are working correctly.

With the Codex `agent_message` patch from #162 applied on top, a live 2x DGX
Spark pair reported:

```
[FAIL] [issue138-hotfix] complete input_item_parsing source lock mismatch
       (old=0, new=0, marker=1)
```

while the other patch correctly reported `already applied and verified
(issue138-applied)`, and both conversions worked. Reading the patched file in
the container confirmed both branches present and the issue138 marker intact:

```python
content = item.get("content")
legacy_assistant_output = (...)          # issue138 branch, live
agent_message_content = item.get("content")   # agent_message branch
```

Only the status check was wrong.

## Change

`apply()` is unchanged. `status()` now uses a separate `_status_source_state()`
that extracts the method via `ast` and classifies:

| source | result |
|---|---|
| stock method, no marker | `stock` |
| applied method, one marker | `applied` |
| one marker, each status anchor present exactly once, in order | `applied-with-extensions` |
| anything else | `PatchError`, as before |

Anchoring on the essential inserted lines rather than the marker alone is what
keeps a damaged source from being reported as applied.

## Verification

Against the source pulled out of the running container (910 lines) and three
deliberately broken variants:

| variant | result |
|---|---|
| rename `legacy_assistant_output` | rejected (anchor count 0) |
| delete the branch body | rejected (does not parse) |
| duplicate the marker | rejected (marker count 2) |
| synthesised stock | `stock` |
| synthesised issue138-only | `applied` |
| container source (both patches) | `applied-with-extensions`, exit 0 |

In the container the status is now
`already applied and verified (applied-with-extensions)` with exit code 0,
where it previously exited non-zero.

## Scope

Nothing in the start path reads this exit code today — compose calls the
patcher without `--status` — so this **changes reporting only**, and the
running service was never affected.

It seemed worth fixing anyway: the recipe carries a growing number of patches,
and any two that touch one method currently make each other's status
unreadable. The same limitation would hit the #162 patch once a third patch
lands on that method.


plotarmordev:
Thank you for isolating the stacked `--status` problem. Closing as superseded by the maintainer integration in #201, which is now on `main`.

Current main handles the exact issue-138 + Codex combined postimage across same-container restarts with an exact method hash and dedicated tamper/restart coverage. That covers this PR's operational objective while preserving a stricter fail-closed source lock. This branch's broader `applied-with-extensions` anchor recognition is therefore no longer needed and would accept more source shapes than the merged implementation.

No contributor action is required.

PR #166, opened 2026-08-31
fix(vision): do not 400 on a bare <image> mention (issue #165)

### LINKED ISSUE #165: Vision-Exp role check false-positives on the literal string `<image>`: any client whose **system prompt** contains that text gets a 400, with no image in the request
## Summary

`patches/hotfix-dsv4-vision-exp.py` enforces "images in user messages only" by
**substring-matching raw text**. For any non-`user` role it treats the literal
7-character string `<image>` as proof that the message contains an image.

`<image>` is ordinary text. It appears in prose, in HTML/markdown examples, and in
the system prompts of real coding agents. So a request with **no image anywhere**
is rejected with:

```
{"error":{"message":"Images are supported in user messages only: images in 'system' messages return a 400 error.","type":"BadRequestError","param":null,"code":400}}
```

This makes `deepseek-v4-flash-vision-exp` unusable with any OpenAI-compatible client
whose system prompt happens to mention `<image>` — the failure is unconditional and
has nothing to do with sending images. We hit it with **opencode**: every single
request failed, including plain text chat.

## Reproduction

Two `curl`s. No image, no special client, nothing but a system prompt.
Replace `8888` with your `VLLM_PORT` if you changed it.

**1. Control — identical request, system prompt WITHOUT `<image>`:**

```bash
curl -sS http://127.0.0.1:8888/v1/chat/completions \
  -d '{
    "model": "deepseek-v4-flash-vision-exp",
    "messages": [
      {"role": "system", "content": "You are a helpful assistant."},
      {"role": "user",   "content": "Say OK"}
    ],
    "max_tokens": 2000,
    "chat_template_kwargs": {"thinking": false}
  }'
```

Result: **200 OK**, content `OK`.

**2. The bug — same request, three extra words containing `<image>`:**

```bash
curl -sS http://127.0.0.1:8888/v1/chat/completions \
  -d '{
    "model": "deepseek-v4-flash-vision-exp",
    "messages": [
      {"role": "system", "content": "You are a helpful assistant. Markdown images look like <image> tags."},
      {"role": "user",   "content": "Say OK"}
    ],
    "max_tokens": 2000,
    "chat_template_kwargs": {"thinking": false}
  }'
```

Result: **HTTP 400**

```
{"error":{"message":"Images are supported in user messages only: images in 'system' messages return a 400 error.","type":"BadRequestError","param":null,"code":400}}
```

The only difference between the two requests is the substring `<image>` inside a
plain-text system prompt. There is no `image_url` part, no attachment, no
multimodal content of any kind in either request.

Server-side log line for #2:

```
(APIServer pid=1) ValueError: Images are supported in user messages only: images in 'system' messages return a 400 error.
(APIServer pid=1) INFO:     10.x.x.x:41610 - "POST /v1/chat/completions HTTP/1.1" 400 Bad Request
```

## Root cause

`patches/hotfix-dsv4-vision-exp.py`, in `ENC_ROLE_INJECT`:

```python
def _dspark_vision_value_has_image(value) -> bool:
    if isinstance(value, str):
        return IMAGE_PLACEHOLDER in value or "<image>" in value   # <-- substring match on free text
    ...
```

used by:

```python
def _validate_no_image_sp_tokens(msg):
    """Official restriction: images in user messages only (system/assistant -> 400)."""
    role = msg.get("role")
    if role in ("user", "developer"):
        return
    if _dspark_vision_value_has_image(msg.get("content")) or _dspark_vision_value_has_image(
        msg.get("content_blocks")
    ):
        raise ValueError(
            "Images are supported in user messages only: "
            "images in " + repr(role) + " messages return a 400 error."
        )
```

The `IMAGE_PLACEHOLDER` check is fine — that token is model syntax and should never
appear in user-authored text. The `"<image>" in value` check is the problem: it
matches an opening tag anywhere in arbitrary prose, with no requirement that it be
a real `<image>path</image>` reference.

The structured check (`block.get("type") in ("image", "image_url")`) is correct and
is not what fires here.

## Suggested fix

Require the documented **paired** form rather than a bare opening tag, e.g. match
`<image>...</image>` (a regex such as `<image>[^<]*</image>`) instead of the
substring `<image>`. That keeps the intended restriction — a genuine
`<image>path</image>` reference in a system/assistant message is still rejected —
while no longer firing on text that merely mentions the tag.

Keeping the `IMAGE_PLACEHOLDER` substring check as-is is correct.

## Impact

Any OpenAI-compatible client that mentions `<image>` in its system prompt cannot use
this model at all, for text or images. Because the check runs per message before
anything else, the failure is total rather than degraded, and the error message
actively misleads: it says images are present in a system message when none are.

## Environment

- Recipe: `MiaAI-Lab/DeepSeek-v4-Flash-DSpark-2x-DGX-Spark` @ `7963d432bfe717f2a7249a53fb4c0673c239c03e` (main, 2026-08-31)
- Image: `ghcr.io/anemll/dspark-vllm-gx10:0.1.1`, digest `sha256:a83948492cf13df455170fb42885f5ef4db54fefe0feff0f841ecbff464ac9d8`
- vLLM: `0.25.2.dev0+g752a3a504.d20260714`
- Checkpoint: official `deepseek-ai/DeepSeek-V4-Flash-Vision-Exp`
- Hardware: 2x NVIDIA DGX Spark (GB10), TP=2 over ConnectX-7, `mp` executor, `--nnodes 2`
- Serving config otherwise stock: `MAX_MODEL_LEN=1048576`, `MAX_NUM_SEQS=6`,
  `GPU_MEMORY_UTILIZATION=0.835`, `kv_cache_dtype=nvfp4_ds_mla`, DSpark k=6 probabilistic.
- Local deviations from the shipped defaults (neither plausibly related to a role
  validation check, listed for completeness): served on port 8000 instead of 8888;
  model served from a local directory path instead of an HF repo id; `DEFAULT_THINKING=low`.

Vision itself works correctly on this deployment — an `image_url` part in a `user`
message returns an accurate description. The only defect is the false positive above.

## Not tested

- Whether the same false positive fires for `assistant`-role messages (the code path
  is shared, so it very likely does, but I did not verify it).
- Whether a paired `<image>...</image>` in a system message is rejected (expected, and
  intended behaviour).

MiaAI-Lab:
## Summary

- Fixes [#165](https://github.com/MiaAI-Lab/DeepSeek-v4-Flash-DSpark-2x-DGX-Spark/issues/165): the Vision-Exp role check treated the substring `<image>` in any non-user message as an image, so OpenCode (and any client whose system prompt mentions the tag) got HTTP 400 on **text-only** requests.
- Keep `IMAGE_PLACEHOLDER` as a substring match. Treat text as an image only when a **paired** `<image>…</image>` is present, or when a structured part is `type` `image` / `image_url`.
- Existing container copies still have the old inject; `--status` now requires the paired-tag mark, and the patcher rewrites a stale encoding.py on recreate (`./stop` then `./start`; `docker compose restart` keeps the old bytes).

## Test plan

- [x] `python3 scripts/test-dsv4-vision-exp-hotfix.py -q` (includes the reporter's system-prompt string, assistant mention, paired-tag 400, and upgrade from the old inject)
- [x] `bash scripts/ci-validate.sh`
- [ ] After merge on the live pair: `./stop-deepseek-v4-flash-dspark.sh` then `./start-deepseek-v4-flash-dspark.sh`
- [ ] Replay both curls from #165 (control + three extra words with `<image>`) — expect 200 on both
- [ ] Confirm a real `image_url` in a user message still works, and a paired `<image>…</image>` in a system message still 400s

Made with [Cursor](https://cursor.com)


MiaAI-Lab (commented):
Cannot GitHub-approve this PR from the author account (own-PR restriction). Live-verified on the 2× DGX Spark pair after stop/start with this branch.

Encoder in the container has the paired-tag mark; the old `or "<image>" in value` inject is gone.

| request | result |
|---|---|
| control system `"You are a helpful assistant."` + user `Say OK` | HTTP 200, `OK` |
| reporter string `"Markdown images look like <image> tags."` | HTTP 200, `OK` (was 400) |
| assistant mention `"example: <image> in markdown"` | HTTP 200, `OK` |
| paired `<image>/tmp/shot.png</image>` in system | HTTP 400, intended role error |
| user `image_url` (local photo) | HTTP 200, 125 image tokens |

CI `validate` is green. Recreate both containers after merge; `docker compose restart` keeps the old encoder bytes.

PR #182, opened 2026-09-01
Add ablation support

mrexodia:
So it doesn't require downloading a new model. Based on https://huggingface.co/drowzeys/keys-DeepSeekV4-Flash-GA-0731-Dspark-Abliterated-Anchored-Tensors

The model happily performs tasks that were previously refused, but I didn't do any crazy tests except to see that regular agentic coding still works.


plotarmordev:
Thank you for the contribution and for demonstrating that the runtime projection can reduce refusals while reusing the official checkpoint.

Closing this because `main` already provides the supported user-facing abliterated mode through `ABLITERATED=1` (introduced in `d8fb750`), including model selection, cache preparation, documentation, and broader validation.

This PR is not mechanically identical: it avoids downloading the alternate checkpoint by applying an in-memory refusal-direction projection to decoder activations while running the official checkpoint. We are choosing not to maintain that second path because it adds a per-layer FP32 projection to the inference hot path, patches model internals, carries a direction artifact that needs provenance/versioning, currently conflicts with `main`, and has not yet shown performance, quality, multimodal, long-context, or concurrency evidence sufficient to justify the extra support surface.

If avoiding the additional checkpoint download becomes a required product constraint, a future proposal would need reproducible direction provenance and hashes, exact-image source locking, CPU fail-closed tests, and live A/B evidence covering decode/TTFT, memory, long context, tools, images, and refusal reduction against the existing `ABLITERATED=1` path.

mrexodia:
I don’t understand the rationale of closing this. The 1x spark recipe supports both methods too?

MiaAI-Lab:
> I don’t understand the rationale of closing this. The 1x spark recipe supports both methods too?

Will re-review it. Thank you 🙏🏻

MiaAI-Lab:
Rebased onto current `main` and landed as #258.

#182 was closed and conflicting. The successor keeps your runtime 18 KiB direction path, and gates it on the Keys Hub terms (`prepare --abliterated` downloads `RESPONSIBLE_USE.md` + the direction; official Vision-Exp weights stay on disk — no 157 GiB checkpoint swap).

MiaAI-Lab (approved):
Approved. Closed/conflicting vs current main; the port with Hub-terms gating is #258.

PR #184, opened 2026-09-01
feat: add request priority scheduling

iangcarroll:
## Summary
- expose vLLM scheduling policy as a validated `SCHEDULING_POLICY` env knob
- preserve `fcfs` by default and allow `priority` for request-level integer priorities
- document the waiting-queue semantics and report the resolved policy at startup
- add a CPU gate covering valid/default values, malformed and injection-shaped rejects, validator behavior, CLI quoting, and docs

## Validation
- `bash scripts/test-scheduling-policy-gate.sh`
- `docker run --rm -v "$PWD:/repo" -w /repo python:3.12-bookworm bash scripts/ci-validate.sh`


iangcarroll:
Closing after a live two-node A/B exposed a runtime incompatibility. With the same vLLM image, checkpoint, patches, and boot sweep, `--scheduling-policy priority` completed the small block-8/16/32 requests but deadlocked one TP rank on the block-64 request for more than ten minutes; subsequent requests queued and timed out. Repeating under FCFS completed block 64 in 1 second, block 128/256 immediately, and all 47 warmup requests in 29 seconds. Priority admission has been moved in front of vLLM instead. The configuration gate itself remains fully CPU-tested, but it should not be merged for DSpark without resolving this live multi-rank failure.

PR #204, opened 2026-09-03
perf(vision-exp): classify routing kind once per forward instead of once per MoE gate

### LINKED ISSUE #203: [Perf] Vision-Exp routing-kind classification host-syncs once per MoE layer (43x per forward)
### Summary

`token_routing_kind()` ends in a `.sum().item()`, which is a device→host sync. It is called from the
MoE gate path, so it runs **once per MoE layer**, i.e. 43 times per forward on
`DeepSeek-V4-Flash-0731` (`num_hidden_layers: 43`). Prefill is not CUDA-graphed, so text-only prompts
pay it too — the sync count is the same whether or not the request contains an image.

### Where

`patches/vision_exp/image_processor.py:80` (at `c444d70`) — the sync itself:

```python
def token_routing_kind(input_tokens: Any, image_token_id: int = IMAGE_TOKEN_ID) -> str:
    """Classify a token row as text-only, image-only, or mixed (issue #175)."""
    ...
        n_img = int((flat == int(image_token_id)).sum().item())   # <-- device->host sync
```

There is no memoisation anywhere in the function or around its call sites.

Two hot-path call sites, one per MoE routing backend:

- `patches/vision_exp/apply.py:266` — inside `_compute_routing`, installed by
  `_wrap_router_compute_routing()` on **every** MoE layer's router (`moe_init` at `apply.py:396`):

  ```python
  if vl is None or capturing or token_routing_kind(input_ids) == "text":
  ```

- `patches/vision_exp/apply.py:176` — inside `fused_topk_bias_split_vl`, which the mega-MoE path
  (`moe_forward`, `apply.py:402`) swaps in as `fused_topk_bias` for the duration of each MoE
  layer's forward:

  ```python
  kind = token_routing_kind(input_tokens)
  ```

Either way the count is one classification per MoE layer per forward.

### Introduced in

`d97c808` (PR #179, fixing issue #175). The classification is correct; only its placement is the problem.

### Why it only hurts under load

`_compute_routing` already short-circuits when `torch.cuda.is_current_stream_capturing()` is true, so
CUDA-graph **replay** is unaffected. But graphs only cover decode. Prefill runs eager, so at high
concurrency a large fraction of steps are eager prefill steps, and each one blocks the launch thread
43 times. The CPU can no longer run ahead to enqueue the next layer's kernels, and the GPU drains
its queue between layers.

### Reproduction

1. Serve `DeepSeek-V4-Flash-0731` with the Vision-Exp patch enabled (`vision_n_layers > 0`).
2. Drive concurrent **text-only** traffic (no images needed — the sync fires regardless).
3. Watch `nvidia-smi` / the vLLM metrics endpoint during the run.

### Observations

Observed by our operator on a 2x DGX Spark (GB10) deployment during the regression window
(dashboard + `nvidia-smi` readings, not a controlled instrumented run — we are reporting the
symptom that led us to the code, and the load-bearing evidence for this report is the sync
count below, which is deterministic):

| Signal | Idle/low load | Under concurrency |
|---|---|---|
| GPU utilisation | ~96% | 71–91% |
| Board power | — | 37–64 W |
| `vllm:num_requests_running` | — | saturated |
| `vllm:gpu_cache_usage_perc` | — | ~10% |
| `vllm:num_preemptions_total` | — | 0 |

The shape of this — running queue full, KV cache nearly empty, zero preemptions, and yet utilisation
and power both down — is a launch-bound stall, not a memory- or scheduler-bound one.

### Sync count, measured

Counting `torch.Tensor.item` calls for one forward's worth of gate classifications (43 layers),
pure-Python, no model loaded:

```
=== c444d70 (unpatched) ===
  text-only   43 MoE gates -> .item() host syncs = 43  kind={'text'}
  with-image  43 MoE gates -> .item() host syncs = 43  kind={'mixed'}
```

### Impact

Every Vision-Exp deployment, including pure-text workloads, on every eager (non-graph-captured) step.

---

antihurt:
Fixes #203.

## Exact-head design

Exact head: `4b841db707ae21603def0932c26cc6f257151426`.

The model owns one mutable one-slot routing-kind cell, shared only with that model's Vision-Exp MoE gates. `embed_input_ids` classifies once per forward: text-only publishes `text` without a tensor scan; multimodal forwards reuse the placeholder-count sync already required by the stock merge check. The first eager MoE gate consumes that model-local publication and memoizes it on the current vLLM `ForwardContext`; later gates read the context without another host sync.

There is no process-global pending kind, owner-ID registry, sequence counter, or log-once state. A second target model cannot consume the first model's publication, and the untagged MTP drafter performs the existing token-scan fallback. The CUDA-graph capture short-circuit remains before routing-kind resolution.

Publication is fail-closed: a locked/immutable `ForwardContext`, a Vision-Exp model with no attachable gates, or an invalid empty multimodal classification raises instead of silently restoring 43 per-layer scans.

## Deterministic validation

Pinned runtime-image source fixture:

- text-only routing scan calls: `43 -> 0`
- image-only routing scan calls: `43 -> 1`
- mixed routing scan calls: `43 -> 1`
- cross-model isolation: PASS
- graph-capture publication behavior: PASS
- locked-context fail-closed behavior: PASS
- `tests/test_issue175_routing_kind_once.py`: PASS
- `scripts/ci-validate.sh`: PASS

The rework removed 357 lines and added 167 relative to the prior PR head. The complete PR footprint against `main` is +585/-40 across three files.

## Local evidence

local evidence gate: blocked

## Files

- `patches/vision_exp/apply.py`
- `patches/vision_exp/image_processor.py`
- `tests/test_issue175_routing_kind_once.py`

### DIFF
--- a/patches/vision_exp/apply.py
+++ b/patches/vision_exp/apply.py
@@ -14,9 +14,11 @@ from torch import nn
 from .image_processor import (
     IMAGE,
     IMAGE_TOKEN_ID,
+    current_routing_kind,
     is_unregistered_router_bias,
     is_vision_exp_weight_name,
-    token_routing_kind,
+    routing_kind_from_mm,
+    set_pending_routing_kind,
     vision_args_from_config,
 )
 from .processor import IMAGE_PLACEHOLDER, register_vision_exp_processor
@@ -152,6 +154,7 @@ def fused_topk_bias_split_vl(
     input_tokens: Any,
     hash_indices_table: Any,
     routed_scaling_factor: float,
+    kind: Any = None,
 ) -> tuple[Any, Any]:
     """Route image placeholder rows with bias_vl and no hash table (issue #175)."""
     from vllm.model_executor.layers.fused_moe.router.fused_topk_bias_router import (
@@ -173,7 +176,8 @@ def fused_topk_bias_split_vl(
         )
 
     vl = _bias_data(e_score_correction_bias_vl)
-    kind = token_routing_kind(input_tokens)
+    if kind is None:
+        kind = current_routing_kind(input_tokens)
     if vl is None or kind == "text":
         return _call(
             hidden_states,
@@ -263,7 +267,14 @@ def _wrap_router_compute_routing(router: Any, gate: Any) -> None:
             capturing = bool(torch.cuda.is_current_stream_capturing())
         except Exception:
             capturing = False
-        if vl is None or capturing or token_routing_kind(input_ids) == "text":
+        # Order matters: capture must never resolve a kind (no host sync,
+        # no .item()) -- decode graphs only ever replay text tokens.
+        if vl is None or capturing:
+            return orig(
+                hidden_states, router_logits, indices_type, input_ids=input_ids
+            )
+        kind = current_routing_kind(input_ids)
+        if kind == "text":
             return orig(
                 hidden_states, router_logits, indices_type, input_ids=input_ids
             )
@@ -279,6 +290,7 @@ def _wrap_router_compute_routing(router: Any, gate: Any) -> None:
             input_tokens=input_ids,
             hash_indices_table=getattr(router, "_hash_indices_table", None),
             routed_scaling_factor=getattr(router, "routed_scaling_factor", 1.0),
+            kind=kind,
         )
         return _append_fused_shared_experts(router, topk_weights, topk_ids)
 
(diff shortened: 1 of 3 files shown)


plotarmordev:
**Maintainer qualification evidence — merge-ready at exact head `4b841db`**

Live two-rank qualification: tested baseline = main through #212 (runtime-identical to the qualified base), transferred forward compositionally — current main additionally contains #206, which is separately qualified, default-off (no compose/default enablement), and file-disjoint from this PR's paths; current main + this head merge cleanly (verified). Candidate = this head; two full A/B/A2 boot sandwiches on the deployment's pinned Vision-Exp checkpoint.

**Safety (all 6 boots, both ranks):**
- clean boot/readiness, zero container restarts
- scheduler fairness gate: decode-lane spread ≤ 1.69× in every arm, zero preemptions
- structured-output gate: candidate == baseline in both attempts
- log scans clean (no routing-kind / image-processing errors); production state restored and verified after each attempt

**Structural (the actual claim):**
- `tests/test_issue175_routing_kind_once.py` — ALL PASS on both ranks in both attempts against the live deployed bytes: text routing scans 43→0, image 43→1, cross-model isolation PASS, locked-context fail-closed PASS
- deployed-byte gate: the three changed files on the running servers are byte-identical to this head

**Performance: no measurable change.** Attempt 1 TTFT +3.02%, attempt 2 TTFT −3.89% (sign flip); decode +0.71% / +3.56%; baseline re-run drift up to 3.66% on identical code — the 3% band is inside this deployment's boot-to-boot noise. Verdict: inconclusive, no reproducible regression.

Merge justification is the structural one: deterministic elimination of 43 per-gate host synchronizations (text) / classification-once (image), fail-closed publication, no process-global state. Independent verification recomputed all numbers from the raw records (confirmed); final architecture review: approved merge-ready.

PR #205, opened 2026-09-03
fix: make the opt-in Responses store bounded and lifecycle-safe

plotarmordev:
Supersedes #62. Credit to @bpc-oss for identifying the unbounded pinned-vLLM store and proposing the exact-source backport. This branch keeps that narrow design, rebases it onto current `main`, and rewrites the lifecycle and deployment integration around the maintainer blockers.

## Problem

Pinned vLLM retains `response_store`, `msg_store`, and background event state forever when its native Responses store is enabled. The original draft also applied while disabled, failed open at startup, and left failure/cancellation races around lazy background streams.

## Decision

- default remains off; every value other than literal `1` is normalized to `0` in both the launcher and the container command
- exact `1` requires a positive terminal-entry cap and applies the backport fail-closed before vLLM import
- worker1, optional worker2, and head all run a non-mutating exact-source check before any rank starts
- terminal response/message/event bundles share LRU eviction; retrieval and `previous_response_id` refresh recency
- continuation preprocessing pins its bundle; tracked producers, queued work, and in-progress work are not evicted
- background event state exists before the task/reader is returned, and completion/failure/cancellation always wakes readers and terminalizes status
- foreground early close/failure removes orphan message history
- active readers retain captured event state after dictionary eviction; non-background streaming retrieval retains stock lazy validation behavior
- publication is full-file source-locked, compile-before-publish, mode-preserving, atomic, verified, and rollback-capable

`DSPARK_RESPONSES_STORE_MAX_ENTRIES` bounds terminal entry count, not bytes or concurrent admission. Active/pinned/tracked work may temporarily exceed it. Stored state is process-local and is lost on any process restart; Docker restart preserves patched writable-layer bytes, not state.

## Validation

Exact head `4618eb0fb7e5e332b4a4c402e906d1dc1077ab08`:

- `scripts/ci-validate.sh` — passed (CPU recipe gate)
- focused hotfix lifecycle suite — 13 passed
- startup/fail-closed suite — 28 passed, including direct-Compose noncanonical enable values reaching service exec as `0`
- live-verifier CPU suite — 9 passed
- exact pinned image `ghcr.io/anemll/dspark-vllm-gx10:0.1.1` smoke: stock hash `fe3a48ab…` → patched hash `1b003313…`; second status `applied`; mode remained `0644`
- exact outgoing range `c444d703…4618eb0` scanned across all five commit patches/history for private-key, GitHub, AWS, OpenAI, Google, and Slack token signatures — no matches

A fresh two-Spark candidate live run has not yet been claimed. The existing dependency-free live verifier now has an optional `--store-capacity 2` gate that verifies LRU touch, oldest-terminal eviction, and newest retention after a controlled cap-2 restart.


plotarmordev:
Superseded by #206. #205 had the correct reviewed tree, but its first commit retained ancestry through the detached contributor PR head merged with main. #206 rebuilds the byte-identical reviewed tree as one feature commit directly on exact current main (`c444d703…`), with no force-push and no contributor-head merge ancestry. Exact clean-head CI passed and the outgoing one-commit range was rescanned. Continuing on #206 keeps both code and history clean.

Issue #191, opened 2026-09-02
[Bug] Vision-Exp + MTP=6 intermittently violates named/required tool contracts at concurrency 4 (142/145)


de1tydev:
## Summary

This is **not an image-understanding failure**. Native Vision-Exp image input, text chat, and the three OpenAI-compatible API surfaces passed. The deployment gate failed because concurrent strict tool calls intermittently violated the requested contract while still returning HTTP 200.

With the official `deepseek-ai/DeepSeek-V4-Flash-Vision-Exp@86f746b36186f0e567729a5c06a8c918caba82a9` checkpoint on the merged Vision recipe, the shipped `scripts/verify-issue136-xgrammar-live.py` campaign failed twice:

| run | result | failures |
|---|---:|---|
| 1 | 142/145 | named: cardinality x2; named: argument schema x1 |
| 2 | 142/145 | named: cardinality x1; required: cardinality x2 |

Every failed case returned HTTP 200. The failing synthetic labels changed between runs. Replaying the first run's three failed labels for 3 sequential and 3 concurrent waves passed 18/18, so this is load/batch-sensitive rather than a deterministic prompt failure.

## Environment

- 2x DGX Spark / GB10, TP=2 over RoCE
- recipe contains MIA `main@8494a492ad02423620c5740cab2a803ef54d0fb7` (including Vision PR #164 and the #136 / vLLM #52805 backport)
- Anemll image `ghcr.io/anemll/dspark-vllm-gx10:0.1.1` with the deployment's pinned compatibility patches
- vLLM `0.25.2.dev0+g752a3a504.d20260714`
- xgrammar `0.2.3`
- `--tool-call-parser deepseek_v4 --reasoning-parser deepseek_v4`
- DSpark speculative decoding, `MTP_NUM_TOKENS=6`
- `--async-scheduling --enable-chunked-prefill --enable-prefix-caching`
- `MAX_NUM_SEQS=6`, `MAX_NUM_BATCHED_TOKENS=8192`, `DSPARK_MAX_INFLIGHT_PREFILLS=1`
- 1M max model length, NVFP4 DS-MLA KV cache

[...]

PR #208, opened 2026-09-03
feat(dspark): opt-in block-k unlock so num_speculative_tokens follows dspark_block_size (Vision-Exp k=5)

de1tydev:
## Summary

Vision-Exp ships `num_nextn_predict_layers=3` and `dspark_block_size=5`. The pinned `SpeculativeConfig.__post_init__` maps `num_nextn_predict_layers` to `n_predict` and rejects `num_speculative_tokens > n_predict` unless it is a multiple of it ("Ensure divisibility for MTP module reuse"), so the recipe has to run `MTP_NUM_TOKENS=6` (0731 has one stage and boots k=5).

That rule is for stock DeepSeek MTP, where one MTP layer is re-run per speculative step. DSpark never re-runs a stage: `models/deepseek_v4/nvidia/dspark.py` stacks `mtp.{0,1,2}` into one non-causal backbone and drafts the whole block in a single parallel pass (anchor + k−1 noise queries), then samples with the Markov head; the checkpoint's own `inference/model.py::DSparkBlock` drafts exactly `dspark_block_size` tokens. k=6 therefore feeds the block one more noise query than it was trained with.

## Change

- `patches/hotfix-vllm-dspark-block-k.py`: source-exact patcher for `config/speculative.py` (stock `3f1abd1c…` → patched `7fffe035…`) that adds `self.method != "dspark"` to that single condition. Atomic same-directory replace, `--check` / `--status`, refuses foreign bytes, symlinks and other vLLM versions.
- `DSPARK_ENABLE_DSPARK_BLOCK_K` (default `0` = image untouched). Exact `1`: compose mounts and applies the patcher at container start; the launcher preflights `--check` on the worker (and worker2 under TP=3) then the head, syncs the patcher to the worker, prints the effective value, and relaxes its own `MTP_NUM_TOKENS` rule to `>= 1`. Pair with `MTP_NUM_TOKENS=5`.
- CPU suite `scripts/test-dspark-block-k.py` (fixture identity, pinned/minimal transform, apply/idempotent, refusals, wiring) plus a `ci-validate` gate; `docs/PATCHES.md`, `docs/ENVS.md`, `.env.dspark.example`, `CHANGELOG.md`.

## Measurements (2× GB10, TP=2, Vision-Exp `86f746b`, same `tools/bench_quick.py`, production idle)

| | k=6 (recipe) | k=5 (this PR) |
|---|---:|---:|
| greedy 8K TTFT / decode | 3.45 s / 55.3 tok/s | 3.45 s / 56.6 |
| greedy 32K TTFT / decode | 13.52 s / 53.6 | 13.54 s / 56.6 |
| temp 0.6 8K decode (3 runs) | 48.6–53.1 | 49.9–56.9 |
| per-position acceptance (greedy) | .89/.73/.49/.34/.23/.15 | .88/.74/.53/.36/.24 |
| mini-bench single stream | 51–55 | 56–64 |
| mini-bench concurrency-4 aggregate | 106–111 | 110 |

The 145-case `scripts/verify-issue136-xgrammar-live.py` gate passed 145/145 twice on the k=5 boot. Per-position acceptance is unchanged, so the gain is the cheaper step; the remaining gap to 0731 (positions 2–4 at 0.53/0.36/0.24 vs 0.66/0.58/0.47) is the Vision-Exp drafter itself. Context: #191 (comment of 2026-09-03).

## Test plan

- `bash scripts/ci-validate.sh` passes locally (includes the new suite and gate).
- Live: `DSPARK_ENABLE_DSPARK_BLOCK_K=1 MTP_NUM_TOKENS=5` boot on 2× GB10: preflight `stock` on both ranks, `dspark-block-k: applied` in both containers, `--speculative-config {"method":"dspark","num_speculative_tokens":5,...}`, 4 DSpark CUDA graphs, capture 40, gate and benches above.

https://claude.ai/code/session_01D4Qky6rbQR7JxqwAPhpFKF


plotarmordev:
Carried into #215 with your commits byte-identical plus an integration merge (stacked on #214 since the two PRs collide in the wiring files). Suites all green. One note from review: the launcher floor with the unlock is k>=1, and repo docs warn k<5 truncates drafts silently — left as you designed, flagged as a maintainer follow-up. You're credited as author on every commit.

PR #220, opened 2026-09-05
Add disk-backed KV cache (kv-disk-tier)

Archmonger:
## Status: source-only publication; maintainer-owned HOLD

Current published head: `76044d33b430631511bed074ff7173001f61c0e4` (a review-fix descendant of `1d2007cb8ca994bd2b083abc99845305935dc226`). This is an experimental, **default-off** disk-backed KV cache, not approval to deploy or merge. `KV_DISK_CACHE_ENABLE=0` and `KV_DISK_CACHE_DIRECT_IO=0` remain the defaults. The tier's mounts and environment delta are included only by the opt-in compose override; existing installations without the enable variable retain the off path.

The implementation is intended to persist each rank's KV shard on its own node and restore evicted prefixes from NVMe. It extends [Doonut's disk-tier work](https://github.com/Doonut/DeepSeek-v4-Flash-0731-DSpark-1M-NVFP4-KV-2x-DGX-Spark/tree/kv-disk-tier); original contribution and authorship are retained. Bind-mounted Python patches and separately built native helpers require qualification against the intended image; no universal vLLM compatibility or multi-node safety claim is made.

### Head `76044d3` (2026-09-17 review-fix)
Addresses the reviewer's three concrete source items from exact-head review `1d2007cb`:
- `dsv4_kv_disk_tier.py`: a block whose unreadable file cannot be retired (`os.remove` raises) no longer stays eligible for lookup. A failed load drops it from LRU/byte accounting and quarantines it, so `lookup()` returns MISS until a fresh store rewrites the file; the un-account is skipped for keys a concurrent store is rewriting.
- `dsv4_shard_tier.py`: completion ACKs are now counted by distinct ZMQ sender identity, so a single/duplicate identity can no longer satisfy the all-nodes `need` and release primary-tier slots before every peer finished its I/O.
- `dsv4_host_kv_alloc.cu`: both `cudaSetDevice` calls are now checked (not just `cudaGetDevice`), failing closed and freeing the pinned block on a failed restore.

Checks run at this head:
- CPU bookkeeping/regression suite `tests/test_dsv4_kv_disk_tier.py`: **32 tests, zero skips, passed** (28 prior + 4 new quarantine/duplicate-sender cases). Note: run against a minimal vLLM/zmq/msgspec CPU stub on the author host, not the pinned image's 2,127-hash-verified tree; the pinned-image exact-head run remains to be repeated by Main.
- Launcher/compose gate `scripts/test-kv-disk-tier-compose-gate.sh`: **6 passed, 0 failed**.
- Full CPU recipe gate `scripts/ci-validate.sh`: **passed** (CPU recipe gates only).
- `py_compile` clean on both changed tier modules.
- **Not tested / still on maintainer hold**: no native CUDA build, GPU memory, actual ZMQ transport, direct I/O, ENOSPC, or two-node serving qualification. `nvcc` was unavailable on the author host, so `dsv4_host_kv_alloc.cu` was reviewed but not compiled. No live boot or two-node run was performed.

### Remaining hold and next-action owner
Maintainer **`plotarmordev`** owns the integration decision, hardware access, maintenance window and restore-to-service plan. **Current canonical main must be integrated and the resulting exact head re-reviewed and requalified before merge.** Before lifting HOLD, the maintainer must select and authorize native build and isolated two-node/direct-I/O byte-exact, ENOSPC, agent restart/reconnect, timeout/slot-reuse, allocator-lifetime and restore-window qualification. Node-wide staging-file cleanup requires an explicit single-tenancy/isolation decision before any shared-node use.

The source review does **not** blanket-resolve all prior conditional limitations: malformed inventory group handling, the 4096-entry worker-key cap, second-timeout I/O fencing, node-wide cleanup and native allocator error/lifetime assumptions remain conditional; these are not reported as observed live failures.

### Historical author observations — not this revision's qualification
Earlier reports described ~150 GB per-node quota, ~35,185 main-group blocks (~36M tokens), and a 4 GiB primary tier of ~503 blocks (~515K tokens), all dependent on the reported geometry. Earlier 2× DGX Spark tests reported unchanged decode/concurrency and cold-prefill results within noise, with experimental direct I/O trading ~8–20% prefill for faster large restores. Reported restores included ~12K tokens at ~1.9 s TTFT versus ~19 s cold and ~750K at ~2.3 s versus 604 s cold. **These are historical, author-reported measurements at earlier revisions/settings, not independently repeated at this head, not guarantees, and not evidence that all performance toggles or current main work together.**


Archmonger:
Marking as draft until I fix some edge cases

plotarmordev:
Thanks for the work here. Persistent prefix reuse is valuable, and I saw your note that you are still fixing edge cases. We had a focused independent review of `ab9294e`; these are the main items to address before re-review:

1. **Make default-off genuinely isolated.** Compose mounts the disk-tier files and `sitecustomize` unconditionally, while `DSV4_ALLOW_EXPANDABLE_SEGMENTS` defaults to 1. The import-hook gates do not check `DSPARK_ENABLE_DISK_TIER`. An off launch should work without a staged `KV_SRC`, leave normal Python/configuration behavior unchanged, and not retain host-KV allocation changes from another experimental knob. [Mounts](https://github.com/MiaAI-Lab/DeepSeek-v4-Flash-DSpark-2x-DGX-Spark/blob/ab9294efd467c4c558da3e85b06295418fdfbda1/docker-compose.dspark.yml#L42-L49), [hook gates](https://github.com/MiaAI-Lab/DeepSeek-v4-Flash-DSpark-2x-DGX-Spark/blob/ab9294efd467c4c558da3e85b06295418fdfbda1/kv-disk-tier/dsv4_sitecustomize.py#L67-L72).

2. **Make both-node setup reproducible.** Provide the clean-checkout staging/build procedure, pinned build and serving image identities, and checks that both nodes receive matching source/library artifacts. `build.sh` currently assumes `/opt/dsv4-kv` is populated and defaults to the optional local Stage-C image; the normal launcher does not stage these files. A short supported procedure is enough—no new deployment framework needed.

3. **Keep compatibility checks narrow and failures visible.** Identify the supported vLLM source/API identity, preserve unrelated checks when exempting the connector, and reject incompatible required patch targets. The test module's catch-all import exception should not turn real import/patch bugs into an entirely skipped suite; skip only genuinely unavailable dependencies.

4. **Fix and test the recovery transitions.** On inspection, [timeout completion](https://github.com/MiaAI-Lab/DeepSeek-v4-Flash-DSpark-2x-DGX-Spark/blob/ab9294efd467c4c558da3e85b06295418fdfbda1/kv-disk-tier/dsv4_shard_tier.py#L1250-L1266) releases a job without first establishing that queued/running peer I/O can no longer access its staging slots. Also, [reconnect reconciliation](https://github.com/MiaAI-Lab/DeepSeek-v4-Flash-DSpark-2x-DGX-Spark/blob/ab9294efd467c4c558da3e85b06295418fdfbda1/kv-disk-tier/dsv4_shard_tier.py#L978-L1007) counts the triggering hello while requesting another from that rank; a late reply can start reconciliation again after READY. Exercise these through the real coordinator/agent path, alongside missing/truncated peer shards and restart/concurrent restore. Required outcomes: no partial-shard hit, no staging-slot reuse while old I/O can touch it, and stable eventual readiness. Focused behavioral tests, not a large mock framework.

5. **Make performance evidence reproducible.** Compare off/on cold prefill, completed store, and restore with identical requests/model/settings. Include correctness checks and per-node byte/completion evidence distinguishing NVMe restore from GPU/CPU prefix hits. State eviction/restart procedure, quota and retained prefix, filesystem/JIT cache state, and reconcile large-context restore claims with the documented default capacity of approximately 562K main-group tokens.

The goal is a small, intentional implementation with clear ownership of memory and failure transitions—not more knobs or abstractions. This was static review, not GPU qualification or an exhaustive CUDA audit.

Please let us know once these are complete and the PR is ready for re-review, with the reproduction commands/results. We can then evaluate it against the current recipe.

Archmonger:
Received! Thank you for your quick review turnaround.

Right now I am focused on fleshing out the experimental `DSV4_HOST_KV` feature into something that has hopes of being production grade.

I will re-label this PR as draft until I handle both `DSV4_HOST_KV` and your review comments.

Archmonger:
> ... reconcile large-context restore claims with the documented default capacity of approximately 562K main-group tokens.

Small note, my documentation regarding default capacity of 562K tokens is now stale. I [fixed a bug](https://github.com/MiaAI-Lab/DeepSeek-v4-Flash-DSpark-2x-DGX-Spark/pull/220/changes/8dd071b340f8fa5848cad499db6c5ad96464d125) in the mem to disk conversion. Quick estimate is it can now fit ~24 million tokens in the default 150gb of disk space, but I still have a TODO item to verify the exact quantity.

Archmonger:
`DSV4_HOST_KV` is now feature complete. I am keeping it labeled as experimental while I continue to stress test, but in theory, it is a strong option.

Regarding the review — all five items are addressed and pushed. Ready for re-review.

## Review Action Summary

**1. Default-off isolation.** The master switch is now `KV_DISK_CACHE_ENABLE` (default off). The disk-tier volumes and `sitecustomize` moved to `docker-compose.dspark-disk-tier.override.yml`, which the launcher adds only when the flag is on, and every import hook in `dsv4_sitecustomize.py` is gated on that flag — so a `KV_DISK_CACHE_HOST_KV=1` left behind by another experiment can no longer leak host-KV allocation changes into an off launch. `KV_DISK_CACHE_ALLOW_EXPANDABLE_SEGMENTS` is gone entirely: with the tier on, `PYTORCH_CUDA_ALLOC_CONF=expandable_segments:True` is kept implicitly and the `OffloadingConnector` is exempted unconditionally. An off launch now needs no staged `KV_DISK_CACHE_SRC` and leaves Python/config untouched.

**2. Reproducible both-node setup.** Added a "Stage on both nodes" section to `kv-disk-tier/README.md`: a clean-checkout `scp` / `build.sh` / `md5sum` procedure, pinned build image (`vllm-dspark-runtime:dspark-nvfp4-stage-c`), serving image (`ghcr.io/anemll/dspark-vllm-gx10:0.1.1@sha256:a8394849…`), model commit (`6821d6ad…`), and a startup preflight that fails loudly on any drifted patch target.

**3. Narrow compat checks, visible failures.** The `sitecustomize` exemption now runs the stock `_verify_kv_transfer_compat` first and swallows only the specific `expandable_segments` rejection (matched on connector + message). The test module skips only genuinely unavailable third-party deps (`vllm` / `zmq` / `msgspec`); our own import/API bugs re-raise and fail the suite instead of silently skipping.

**4. Recovery transitions fixed + tested.** Timeout completion is now two-stage: the first timeout abandons the result but *holds* the job one extra window so queued/running peer I/O can drain before its staging slots are reused; only a second timeout force-fails. Reconnect reconciliation now uses a monotonic hello epoch — agents echo it, stale/late hellos are ignored, and the triggering hello is no longer double-counted, so a late reply can't restart reconciliation after READY. Added focused tests (timeout hold-then-fail, partial-shard drop, reconnect/stale-hello); `tests/test_dsv4_kv_disk_tier.py` runs 10/10 in the serving image.

**5. Reproducible perf evidence.** Added a "Performance evidence" section to the README with the exact command (`scripts/benchmark-0731.py`, fixed lengths/concurrency/settings). Disk tier off vs on: decode and concurrency unchanged, cold prefill within noise of off; the experimental host-KV path (`KV_DISK_CACHE_HOST_KV=1`) costs ~8–20% prefill with decode unchanged. Restore evidence: a 12k-token prefix reloads from NVMe in ~1.9 s TTFT vs ~19 s cold fill; a ~750K-token fill restores in ~2.3 s vs 604 s cold fill. The capacity mismatch you flagged is also corrected — the 64× aliased-view over-count is fixed, so the default 150 GB quota now caches ~35,185 blocks ≈ 36M main-group tokens (not ~562K), reconciling the large-context restore claims.

**Reproduction**
- `bash scripts/ci-validate.sh` — green.
- `tests/test_dsv4_kv_disk_tier.py` in the pinned serving image — 10/10 pass.
- Booted the two-node TP=2 lane off and on: `primary=lru (503 blocks), 1 secondary tier`, 2/2 shard agents, `/v1/models` and `/v1/chat/completions` 200.
- Benchmarks and restore numbers: `kv-disk-tier/README.md#performance-evidence`.

plotarmordev:
**Maintainer status: CHANGES REQUESTED / HOLD at the current head.**

Reviewed head: `15f1ab22642292880cf86b27315db14fa700bffe`

Thank you for the detailed 2026-09-06 action summary on the earlier five review items (from the `ab9294e` review). This is a fresh static pass at the current head; it found two **new** default-off-isolation defects that are distinct from the earlier items:

1. **Ungated env additions in the base compose.** The base `docker-compose.dspark.yml` now sets `VLLM_SKIP_INIT_MEMORY_CHECK:-1` and `VLLM_ENGINE_READY_TIMEOUT_S:-3600` without gating them on the disk-tier flag, so they apply to flag-off launches too. The repo's own `docs/ENVS.md` lists `VLLM_SKIP_INIT_MEMORY_CHECK` as "Not in Anemll 0.1.1 registry — avoid on Anemll". This contradicts the "default-off is genuinely isolated" property from the earlier review.
2. **`COMPOSE_FILE` override regression.** `compose_base()` drops the `$COMPOSE_FILE` argument (absolute, env-overridable) in favor of a hardcoded relative `-f docker-compose.dspark.yml`, which silently breaks `COMPOSE_FILE` overrides and makes the result depend on the caller's working directory.

**What static review established.** The integration surface (compose mounts, sitecustomize hooks, stop-side `/dev/shm` cleanup) was reviewed; the author's measured restore/latency numbers and the 10/10 in-image test result remain author-measured, not independently verified, and the capacity math and HOST_KV path are self-flagged as unverified/experimental. The CUDA and shard-protocol internals have had only a partial review.

**Still needed.** Fix the two items above so flag-off launches are byte-identical in behavior to before; then deeper runtime review of the CUDA/shard internals; then live validation (flag-off boot, flag-on boot, and restore measurement against identical requests/settings).

**What would unblock.** The two isolation fixes, a maintainer re-review of the runtime internals, and the live off/on validation evidence.

plotarmordev (commented):
Source-only COMMENT at `5e3d11366d468fbd48f5dcb6079aaca542baa59a`; publication permitted under the existing maintainer-owned HOLD, not a merge/deployment approval.

The concrete D1–D5 corrections are addressed in source: actual ndarray specs and slot zero are handled without truth tests, optional dependency skipping is exact top-level absence only, completion pin retirement follows the real pinned hook, installed import errors propagate through the existing availability helper (including both remaining direct-map/skip sites), and explicit stale gen=0 is distinct from a spontaneous hello. The inventory fixture now uses the pinned FileMapper, and duplicate-rank coverage checks readiness/lookup rather than log wording. Main supplies 28 successful CPU bookkeeping tests with zero skips and six successful compose cases at this head; I read those receipts without rerunning validation.

Residual conditional concerns are retained, not blanket-approved:
- I2/I16: `dsv4_kv_disk_tier.py:140-160,330-358,388-395,472-494`: failed unlink can leave a lookup-visible file; a buffered transient error can retire it; present replacements survive failed-store reconciliation, but in-flight writers/eviction and cap accounting are not fully qualified. fs_capped is not the shipped shard_dist selection.
- I9: `dsv4_vllm_patches.py:1248-1269`: both config shapes are supported, but detection exceptions warn then delegate; not a universal fail-closed guard.
- I11: `dsv4_shard_tier.py:670-717`: real mapper round-trip is covered; invalid out-of-range directory group values can still fail before the mapper try. The pinned mapper's own generated group suffix comes from four key bytes; malformed external inventory is not qualified.
- I15: `dsv4_vllm_patches.py:1811-1831,2010-2014`: slot maps use a bounded slot namespace; the 4096 worker-key cap does not prove oldest entries are stale. Pinned worker code defers stores until the next step; backlog beyond the cap remains unqualified, not a proven ordinary-run failure.
- I17: `dsv4_shard_tier.py:738,1131-1155,1265-1273`: normal agents derive identity from rank; synthetic duplicate identities cannot satisfy the initial rank barrier alone, but ACKs are counted without identity deduplication. The test is not transport/reconnect safety proof.
- I18: `gpu_clear.sh:21-28,36-47`: unknown GPU state rejects, but node-wide mmap deletion remains unsafe for sibling instances; documentation is not enforcement.
- I22: `dsv4_shard_tier.py:1474-1504`: grace-window result delivery is correctly described; the second timeout does not fence outstanding I/O before slot release.
- I24: `dsv4_host_kv_alloc.cu:35-51,70-80`: cudaGetDevice failure is checked; cudaSetDevice outcomes and asynchronous free/shutdown lifetime remain native qualification concerns.

Hold owner: **plotarmordev**. Unblock requires a chosen integration/maintenance/isolation direction, current canonical-main integration and exact-head checks, authorized native build plus two-node/direct-I/O byte-exact and failure/reconnect/allocator/restore-window qualification, and resolution or explicit supported-contract disposition of the conditional concerns above. No immediate author benchmark/rebase demand, no force/rebase, and no merge. Historical performance observations in the updated PR body do not qualify this revision.

plotarmordev (changes requested):
## Hold update

Public head checked: `1d2007cb8ca994bd2b083abc99845305935dc226`.

**What passed:** The default-off Compose render contained no disk-tier environment additions or disk mount. Several earlier source defects are fixed; I am not asking you to redo those fixes. We did not run the enabled two-node disk tier in this maintenance window.

**What still needs work in the current source:**
- In `dsv4_kv_disk_tier.py`, a block can remain eligible for lookup if retiring an unreadable file fails. Failed retirement must not cause repeated attempts to promote the same unusable block.
- In `dsv4_shard_tier.py`, completion ACKs are counted without checking distinct sender identity. Completion must mean the required distinct participants have completed.
- In `dsv4_host_kv_alloc.cu`, check the `cudaSetDevice` results as well as `cudaGetDevice`.

These are source findings, not newly reproduced live failures. The earlier reviewer also retained concerns about malformed inventory, in-flight store/eviction accounting, slot/job lifetime, timeout fencing, and cleanup affecting sibling instances. They are not cleared by the default-off render or by a successful ordinary boot.

**Next action — author:** Address the concrete source items above, with focused behavioral checks where possible. **Next action — me, @plotarmordev:** Decide the supported isolation/cleanup contract and disposition of the remaining concerns before requesting broad hardware work. I also own the pending launcher integration with #249.

**Before ready:** Review the corrected exact head, then agree the enabled test plan. It needs observed store/restore behavior and correct data on both ranks, plus the relevant error/recovery checks. Verify effective settings on both nodes; do not assume ambient shell settings reach the worker. No immediate benchmark or shared-machine run is requested.

Archmonger:
Addressed the three concrete source items from the exact-head review `1d2007cb` in new head `76044d3`:

- **`dsv4_kv_disk_tier.py`**: a block whose unreadable file cannot be retired (`os.remove` raises) no longer stays eligible for lookup. A failed load now drops it from LRU/byte accounting and quarantines it so `lookup()` returns MISS until a fresh store rewrites the file; the un-account is skipped for keys a concurrent store is rewriting.
- **`dsv4_shard_tier.py`**: completion ACKs are counted by distinct ZMQ sender identity, so a single/duplicate identity can no longer satisfy the all-nodes `need` and release primary-tier slots before every peer finished its I/O.
- **`dsv4_host_kv_alloc.cu`**: both `cudaSetDevice` calls are now checked (not just `cudaGetDevice`), failing closed and freeing the pinned block on a failed restore.

Checks run at this head: the CPU test suite `tests/test_dsv4_kv_disk_tier.py` ran **32 tests, zero skips, passed** (28 prior + 4 new cases); `scripts/test-kv-disk-tier-compose-gate.sh` **6 passed, 0 failed**; full `scripts/ci-validate.sh` **passed** (CPU recipe gates). `py_compile` clean. Not tested: `nvcc` was unavailable on the author host, so `dsv4_host_kv_alloc.cu` was reviewed but not compiled; no native build, GPU, transport, direct-I/O, ENOSPC, or two-node run was performed. The CI `validate` workflow run for this fork commit requires maintainer approval before it executes.

Archmonger:
I am going to leave the PR description in the format that you modified it to, updated to include my changes.

Note that this PR description does not conform to this repo's PR template, but I'm assuming you (or your LLM) decided to deviate. If this was unintentional, then you may want to add a memory/agents.md in your harness to always use PR templates (and/or github issue templates) when they are available.

PR #223, opened 2026-09-05
fix(dspark): recompute last 128-token draft window on prefix-cache hit (Anemll #2 backport, opt-in)

plotarmordev:
## What

Backports Anemll/dspark-vllm-gx10 PR#2 (`4afc5e7eeb`): on a prefix-cache hit, the DSpark draft's SWA cache now always recomputes the last 128-token draft window instead of trusting the stale prefix — fixing degenerate/truncated output on repeated identical prompts. Our lane (prefix caching ON + DSpark draft) had no `dspark_window_size` coverage anywhere (verified in tree and pinned image).

Opt-in `DSPARK_ENABLE_DSPARK_SWA_PREFIX` (default `0`), house pattern: upstream-byte-exact regions + `[dspark-swa-prefix]` marks, `kv_cache_manager.py` whole-file pins (stock `be9c5091` → patched `09f0e990`), `scheduler.py` region-count discipline with pure-transform pins (`e25d4c9a` → `69fc8118`), both targets preflighted before either atomic replace, fixtures from the pinned image, wiring mirrored on #208.

## Verification

- 16/16 hermetic tests (rc=0); `scripts/ci-validate.sh` rc=0 (includes the #221 passthrough guards); CLI apply smoke rc=0
- Pinned-image boot-chain proof on a live host: grammar-advance + empty-encoder + issue27 then this patcher — apply / idempotent / status / import, rc=0, live serve untouched

Repeated-prompt output-quality A/B in one boot window shared with the RoPE gate (per-position acceptance + 128K quality). Nothing changes until then (flag stays 0).


plotarmordev:
Live gate complete on the two-rank lane (flag ON, shared window with the RoPE gate, 2026-09-05):

- **Repeated-prompt output quality: PASS.** 4× identical ~1.2K-token prompt: first cold (cached=0), repeats hit `cached_tokens=1024` (4×256-token blocks) — every completion coherent and complete; no degenerate or truncated output on repeated identical prompts with the prefix cache engaged.
- Safety: restarts 0 on both ranks, warmup 35/35, cap=1 confirmed, standard boot-shape battery green.

Same window also measured per-position draft acceptance with the fix on (prose/code/agent, C1/C4): no regression vs the flag-off baseline within the measured noise band; the pos3 acceptance cliff (0.16–0.32) is unchanged by this patch (it's drafter-intrinsic, unrelated to the SWA-prefix path).

Nothing else gates the merge.

PR #225, opened 2026-09-05
fix(dsml): orphan-invoke recovery for tool-call content leakage (vLLM #52645 backport, opt-in)

plotarmordev:
## What

Ports open vLLM #52645 (DSML orphan-invoke recovery) onto the pinned `g752a3a504` parser engine. Verified absent from the pinned image (recovery-symbol grep rc=1): foreign/orphan tool invocations leak raw DSML markup into response content — a tool-call content-corruption class on our exact stack.

Opt-in `DSPARK_ENABLE_DSML_RECOVERY` (default `0`), house pattern: 6 sole-owned parser files, 21 source-exact regions, whole-file stock+patched sha256+size pins, preflight-all-then-atomic-per-file publish with rollback, idempotency marks, fail-closed `--check`/`--status`.

## Verification

- `scripts/test-dsml-recovery.py`: **37/37** (rc=0) — pins, patcher fail-closed/idempotency/rollback, the upstream 16 engine + 5 delegating regression scenarios replayed on pinned fixtures, 16-case stock/patched normal-path parity, wiring
- `scripts/ci-validate.sh`: rc=0
- Live docker smoke on the pinned image's real stack: **stock leak reproduced** (`tools_called=False`, 154-char DSML in content), then apply ×6, recovery + rollback + foreign-wrapper + normal-path + delegating flows all pass, re-apply idempotent

Agent tool-call acceptance parity vs the measured 42.3% C1 baseline, plus the stock-leak repro and the 145-case gate. Flag stays 0 until then.


plotarmordev:
Live gate complete on the two-rank lane (flag ON, 2026-09-05):

- **Apply + identity**: all 6 parser files `patched`, restarts 0, warmup 35/35.
- **Acceptance parity**: agent C1 43.5% overall (baseline 42.3%, +1.2 inside the measured noise band); all six cells (prose/code/agent, C1/C4) flat within noise — no draft/throughput regression from the patch.
- **Foreign-invoke probe**: induced orphan tool invoke under `tool_choice=auto` — clean structured call, zero DSML markup in content.

Nothing else gates the merge.

Issue #144, opened 2026-08-26
reasoning_effort: high/max breaks prefix caching (0% hit) — low/off cache normally


d3vilbug:
## Summary
`chat_template_kwargs.reasoning_effort` set to `high` or `max` causes the prefix cache to never hit — even for a byte-identical, repeated system-prompt prefix across sequential requests. `low` and no-thinking both cache normally (~95% hit rate). Reproduced repeatedly across multiple days/server restarts.

## Environment
- Model: `deepseek-ai/DeepSeek-V4-Flash-0731`, TP=2 on 2x DGX Spark (GB10)
- Image: `ghcr.io/anemll/dspark-vllm-gx10:0.1.1`
- `--enable-prefix-caching --enable-chunked-prefill --kv-cache-dtype nvfp4_ds_mla`

## Repro
6+ sequential chat-completion requests, identical 6000-token system-prompt prefix ("shared") vs a prefix that differs at the start each call ("unique", miss control). Checked `usage.prompt_tokens_details.cached_tokens` per response, varying only `chat_template_kwargs.reasoning_effort` (off / low / high / max).

## Results (reconfirmed multiple times, incl. after full server restarts)

| reasoning_effort | cache hit rate | hit latency | miss latency | speedup |
|---|---|---|---|---|
| off | 95% | ~1.4-1.7s | ~3.8-4.0s | ~2.3-2.7x |
| low | 95% | ~1.6-1.7s | ~3.9-4.0s | ~2.3-2.5x |
| high | 0% | ~4.2s | ~4.1s | ~1.0x |
| max | 0% | ~4.1-4.3s | ~4.1-4.3s | ~1.0x |

At `high`/`max`, `cached_tokens=0` on *every* request including requests using the exact same system prompt as request 1 — rules out ordinary cold-start.

## Investigation (ruled out, in case it saves time)
- **Prompt content is deterministic**: called `encode_messages()` directly twice with identical `max`-effort input — byte-identical SHA256 both times. Not a template non-determinism bug.
- **`cache_salt`**: unset in all test requests; no DSpark patch sets it based on reasoning_effort.
- **`need_extra_keys()`** (`v1/core/kv_cache_utils.py`): only checks `mm_features`/`lora_request`/`cache_salt`, none apply here.
- **Model Runner V2 `thinking_token_budget` check** (`config/vllm.py`): unrelated feature/parameter, not sent in these requests.
[...]

PR #227, opened 2026-09-05
fix(encoder): align effort directive after shared prefix for cross-effort cache sharing (issue #144, opt-in)

plotarmordev:
## What

Fixes the effort-directive prefix-cache partition (issue #144's real mechanism, diagnosed on this lane): the reasoning-effort directive is front-inserted and **not 256-aligned** (max/xhigh=93 tok, high/DEFAULT≈79, low/off/medium=0), so it shifts every 256-token cache-block boundary and partitions the prefix cache into effort buckets — cross-bucket hits are exactly 0, intra-bucket ~96–98%.

**Chosen fix:** relocate the effort directive to the **end of the leading system-message run**. The shared system prefix then hashes identically across effort levels; only the suffix (directive + user turns) differs. Padding was rejected (chained block hashes + KV content-dependence make pad tokens incapable of restoring sharing); per-effort cache warming was rejected (3× KV at ~97% memory pressure).

**Proof (live tokenizer + real-BPE CI tier):** cross-bucket shared full blocks **0 → 18/18** on a 4646-token agent prompt; junction cost 1 token; low/off/medium rendering **byte-identical to stock**.

Opt-in `DSPARK_ENABLE_ISSUE144_EFFORT_ALIGN` (default `0`), house pattern: both encoder pre-states sha-pinned + fixtured, applies after the issue21/vision-exp/assistant-final co-patchers, atomic, idempotent, `--check`/`--status` fail-closed. Full wiring (compose/launcher/env/docs/ci-validate locks).

## Verification

- `scripts/test-issue144-effort-align.py`: **21/21** (rc=0)
- `scripts/ci-validate.sh`: rc=0 (incl. new gate + passthrough locks)
- Real-snapshot `--check` on a live host: rc=0, sha unchanged

Cross-bucket cache-hit measurement (high/max after low-warmed prefix must hit >0, targeting ~95%) plus output-quality parity at high/max vs stock rendering. Flag stays 0 until then.

Refs #144.


plotarmordev:
Live gate complete on the two-rank lane (flag ON, 2026-09-05):

- **Cross-bucket cache sharing restored**: shared ~1.2K-token prefix warmed at `low`, then queried cross-effort — `low→high` cached_tokens=1024 and `high→max` cached_tokens=1024 (4×256 blocks). Stock gives **0** on both transitions; intra-bucket controls also 1024 as expected.
- **Termination intact**: `reasoning_effort=max` on a simple task finishes `stop` with correct answer and reasoning present (no rambling). A deliberately hard prompt that fails to terminate at max also fails identically at `low` — and low rendering is byte-identical to stock, so that behavior is a model/task artifact, not a fix regression.
- Safety: restarts 0, warmup 35/35.

Nothing else gates the merge.

PR #238, opened 2026-09-06
fix(pi): ship "input": ["text", "image"] so pi forwards screenshots to the vision model

Vincebye:
## Problem

`pi-models.dspark.example.json` declared the `deepseek-v4-flash-vision-exp` model entry with:

```json
"input": ["text"]
```

Pi keys modal image support off a model's `input` array (the default is `["text"]`; a model is only allowed to receive images when the entry explicitly includes `"image"`). Because this is a **vision** model but the shipped example was text-only, anyone who followed the README and copied the example into `~/.pi/agent/models.json` got this on every image:

```
Current model does not support images.
The image will be omitted from this request.
```

The image was dropped **before the HTTP request was sent** — confusing, because the endpoint itself accepts OpenAI `image_url` / base64 data-URL image parts (and `.env.dspark.example` / the README already advertise image support). This is a pure client-config bug, not a server bug: the failure happens in Pi's local request builder, before the request ever reaches `/v1/chat/completions`.

## Fix

- `pi-models.dspark.example.json`: change `"input": ["text"]` → `"input": ["text", "image"]` so a fresh copy lets Pi forward screenshots as base64 `image_url` parts.
- `README.md`: add a "pi image support" note in the pi section that documents the requirement, the exact symptom, the fix for users who already copied an older entry (add the `"image"` value and start a **new** pi session — a `/reload` inside the same session is not enough), and the server-side rule that `image_url` belongs only on `user` turns.
- `CHANGELOG.md`: record the fix.

## Notes

- No server-side change is needed. The Vision-Exp endpoint already accepts OpenAI-compatible `image_url` content parts, and the project's own vision hotfix consumes them.
- This mirrors the upstream Pi guidance (Pi's `models.md` and the existing OpenAI-compatible VLM issue) that vision-capable custom models must declare `"input": ["text", "image"]`.


plotarmordev:
**Maintainer merge preparation — reviewed head `6c4368fc2b30ad1709af136412615ed186d2b7ef`.**

**Why merge:** Correct the pi example so the vision model declares image input capability.

**Changes and verification:** The parsed JSON contract and byte preservation were checked: only the intended input capability changes. README reload guidance is grounded in pinned public pi source/documentation rather than an unsupported mandatory full-restart claim. The full `scripts/ci-validate.sh` CPU recipe gate passed at this exact head after integrating current main. Worktree and exact-range secret scans passed with a documented existing synthetic-canary exception. Independent HELP code review and Main execution evidence are retained.

**Limits:** This changes the shipped example, not already-installed client configurations. Actual screenshot delivery and deployed vision correctness were not exercised.

Contributor commit ancestry is preserved in this fast-forward maintainer update. Hosted [validate run 34475643892](https://github.com/MiaAI-Lab/DeepSeek-v4-Flash-DSpark-2x-DGX-Spark/actions/runs/34475643892) passed on this exact head. Merging the scoped change. No serving deployment change is included. Thank you for the contribution.

PR #240, opened 2026-09-07
fix(stop): anchor docker name filters to the project's own containers

shiaho777:
## Description and Motivation

`docker --filter name=` takes an **unanchored** regular expression (moby `pkg/filters` uses `regexp.MatchString`). The stop script passed plain patterns like `name=${project}-vllm-dspark` to `docker rm -f` at **seven** sites, so a routine `./stop-deepseek-v4-flash-dspark.sh` would force-remove **any** container whose name merely contains that substring — e.g. a neighbouring checkout's `deepseek-v4-flash-vllm-dspark-old` — and regex metacharacters in `LEGACY_PROJECT_NAME` (derived from the checkout directory name, dots included) were interpreted as regex. The start script's own already-running check (`start-deepseek-v4-flash-dspark.sh` line ~1262) correctly uses `grep -qx` exact matching; the stop side did not.

What changes:

* New `project_name_filters()` helper in `stop-deepseek-v4-flash-dspark.sh`: escapes regex metacharacters in the project name and anchors the pattern to the compose container-name shape `<project>[-_]<service>([-_]<index>)?` (v2 `-` separator, v1 `_`, optional numeric index, so compose-named and bare manual names still match).
* All seven `--filter name=` call sites (force-remove sweep, sidecar sweep head/worker, main head, main worker) consume the anchored `RANK_NAME_RE` / `SIDECAR_NAME_RE`.
* Label-based filtering (`com.docker.compose.project=`) is unchanged.

No knob, default, or serving behavior changes; this only narrows which containers a stop may select. Safe on both ranks: the remote paths receive the same pre-computed anchored regex via the existing heredoc/ssh transport (covered by an end-to-end test below).

## Related Issues

None filed.

---

## Testing

* New CPU suite `scripts/test-stop-name-filter.py` (11 tests, wired into `ci-validate.sh`), modelled on `scripts/test-env-normalisation.py`:
  * extracts the shipped `project_name_filters()` from the stop script and evaluates its regexes against a match/reject matrix with Python `re` (equivalent to Go's RE2 for this construct subset): own containers (`-1`, `-12`, v1 `_1`, bare) match; suffix overmatch (`-old`, `-1-backup`, `-1a`, `2-1`, `x-1`), prefix overmatch (`my-…`), and cross-service names are rejected;
  * metachar project names (`deepseek.v4.foo`, `ab+c[d]`) match literally and do not act as wildcards;
  * end-to-end heredoc assembly: runs the real `force_rm_project_containers` local path against a fake `docker` binary and asserts the filters arrive anchored and unmangled (`name=^deepseek-v4-flash[-_]vllm-dspark([-_][0-9]+)?$`), including escaped metachar projects;
  * source-shape guards: no unanchored `name=${project}-` remains in code; all call sites use the anchored variables.
* `bash -n stop-deepseek-v4-flash-dspark.sh` clean.

Not run: live two-node stop on the DGX Spark pair (no hardware here); the change is filter-string-only and the assembled remote command is covered by the fake-docker test above.

---

## Checklist:

- [x] New runtime behavior is opt-in and default-off; defaults are unchanged (this PR changes no knob or default — it narrows stop-time container selection).

Note: this PR is one of a batch of small script-bug fixes; the CHANGELOG entries sit at the same file position by convention, so the last ones to merge may need a trivial rebase — happy to do that on request.


plotarmordev:
**Maintainer merge preparation — reviewed head `3ecf60f74017871a95df19cd0c8d8e4c8828115c`.**

We retained the anchored container-name fix and added a narrow quoting repair so the literal regex/project values survive the local and worker shell boundaries. Exact Compose-label ownership remains unchanged. Your original commit and authorship are preserved; this is a fast-forward maintainer update, not rewritten history.

**Why merge:** stopping one deployment should not select a neighbouring container merely because its name contains the same substring.

**Verification:** the full `scripts/ci-validate.sh` CPU recipe gate passed at this head. The full-script recorder regression failed on frozen main with incorrect selections and on the submitted PR's quoting cases; all four regression tests pass on this revision. Worktree and exact commit-range secret scans passed after reviewing one existing synthetic-canary false positive.

**Limits:** Docker/SSH/removal operations were isolated recorders; no serving containers were stopped. This validates command construction and selection behavior, not a live Docker/Compose shutdown or every NFS path. Hosted [validate run 34473280381](https://github.com/MiaAI-Lab/DeepSeek-v4-Flash-DSpark-2x-DGX-Spark/actions/runs/34473280381) also passed on this exact head. Merging the scoped fix; no deployment update is part of this merge.

Thank you for the contribution.

PR #241, opened 2026-09-07
fix(bench-patches): generate TTFT prompts with integer arithmetic

shiaho777:
## Description and Motivation

`scripts/bench-patches.sh`'s short-context TTFT section (the benchmark's primary output for judging hotfix #49486) never measured what it claimed:

```bash
prompt=$(python3 -c "print('hello ' * ($prompt_tokens * 4 / 3 // 6))" 2>/dev/null || echo "hello world")
```

In Python 3, `/` yields a float (`256*4/3//6` = `56.0`) and `'hello ' * 56.0` raises `TypeError: can't multiply sequence by non-int of type 'float'` (reproduced). The `|| echo "hello world"` fallback then fired **silently for every labelled size** — `prompt=256/512/1024/2048` (and 4096/8192 in `--full`) all sent the identical 2-token prompt. Any TTFT-versus-prompt-length conclusion drawn from this table is void and should be re-measured after this fix.

The fix is the minimal one: integer arithmetic only (`$prompt_tokens * 4 // 3 // 6`), plus a comment recording why. The actual served token count is still printed per row from the server response (`usage.prompt_tokens`), so operators can keep verifying prompt sizes against the log.

No knob, default, or serving behavior changes — benchmark script only.

## Related Issues

None filed.

---

## Testing

* New CPU suite `scripts/test-bench-patches-prompt.py` (wired into `ci-validate.sh`):
  * pins the shipped expression to integer arithmetic (no single `/`);
  * extracts the expression from the script and evaluates it for every bench size (256…8192), asserting a real repeated-word prompt comes out and the fallback string is never what gets sent;
  * asserts all six labelled sizes produce *distinct* prompt lengths so the TTFT table cannot silently collapse rows again.
* `scripts/bench-patches.sh` is added to the `bash -n` gate in `ci-validate.sh` — it was one of the shell scripts with no syntax gating at all.
* `bash -n scripts/bench-patches.sh` clean.

Not run: the live benchmark itself (needs the two-node DGX Spark lane; no hardware here).

---

## Checklist:


plotarmordev:
**Maintainer merge preparation — reviewed head `c1bfd7f58966b50ff47fb4cf61c54549cd15d781`.**

**Why merge:** Restore the advertised size-dependent benchmark prompts instead of silently using the tiny fallback.

**Changes and verification:** The final host regression executes the shipped prompt assignment across all six declared sizes. Frozen main produces the fallback and fails; integer arithmetic fixes it. No duplicate arithmetic oracle or source-string assertion is used. The full `scripts/ci-validate.sh` CPU recipe gate passed at this exact head after integrating current main. Worktree and exact-range secret scans passed with a documented existing synthetic-canary exception. Independent HELP code review and Main execution evidence are retained.

**Limits:** Prompt-size labels remain approximate, not tokenizer measurements. Historical measurements from the faulty path are not repaired by this change and need new separately authorized collection.

Contributor commit ancestry is preserved in this fast-forward maintainer update. Hosted [validate run 34475221804](https://github.com/MiaAI-Lab/DeepSeek-v4-Flash-DSpark-2x-DGX-Spark/actions/runs/34475221804) passed on this exact head. Merging the scoped change. No serving deployment change is included. Thank you for the contribution.

PR #242, opened 2026-09-07
fix(bench-baseline-issue22-only): exec the issue22 hotfix where it actually exists

shiaho777:
## Description and Motivation

`scripts/bench-baseline-issue22-only.sh` Step 3 could never succeed:

```bash
docker exec deepseek-v4-flash-vllm-dspark-1 bash /tmp/hotfix-nvfp4-ds-mla-issue22.sh
```

Compose mounts the repo's `patches/` directory read-only at `/opt/dspark-patches` (`docker-compose.dspark.yml:108`); nothing exists at `/tmp/hotfix-*` in the container. Under `set -euo pipefail` the script aborted at Step 3 on every run — the TTFT benchmark never ran, and the cluster was left in the `DSPARK_SKIP_HOTFIX=1` baseline state with no automatic restart of the patched stack.

The fix points the `docker exec` at the compose-mounted path `/opt/dspark-patches/hotfix-nvfp4-ds-mla-issue22.sh` (the same path the boot command uses at compose line 436). Re-application is safe: the hotfix header declares "Safe to re-run (idempotent — skips already-applied patches)", and `DSPARK_SKIP_HOTFIX=1` does **not** gate issue22 (only `DSPARK_SKIP_ISSUE22_HOTFIX` does, per compose line 436), so the Step-2 boot has already applied it and Step 3 becomes an explicit ensure plus state check. A comment in the script now records both facts so the next reader doesn't "fix" it back.

No knob, default, or serving behavior changes — benchmark script only.

## Related Issues

None filed.

---

## Testing

* `bash scripts/ci-validate.sh` — passes (no new failures).
* New CPU suite `scripts/test-bench-baseline-hotfix-paths.py` (wired into `ci-validate.sh`):
  * bans `/tmp/hotfix-` references in both bench-baseline scripts;
  * extracts every `docker exec … bash /…` in-container hotfix path from both scripts and asserts each maps to a real file under `patches/`;
  * pins the compose mount contract (`:/opt/dspark-patches:ro`) so a future mount move breaks CI, not the benchmark;
  * pins the hotfix's idempotency promise in its header, which Step 3 relies on.
* `scripts/bench-baseline-issue22-only.sh` added to the `bash -n` gate (it was ungated); `bash -n` clean.

Not run: the live two-node baseline (needs the DGX Spark pair; no hardware here).

---

## Checklist:

- [x] I have read the README and `docs/ENVS.md` and kept my changes consistent with them.
- [x] My pull request has a sound title and description.
[...]

PR #243, opened 2026-09-07
fix(bench-baseline): tear down both nodes; drop backgrounded-launcher PID handling

shiaho777:
## Description and Motivation

Both `scripts/bench-baseline-issue22-only.sh` and `scripts/bench-baseline-no-patches.sh` manage a full restart of the two-node lane, but their container lifecycle was broken in three ways:

1. **Head-only teardown.** Each script ran a bare `docker compose -p deepseek-v4-flash … down`, which bypasses `./stop-deepseek-v4-flash-dspark.sh` and leaves the **worker rank serving**. The following start then fails the launcher's worker stale-rank precheck (`start-deepseek-v4-flash-dspark.sh` line ~1273).
2. **Failure invisible for 10 minutes.** The start script was launched in the **background** (`bash start-… &`) and its exit code was never checked; the scripts then sat in a 60×10s curl loop waiting for an API that could never come, and proceeded toward the benchmark anyway.
3. **PID-reuse hazard.** `kill $START_PID` fires at the final step — minutes after the launcher has already exited (it blocks until the API is up, then exits 0). By then the PID may have been recycled to an unrelated user process.

The launcher already blocks until the API is ready (or exits non-zero), so the fix simplifies rather than adds: **tear down through `./stop-deepseek-v4-flash-dspark.sh`** (head + worker, exits non-zero if a node is unreachable, so the bench fails fast instead of starting a half-up cluster) and **run the start script in the foreground**, so a failed start aborts the benchmark immediately under `set -euo pipefail`. All `START_PID` bookkeeping, `kill` calls, and redundant local wait loops are removed.

No knob, default, or serving behavior changes — benchmark scripts only.

## Related Issues

None filed.

---

## Testing

* New CPU suite `scripts/test-bench-baseline-lifecycle.py` (wired into `ci-validate.sh`) pins the corrected structure: no bare `docker compose down`, teardown via the stop script before every start, launcher never backgrounded, no `START_PID`/`kill`, no local wait loop. (Assertions run on code lines only, so the explanatory comments don't trip the guards.)
* `bash -n` clean on both scripts; `bench-baseline-no-patches.sh` added to the `bash -n` CI gate (it was ungated).

Not run: the live two-node baseline (needs the DGX Spark pair; no hardware here).

---

## Checklist:


plotarmordev:

Reviewed head: `470bf4ab72d5323af5ad584ebd90697459ffd917`

**What static review established.** All three motivating defects were confirmed on current main in the two bench-baseline scripts: teardown used a bare head-only `docker compose … down` (leaking the worker rank), the launcher was backgrounded with its return code unchecked, and `kill $START_PID` carried a PID-reuse hazard. The launcher does block until the API is ready, and the stop script covers both nodes and counts failures, so running the launcher foreground and tearing down via the stop script is the correct restructure. The scripts now abort fast on start/stop failure — intended.

**Still needed.** No check runs are recorded at this head; the CPU recipe gates still need a CI run. This touches the same two scripts as #242 and #246 — land in the order 243 → 242 → 246. Because these scripts produce benchmark numbers, one live bench run is needed before their results are trusted again; that is a later qualification step, not a request to run live workloads now.

**What would unblock.** A green CI run at this head, with live bench qualification to follow before the numbers are cited.

shiaho777:
Conflict note: `scripts/ci-validate.sh` overlapped in the `bash -n` list — I kept **both** main's `scripts/selftest-runtime-ablation.sh` and this PR's `scripts/bench-baseline-no-patches.sh`.

---

Rebased onto current `main` and pushed — the head is now `c7045c39`.

Ready for re-review.

plotarmordev (approved):
SIX final review approved at ff1203e4e7224a31184d32cdedd68ee5c2bd3d89. Both benchmark scripts use coordinated teardown and await the foreground launcher; the inert lifecycle suite exercises abort-before-measurement and ordering. Production bench scripts are unchanged from the original reviewed PR head, and the source-shaped suite has been replaced by behavioral coverage. Corrected documentation limits: SSH options bound connection establishment, not remote command runtime; an entirely unreachable worker remains warning-only in stop. Exact-head hosted validate succeeded: https://github.com/MiaAI-Lab/DeepSeek-v4-Flash-DSpark-2x-DGX-Spark/actions/runs/34962074694 (CPU recipe gates); scoped local lifecycle and individual shell syntax checks also passed. Full publication tree/range content gates passed with no new unexplained secret or deployment-local content. No live two-node benchmark or runtime qualification was performed; maintainer qualification remains required before citing benchmark numbers. This lands first in 243 -> 242 -> 246; no other PR is included in this decision.

PR #244, opened 2026-09-07
fix(start): worker stale-rank precheck exits 3 (already-running contract)

shiaho777:
## Description and Motivation

README states the contract: *"After a reboot, dockerd may already have restored the ranks (restart: unless-stopped); start then exits **3** (already running), not 1 … systemd: `SuccessExitStatus=3`."* The head already-running path honours it (`exit 3` + `already_running_hint`), but the **worker and worker2 stale-rank prechecks** did not:

```bash
ssh "$WORKER_HOST" "if docker ps … ; then echo '…stale rank…' >&2; exit 1; fi"
…
worker_rc=$?
echo "Cannot start: worker check on $WORKER_HOST failed (ssh exit $worker_rc)." >&2
exit "$worker_rc"
```

A stale worker rank made the launcher exit **1** (the remote probe's inner `exit 1`), and an unreachable worker made it exit **255** — both surfaced through the same "Cannot start" message. A systemd unit configured per the README (`SuccessExitStatus=3`) therefore treated the most common already-up topology — head-only reboot, dockerd restored only the worker — as a **failed start**, with no way to distinguish it from real ssh failure.

The remote probes now answer with **sentinel 42** when the container exists:

* `42` → `already_running_hint` + `exit 3` — same already-running contract as the head path;
* `0` → proceed;
* anything else (e.g. ssh 255) → `exit 1` with the ssh status in the message.

Behavior is otherwise unchanged; the ci-validate #72 guard strings (`exit 3`, `SuccessExitStatus=3`) remain present and the existing guard still passes.

## Related Issues

None filed. (Adjacent to #72, which introduced the head-side exit 3.)

---

## Testing

* `bash scripts/ci-validate.sh` — passes (no new failures; the #72 grep guard still green).
* New CPU suite `scripts/test-start-worker-stale-exit.py` (wired into `ci-validate.sh`): extracts the shipped worker/worker2 check blocks from the launcher and runs them against fake `ssh`/`docker` binaries —
  * clean worker → block passes;
  * stale rank → exit 3, stderr carries the already-running hint ("treat exit 3 as already-up") and the node-identifying message;
  * unreachable worker → exit 1 with `ssh exit 255` in the message;
  * source guards: exactly two `exit 42` sentinels, no `exit "$worker_rc"` passthrough left.

Not run: a live supervised reboot on the two-node lane (no hardware here); the exit-code paths are fully exercised by the fake-binary tests above.

---

## Checklist:

- [x] I have read the README and `docs/ENVS.md` and kept my changes consistent with them (this PR brings the worker path in line with the documented contract).
- [x] My change is reproducible and verified (CI validate + focused suite; no serving behavior change).
- [x] New runtime behavior is opt-in and default-off; defaults are unchanged (exit-code contract only, per README).
- [x] I updated the README, `docs/ENVS.md`, `.env.dspark.example`, and `CHANGELOG.md` for any knob, default, or behavior I changed (CHANGELOG entry added; README already documents the exit-3 contract this PR enforces).


plotarmordev:

Reviewed head: `87e6ed6ebcfbd9eabc8b5a4cf8d3abff4c0ba9bf`

**What static review established.** On current main the worker/worker2 stale-rank prechecks exit the raw return code (1 for stale rank, 255 for unreachable), while the head path and the README document exit 3 with `SuccessExitStatus=3` for the already-running case. This PR aligns the worker paths to sentinel exit 3 plus the existing `already_running_hint`, and maps ssh failure to exit 1. The 1→3 change for the stale-worker case is a deliberate contract alignment with the documented behavior; nothing in the repo matches on exit 1 for that case.

**Still needed.** No check runs are recorded at this head; the CPU recipe gates still need a CI run. If a systemd-supervised start is in use, a supervised-start smoke of the exit-3 path is warranted.

**What would unblock.** A green CI run at this head. Sequencing note: same launcher preflight region as #249 and #250; land 249 → 244 → 250.

shiaho777:
Rebased onto current `main` and pushed — the head is now `9d71805c`.

Ready for re-review.

plotarmordev (approved):
Approved at `8cd7024442b85a8c2fe95ce605ec79644ffcb4da` after integrating merged #249 without rewriting author history.

The two stale-worker probes keep both contracts: they use #249's `dssh` wrapper, return exit 3 for an existing rank, and return exit 1 for a connection/probe failure. The launcher bytes exactly match the earlier independently reviewed and live-qualified merge resolution. That qualification observed an actual stale worker and an already-running cluster without changing their container state. TP3 was covered by CPU fixtures, not a live TP3 run.

The fixture initially failed because its extracted script lacked `dssh`. It now includes the actual wrapper and handles its SSH options. Four behavioral cases pass. Two source-text checks and incidental message-wording assertions were removed rather than rewritten to pin the new source.

`bash scripts/ci-validate.sh` passed at this exact committed head. Image/torch-dependent skips remain reported. Publication scope and secret checks passed; no deployment-local files were introduced.

The original integration/runtime hold is satisfied. Normal merge still waits for the new hosted `validate` result. No local serving operation or fresh GPU qualification was performed: TheGrill owns those resources.

PR #246, opened 2026-09-07
fix(bench-baseline): stop grep -c counters double-printing zeros

shiaho777:
## Description and Motivation

Both bench-baseline scripts verify patch state after the baseline boot with counters like:

```bash
c22=$(grep -c 'nvfp4_ds_mla' '$VLLM/models/deepseek_v4/sparse_mla.py' 2>/dev/null || echo 0)
```

On GNU grep, `grep -c` with zero matches prints `0` **and** exits 1 — so the `|| echo 0` appends a *second* `0` and the variable becomes the two-line string `0\n0` (reproduced). The subsequent `echo $c22 $c49 $c50 $c07` word-splits into up to **eight** fields, and the `cut -d' ' -f1…-f4` report the operator eyeballs ("expect 0 0 0 0" / "Issue #22 ≥2, others = 0") reads **shifted** values whenever some counters matched and others didn't — e.g. a real #49486 hit can be reported as 0 while #50312 shows #49486's count. The verification step meant to prove "baseline applied exactly the intended patches" was unreliable in exactly the mixed states it exists to catch.

Fix: `|| true` plus a `${c:-0}` default. That yields exactly one token in all three cases — match (`N`), no match (grep already prints `0`; `|| true` swallows the exit code), and missing file (empty → default `0`).

No knob, default, or serving behavior changes — benchmark scripts only.

## Related Issues

None filed.

---

## Testing

* New CPU suite `scripts/test-bench-baseline-patch-counters.py` (wired into `ci-validate.sh`): extracts the shipped CHECKS payload from both scripts and runs it against a fake vLLM tree —
  * all files missing → `0 0 0 0`;
  * first counter 2, second counter zero-match (file exists), rest missing → `2 0 0 0` with exactly four fields (the old double-print turned this into 5+ fields and shifted positions);
  * all files present with zero matches (the exact double-print case) → four single-token zeros;
  * source guard: the `grep -c … || echo 0)` idiom cannot come back.
* `bash -n` clean on both scripts.

Not run: the live two-node baseline (needs the DGX Spark pair; no hardware here).

---

## Checklist:


plotarmordev:

Reviewed head: `0bd391c1f91db941c641add2d997ad28c346dff3`

**What static review established.** GNU `grep -c` prints `0` and still exits 1 on no match, so the `grep -c … || echo 0` idiom at eight sites across the two bench-baseline scripts double-prints zeros in mixed match/no-match states and shifts the report fields. The `|| true` + `${c:-0}` replacement fixes the counting; the shifted-field consequence was verified by reading the reporting path. One review note carried from a second pass: some of the new tests in this batch (including here) mix behavioral checks with source-text assertions, so the guard coverage is not purely behavioral — fine for now, worth knowing when the tests are next touched.

**Still needed.** No check runs are recorded at this head; the CPU recipe gates still need a CI run. This touches the same two scripts as #243 and #242 — land in the order 243 → 242 → 246. And because these are bench scripts, their numbers should be re-qualified with a live bench run (after the batch settles) before anyone trusts the reported counters again; that is a later qualification step, not a request to run live workloads now.

**What would unblock.** A green CI run at this head, sequenced after #243/#242.

shiaho777:
Rebased onto current `main` and pushed — the head is now `3513615b`.

Ready for re-review.

plotarmordev (approved):
Final source review at daae0018652090b8a69ff606b1485777c20093ef: all eight counters retain one output token for matches, zero matches and missing files; behavioral payload tests remain and source-shape assertions are removed. Main 9826756 containing PR242 is integrated normally, preserving the hotfix and lifecycle fixes. Hosted validate run 34968316755 succeeded at this exact head; actual log inspection confirms all twelve late guards pass and no recipe guard failure. Live counter and benchmark measurement qualification remains maintainer-owned and deferred; CPU success is not a live performance claim.

PR #247, opened 2026-09-07
fix(ops): status/logs exit non-zero on probe failure; status watches VLLM_PORT

shiaho777:
## Description and Motivation

Two defects made `./status-…` / `./logs-…` unusable as health probes:

**1. Wrong port in the status listing.** The script resolved `PORT="${PORT:-8888}"` *before* sourcing `.env.dspark`, and the `ss` listing used `$PORT` while the API probe used `${VLLM_PORT:-8888}`. With `VLLM_PORT=9000` in `.env.dspark`, the API probe hit 9000 but the port listing watched 8888. `VLLM_PORT` (the compose knob) is now resolved once after the env file is sourced, with `PORT` kept as a legacy alias, so the listing and the probe agree.

**2. Always exit 0.** Every probe ended in `|| true`: the scripts exited 0 with containers down, ssh failing, and the API returning errors — a supervisor or runbook treating them as health checks got a green light on a dead cluster. Real probe failures are now counted and surfaced:

* **status**: head/worker/worker2 `compose ps`, image inspects, and the API curl count as failures; the container-grep and `ss` listings stay informational (no-match is a state, not a probe failure);
* **logs**: worker/worker2 ssh failures count; a failing `compose logs` does not (a stopped cluster legitimately has no logs);
* both print the full diagnostic output as before, then exit 1 with a summary when any counted probe failed.

ssh probes also gained `-o BatchMode=yes -o ConnectTimeout=10` (mirroring the stop script's hardening): a health probe that hangs forever on a stalled node is worse than one that fails fast.

## Related Issues

None filed.

---

## Testing

* New CPU suite `scripts/test-status-logs-probes.py` (wired into `ci-validate.sh`) runs both shipped scripts end-to-end against fake `docker`/`ssh`/`curl`/`ss` binaries:
  * healthy cluster → exit 0;
  * API down → exit 1 with `API probe failed`;
  * worker unreachable → exit 1 with the node in the warning;
  * `ss` watches `VLLM_PORT` (9999), never the legacy 8888 default;
  * every ssh probe carries `BatchMode=yes` + `ConnectTimeout=10`;
  * logs: `compose logs` failing on a stopped cluster is not an error.
* `bash -n` clean on both scripts.

Not run: against a live cluster (no hardware here); behavior is covered by the fake-binary end-to-end runs.

---

## Checklist:

- [x] My change is reproducible and verified (CI validate + focused suite; no serving behavior change).
- [x] New runtime behavior is opt-in and default-off; defaults are unchanged (ops scripts only; exit codes now reflect reality).
- [x] I updated the README, `docs/ENVS.md`, `.env.dspark.example`, and `CHANGELOG.md` for any knob, default, or behavior I changed (CHANGELOG entry added).


plotarmordev:

Reviewed head: `3c2aacd1d88e9f62be6927429dd9202e6e34f054`

**What static review established.** On current main, `status` resolves `PORT` before sourcing the env file while the API URL uses `VLLM_PORT`, so the port probe and the API check can disagree; and both `status`/`logs` end every probe with `|| true`, so they always exit 0 even on a dead cluster. This PR resolves the port after the env source (keeping the `PORT` legacy alias), counts probe failures, exits 1 on real failures, and gives the ssh probes BatchMode/ConnectTimeout. No in-repo consumer depends on the always-exit-0 behavior. The exit-code contract change (0 → 1 on dead cluster) is the fix, not a regression — but any external supervisor relying on always-0 would flip.

**Still needed.** No check runs are recorded at this head; the CPU recipe gates still need a CI run.

**What would unblock.** A green CI run at this head.

shiaho777:
Conflict note: `scripts/ci-validate.sh` test wiring overlapped — I kept **both** main's `test-gb10-install-failclosed` and this PR's `test-status-logs-probes` invocation.

On your test-hygiene note (behavioral vs source-text assertions): agreed, and the next touch of these tests should move the remaining source-shape checks to behavioral ones.

---

Rebased onto current `main` and pushed — the head is now `6d285852`.

Ready for re-review.

plotarmordev (approved):
Final SIX review at `541174de7280d6c62c6e4a945cff9b73e55909cc`: approved for normal merge after integrating current main `827abe4b3953278fcf2719c6bd9b6c9a59f3a5e4` without conflicts. Original authorship and ancestry preserved; only the scoped probe-test correction and clean integration were added.

Hosted validate passed at this exact head: https://github.com/MiaAI-Lab/DeepSeek-v4-Flash-DSpark-2x-DGX-Spark/actions/runs/34955928175 (CPU recipe gates, including all seven status/logs cases). Inspected the complete log: no recipe-gate FAIL or unittest failure; seven FAIL/FAIL-CLOSED diagnostic lines belong to negative fixture tests that completed successfully. Worktree and every outgoing commit were scanned for credential patterns and deployment-local content, with no matches.

The probes now report counted failures with exit 1 and resolve the status port after sourcing the environment. Important qualification: BatchMode/ConnectTimeout bounds connection establishment, not an entire SSH command. Head compose-log failure remains nonfatal; remote compose-command failures remain counted through SSH status, so this does not promise exit 0 for every stopped-worker scenario. No live cluster or benchmark was run; external supervisors relying on the previous always-0 result will observe the intended contract change.

PR #248, opened 2026-09-07
fix(ab-boot): escape sed replacement; memory gate fails closed

shiaho777:
## Description and Motivation

`scripts/ab-boot.sh` (the A/B boot+measure driver) had two defects:

**1. sed injection into the operator's live `.env.dspark`.** The script rewrote `K=V` with K unescaped as a grep/sed -E pattern and V unescaped as the sed *replacement*:

```bash
grep -qE "^${K}=" .env.dspark
sed -i -E "s|^${K}=.*|${K}=${V}|" .env.dspark
```

An `&` in V expands to the whole match, a `|` ends the `s|||` expression early, and a newline in V splits the env file mid-edit — any of these silently corrupts `.env.dspark`, which the very next lines stop/start the cluster from. Now: K is validated as an env-var name (alnum+underscore is literal in ERE, so no pattern injection), multiline V is rejected outright, and `&`, `|`, `\` are escaped for the replacement via `V_SED`.

**2. Memory gate fail-open.** The script runs `set -uo pipefail` **without `-e`**. When the worker ssh failed, `MW` came back empty and `[ "" -lt "$GATE" ]` errored (rc 2) inside the `||` — the gate condition evaluated **false**, so the benchmark ran against a possibly-broken boot instead of exiting 2. Unreadable MemAvailable on either node (empty or non-numeric) now exits 2 with a clear message before the threshold comparison. Also: `WORKER` now takes only the first `WORKER_HOST=` line (`head -n1`) in case the env file ever carries a duplicate.

No knob, default, or serving behavior changes — A/B tooling only.

## Related Issues

None filed.

---

## Testing

* New CPU suite `scripts/test-ab-boot-sed-memgate.py` (wired into `ci-validate.sh`): extracts the shipped env-edit block and memory-gate block from the script and runs them —
  * escape matrix: `A&B`, `a|b`, `a\b` values are written literally; metachar key (`MAX.NUM`) and multiline value are rejected without touching the file; missing key still exits 3;
  * gate: above/above → pass; below → exit 2 `MEM_BELOW_GATE`; empty/garbage worker value → exit 2 (fail-closed, the original bug); empty head value → exit 2;
  * source guards: `head -n1` on the WORKER line, `V_SED` used in the replacement.
* `scripts/ab-boot.sh` added to the `bash -n` CI gate (it was ungated); `bash -n` clean.

Not run: a live A/B boot (needs the two-node lane; no hardware here).

---

## Checklist:

- [x] New runtime behavior is opt-in and default-off; defaults are unchanged (tooling-only fix).


plotarmordev:

Reviewed head: `8a4897467f82a10254c907015fbf698e19e8dbfb`

**What static review established.** Both `ab-boot.sh` defects were confirmed on current main: K/V flow unescaped into grep/sed (metacharacters in the sed replacement can corrupt the substitution), and the memory gate is fail-open — under `set -u` without `-e`, an unreadable MemAvailable leaves the variable empty, the `[ "" -lt GATE ]` comparison errors inside `||`, and the bench proceeds anyway. This PR validates K as an env-var name, rejects multiline V, escapes `& | \` in the replacement, and makes the gate fail closed. Tooling only; the stricter failures are the intended behavior change.

**Still needed.** No check runs are recorded at this head; the CPU recipe gates still need a CI run.

**What would unblock.** A green CI run at this head.

shiaho777:
Rebased onto current `main` and pushed — the head is now `7174a3d3`.

Ready for re-review.

plotarmordev (approved):
Final SIX approval at `c0140218f0b955aff6e2dfa7656c6c69fed08b6c`, after normal integration of main `82ee8348574c6b37bef097eb2d89d3b99c20eb73`. The only manual integration resolution retained both independent CHANGELOG additions; source behavior and original author ancestry were preserved. Main's test-only correction removes the two source-shape assertions, leaving twelve behavioral cases.

Exact-head hosted validate passed: https://github.com/MiaAI-Lab/DeepSeek-v4-Flash-DSpark-2x-DGX-Spark/actions/runs/34957321460. The full log was inspected, including late gates: no recipe-gate FAIL or unittest failure; the seven FAIL/FAIL-CLOSED diagnostic lines are expected negative-fixture output. Worktree and every outgoing commit were scanned for credential/deployment-local patterns with no matches.

Reviewed the actual env-edit and memory-gate paths: replacement metacharacters are escaped, multiline values rejected, missing keys retain exit 3, and empty/nonnumeric MemAvailable fails closed with exit 2. This is scoped sed-replacement robustness, not a blanket claim that arbitrary values are safe to source as shell. No live boot, benchmark, model, server, container or SSH operation was performed; the removed first-WORKER source assertion is not replaced with a claim of runtime host-selection coverage.

PR #249, opened 2026-09-07
fix(start): route every ssh/scp through fail-fast dssh/dscp wrappers

shiaho777:
## Description and Motivation

Only the two preflight probes in `start-deepseek-v4-flash-dspark.sh` carried `-o BatchMode=yes -o ConnectTimeout=10`. The ~100 ssh/scp invocations **after** preflight (GID resolve, compose/env/patch sync, remote compose, image checks, stale-rank checks) had neither — so if the worker drops or the link stalls after preflight, the launcher blocks on TCP retransmits for many minutes with no output, indistinguishable from a slow start, at exactly the moment `restart: unless-stopped` makes cluster state ambiguous.

The launcher now defines two wrappers next to its other globals and routes every invocation through them:

```bash
dssh() { ssh -o BatchMode=yes -o ConnectTimeout=10 "$@"; }
dscp() { scp -o BatchMode=yes -o ConnectTimeout=10 "$@"; }
```

BatchMode can never prompt (passwordless keys are already a documented prerequisite) and ConnectTimeout fails a stalled connect in 10 s instead of hanging. The two preflight probes use the same wrappers instead of duplicated inline options. This PR also includes the worker stale-rank check sites, so they fail fast too.

One harness follows the change: `scripts/test-env-normalisation.py` extracts the publish block from the launcher (which now calls `dssh`), so the harness injects the shipped wrapper definition and its fake `ssh` binary skips `-o` option pairs before the host argument.

## Related Issues

None filed.

---

## Testing

* `bash scripts/ci-validate.sh` — passes (no new failures; `test-env-normalisation.py` harness updated and green).
* New CPU suite `scripts/test-start-ssh-wrappers.py` (wired into `ci-validate.sh`):
  * pins the wrapper definitions (options first, correct order);
  * pins that the wrappers do not recurse (the mechanical rewrite briefly produced `dssh() { dssh "$@"; }` — caught and fixed, now guarded);
  * proves no bare `ssh "`/`scp "` invocation survives in the launcher;
  * exercises both wrappers against fake `ssh`/`scp` binaries and asserts the options precede the target;
  * the "ssh exit $worker_rc" diagnostic text is untouched.

Not run: a live worker-drop mid-start on the two-node lane (no hardware here).

---

## Checklist:

- [x] My change is reproducible and verified (CI validate + focused suites).
- [x] New runtime behavior is opt-in and default-off; defaults are unchanged (transport hardening only).

Notes for the batch: (1) CHANGELOG entries sit at the same file position by convention, so the last PRs to merge may need a trivial rebase; (2) if #244 (worker stale-rank exit 3) merges first, its `FAKE_SSH` in `scripts/test-start-worker-stale-exit.py` needs the same option-skipping one-liner as the env-normalisation stub here — happy to rebase whichever lands second.


plotarmordev:

Reviewed head: `720ca99805c43ce257c2692da3a794f0efd745d6`

**What static review established.** On current main only the two preflight probes carried ssh options; the roughly 58 other `ssh`/`scp` call sites did not. This PR routes them through `dssh`/`dscp` wrappers with `BatchMode=yes` and `ConnectTimeout=10`, and the test-env-normalisation harness update is consistent. BatchMode never prompting is fine here because passwordless keys are already a documented prerequisite. One scope note: `files/nfs-share.sh` still has two unwrapped plain-ssh sites (docker volume rm/create), so the "every ssh/scp" framing is slightly wider than the diff — cosmetic, not a blocker. Also for accuracy: `BatchMode`/`ConnectTimeout` bound connection setup and authentication prompting, but do not by themselves guarantee a bounded lifetime for every remote command.

**Still needed.** No check runs are recorded at this head; the CPU recipe gates still need a CI run, and because this touches every sync path, one live start is warranted before merge.

**What would unblock.** A green CI run plus one clean live start at this head.

shiaho777:
Also addressed your two review notes:

* **Scope**: wrapped the remaining plain-`ssh` sites — `files/nfs-share.sh` (`nfs_ensure_host_volume`; it hardens inline because the stop script also sources that file and defines no `dssh`) and **three sites main added after this PR was opened** (the runtime-ablation verify helper and its sync block). `scripts/test-start-ssh-wrappers.py` now also asserts every `ssh` invocation in `nfs-share.sh` carries the options.
* **Accuracy**: agreed — `BatchMode`/`ConnectTimeout` bound connection setup and auth prompting, not the lifetime of every remote command; the PR description stays as you reviewed it.

Two harnesses needed to follow the wrapper change (they run extracted launcher chunks outside it): `test-dspark-api-keys.py`'s worker-sync loop and `test-nccl-ib-hca-gid-resolve.sh`'s resolver. Both now inject the shipped `dssh`/`dscp` definitions (with plain-ssh fallbacks) and their ssh stubs skip the `-o` pairs.

---

Rebased onto current `main` and pushed — the head is now `b9b095e9`.

Ready for re-review.

plotarmordev:
**Maintainer-owned compatibility hold; current CPU CI requirement addressed.**

At `b9b095e9adffb0f8ddc93f3086877b259a995363`, `validate` is green ([run](https://github.com/MiaAI-Lab/DeepSeek-v4-Flash-DSpark-2x-DGX-Spark/actions/runs/34558399674)); the previously noted NFS SSH sites now carry the options. The remaining gate is ours, not missing author work.

**Next-action owner: maintainer (plotarmordev).** Once an owner-approved maintenance window is available, with advance notification before local tests, qualify one clean two-node start on the exact integrated head proposed for adoption: worker sync, patch/env publication, and Compose startup on both ranks. Record that head and output, retain exact-head CPU/integration checks, then re-review the hold. If integration changes the head, historical green CI does not qualify the new result. The serving hardware is currently in use; no live run, author test request, or stale rebase request is being commissioned now.

For precision, `ConnectTimeout=10` limits connection establishment (including the initial SSH handshake), not the duration of an established remote command or a guaranteed ten-second failure after a mid-start connection drop. The launcher comment must not be read as that stronger guarantee; a mid-start-drop experiment is not an additional gate. This PR is not being merged here.

plotarmordev (approved):
Approved for normal merge at `b9b095e9adffb0f8ddc93f3086877b259a995363`, first in the #249 → #244 order.

The merge candidate against current main `e369ce6` has exactly the same tree as the independently reviewed integration candidate. `bash scripts/ci-validate.sh` passed again on that tree, and the published head's hosted `validate` check is green. Image/torch-dependent skips remain skips, not runtime passes. The publication scan found no outgoing secret or deployment-local additions; the sole full-tree address match was unchanged public benchmark documentation.

The earlier authorized two-node qualification passed worker sync, environment/patch publication, cold startup and 18/18 warmup requests. A real connection-refusal case failed promptly without creating containers. We did not test a silent mid-start network partition or TP3 live.

The original qualification hold is lifted for this PR. This review performs no deployment: TheGrill now owns the local model and machines. #244 must keep these transport wrappers when its overlapping worker probes are integrated.

PR #250, opened 2026-09-07
fix(start): re-check the head port immediately before the head bind

shiaho777:
## Description and Motivation

The launcher's head-port precheck (`ss -ltn "( sport = :$VLLM_PORT )"`) runs at the already-running/stale-rank preflight — **minutes** before `compose up -d` actually binds the port, with file syncs, GID resolve, and the worker `up -d` calls in between. A process that takes the port in that window surfaces only as a failed bind at `up -d`, **after** worker ranks have already started — leaving a half-up cluster and a confusing error.

The check is now `assert_head_port_free()`, called twice:

1. **early** (unchanged position) — fail fast before the expensive worker sync;
2. **immediately before the head `up -d`** — narrowing the check→bind window to the bind itself. The late failure message says explicitly that the worker ranks are already up and need `./stop-deepseek-v4-flash-dspark.sh` before retrying.

The race cannot be eliminated entirely (any process can still grab the port in the final milliseconds — docker's bind error covers that), but the actionable window shrinks from minutes to nothing, and the late failure now comes with cleanup guidance instead of a buried docker error.

## Related Issues

None filed.

---

## Testing

* New CPU suite `scripts/test-start-port-recheck.py` (wired into `ci-validate.sh`):
  * extracts the shipped `assert_head_port_free()` and runs it against a fake `ss`: port busy → exit 1 with the conflicting-service message; port free → pass; no `ss` binary → check skipped (same as before);
  * pins call placement: defined once, called exactly twice, early call precedes the worker sync, late call immediately precedes `compose_base 0 "" up -d` with the `./stop-…` guidance.

Not run: a live mid-start port grab on the two-node lane (no hardware here).

---

## Checklist:

- [x] My change is reproducible and verified (CI validate + focused suite).
- [x] New runtime behavior is opt-in and default-off; defaults are unchanged (failure-timing hardening only).


plotarmordev:

Reviewed head: `70b448fefef6f7c8b097ca0830c3622054229973`

**What static review established.** The TOCTOU window is real on current main: the head-port check runs once in preflight, then roughly 540 lines of syncs and worker bring-up pass before the head actually binds the port. This PR converts it to `assert_head_port_free()` called early and again immediately before the head `up -d`, with worker-cleanup guidance on late failure. The residual bind-time race remains covered by the docker bind error, as the PR acknowledges.

**Still needed.** No check runs are recorded at this head; the CPU recipe gates still need a CI run. Ideally one live start exercising the late-failure path.

**What would unblock.** A green CI run at this head. Sequencing note: this touches the same launcher preflight region as #249 and #244; landing order 249 → 244 → 250 minimizes rebase churn.

shiaho777:
Rebased onto current `main` and pushed — the head is now `830f69f5`.

Ready for re-review.

plotarmordev (changes requested):
## Hold update

Public head checked: `830f69f54a57b8bdc81d8408cdb6e2502b40f78e`.

**Change needed:** The missing-`ss` test in `scripts/test-start-port-recheck.py` still searches the host's `/usr/bin` and `/bin`. It can therefore find a real `ss` and change result depending on what is listening on the host. Isolate that fixture's PATH. Remove the `CallPlacement` source-text checks; they do not prove the port guard's behavior.

**What we tested:** The published fixture failed in the combined CPU run with a listener present and passed when that host condition changed. The corrected fixture passed. In the integrated live run, a listener appeared after the worker was created; the late guard refused startup before creating the head. That runtime check passed, but it does not publish the fixture fix.

**Next action — me, @plotarmordev:** Publish and review the local fixture correction. The author does not need to redo it or repeat hardware testing now.

**Before ready:** On the published corrected head, run `python3 scripts/test-start-port-recheck.py -q` and `bash scripts/ci-validate.sh`. Preserve free-port, occupied-port and genuinely missing-tool behavior. Land after #249 and #244. The runtime reviewer approved the behavior; the unpublished test correction is why this remains held.

PR #149, opened 2026-08-27
security: minimize serving container capabilities

SWOOPPMAIN:
## Summary
- Drop Docker default capabilities; retain only boot-time SETUID, SETGID and SETPCAP for the one-way transition to the numeric non-root serving identity.
- Execute vLLM and health checks with an empty capability bounding set and no-new-privileges; IPC_LOCK is not retained.
- Mount checkpoint/cache root and repository patch sources read-only, with only named runtime/JIT caches and temporary storage writable.
- Add explicit one-time runtime-cache ownership migration, checkpoint-overlap refusal, and absolute runtime-user path/dry-run instructions. Normal preparation never changes ownership.

## Latest published source and CPU follow-up
Current published head: `ff6a72a375fe3224116d0484328b56eb5db63a51`, a test-only descendant of the production-source review below. Incidental source/wording assertions were removed, not re-pinned. Main retained and ran the 34-check inert ownership suite, 44 grammar behavior tests, 28 hotfix-chain cases, 9 MXFP4 patcher cases and 19 shared-memory behavior tests.

[Hosted validate 34982419725](https://github.com/MiaAI-Lab/DeepSeek-v4-Flash-DSpark-2x-DGX-Spark/actions/runs/34982419725) succeeded with all twelve late guards individually verified. It checked out `d45ae2f7e2a7e267259a084266a4863df8181eac`. This is a GitHub merge-ref tree that differs from the branch tree: it is not exact-branch-tree or current-main landing clearance. No real ownership migration, capability state, two-node upgrade or serving qualification is established; the maintainer-owned HOLD remains.

## Earlier production-source review and CPU evidence
Earlier review head: `90bb91bd21e9139b21d7d718c00c0e54a9c8442f` (normal integration of main anchor `f3082a674134bb9bf66894a0c9cf5d2ae97b1aea`).
Main ran the 37-check inert/unprivileged ownership suite and actual `--help` successfully at this exact head. These checks do not exercise real chown, container capability state, or the two-node upgrade; earlier disposable-container evidence is historical, not qualification of this head. No full current-main or hosted-CI pass was claimed at that earlier review.

## Maintainer-owned HOLD — do not merge
Owner: maintainer/Main for integration order, maintenance-window authorization and hardware access; the designated merge owner must retain this hold.
Reason: canonical main has advanced beyond the integrated anchor, and actual rooted migration plus the previously failing two-node in-place upgrade remain unqualified.
[...]

PR #253, opened 2026-09-08
ci: gate compose rendering, issue #43 coverage, derived hotfix existence

shiaho777:
## Description and Motivation

Three CI coverage gaps, each of which lets a load-bearing regression pass green:

**1. Compose rendering was never gated.** Every compose gate in `ci-validate.sh` is a text grep, so a `docker-compose.dspark.yml` that no longer renders (bad YAML, duplicate keys, broken interpolation) passed CI as long as the grepped substrings survived. `docker compose config -q` now renders all three combos (base, +NFS override, +Stage-C override) and fails the run otherwise. The gate requires the docker compose plugin — preinstalled on ubuntu-24.04 runners; a missing plugin is a loud CI failure, not a silent skip.

**2. Issue #43 had zero CI coverage.** `patches/hotfix-dsv4-issue43-decode-fairness-and-diag.py` is applied unconditionally with `|| exit 1` on every boot, yet neither of its tests ran anywhere:
* `tests/sim/test_issue43_scheduler_sim.py` (pure CPU; exits 1 when bounded-service or diag-budget invariants are violated across the six cold cells) now runs in the unit-test section;
* `tests/test_issue43_patchapply.py` needs the pinned runtime image (`docker run … cat scheduler.py`), so — like the live verifiers — it is py_compile-gated only, along with `tests/test_issue175_routing_kind_once.py` (also previously unreferenced anywhere).

**3. The mounted-hotfix existence list drifted.** The curated "mounted hotfix files must exist" list (20 entries) had silently fallen 12 files behind the compose mount list — including the always-on issue #43 patch: deleting it would have passed CI while breaking every live boot. The check now **derives** the list from the `hotfix-*` references in `docker-compose.dspark.yml` + `start-deepseek-v4-flash-dspark.sh` (38 files today), so future patch additions cannot drift again. All 20 previously curated files are covered by the derived set.

Also: `scripts/ab-measure.sh` and `scripts/run-audit.sh` join the `bash -n` gate — they were the last two ungated shell scripts.

## Related Issues

None filed.

---

## Testing

* `bash scripts/ci-validate.sh` — passes locally; new gates verified individually:
  * `ok compose config (base)`, `(nfs override)`, `(stage-c override)` — all three render (only unset-variable warnings, as in production use);
  * `ok test_issue43_scheduler_sim` — the simulator prints its six-cell table and PASSes (exit 1 on violated invariants, confirmed by reading its exit paths);
  * derived existence check run standalone: 38 referenced hotfix files, 0 missing; deleting any one flips the gate to `bad` (loop tested).
* `bash -n scripts/ci-validate.sh` clean.

Note for the maintainer: `ci-validate.sh` runs `set -euo pipefail`, so sections after a failing test never execute — pre-existing behavior, unchanged here (locally I verified the late sections standalone for that reason).

---

## Checklist:

- [x] My change is reproducible and verified (CI validate; the new gates were each exercised).
- [x] New runtime behavior is opt-in and default-off; defaults are unchanged (CI-only change).
- [x] I updated the README, `docs/ENVS.md`, `.env.dspark.example`, and `CHANGELOG.md` for any knob, default, or behavior I changed (CHANGELOG entry added; no user-facing knob/docs surface).

Note: this PR is one of a batch of small fixes; CHANGELOG entries sit at the same file position by convention, so the last ones to merge may need a trivial rebase — happy to do that on request.


plotarmordev:
**Maintainer status: CHANGES REQUESTED — rebase required, then execute the updated CPU checks.**

Reviewed head: `fc3f1aa5031b74c3ef322e76f048336d0a6597e8`

**What static review established.** The coverage premise is real on current main: the mounted-hotfix existence check in `ci-validate.sh` is a hand-maintained 21-entry list that has drifted behind the compose/start references (missing, among others, the always-on issue-#43 patch that the boot chain applies with `|| exit 1`); there is no `docker compose config -q` render gate; and the issue-#43 scheduler tests exist but run nowhere. Deriving the list from compose/start references (39 refs on main, none missing) is the right fix direction, and the added render gate and orphan-test execution are genuine coverage gains.

**Why changes are requested.** This head is CONFLICTING with current main: main independently added the runtime-ablation hotfix entry to the exact curated list this PR replaces. The rebase must preserve that newer CI coverage rather than dropping it.

**Still needed.** Rebase onto current main preserving all existing entries, then execute the updated CPU checks (`ci-validate.sh`) at the rebased head.

**What would unblock.** A rebased head with the CPU gates executed and green, including the new render gate and the derived hotfix-list check. Sequencing note: this PR should land after the other `ci-validate.sh`-touching PRs in the batch so it absorbs their insertions in one rebase.

shiaho777:
Rebase handled your change request:

* Kept main's newer coverage: the derived existence check now **also asserts a required subset** of unconditionally-applied boot patches (`hotfix-dsv4-runtime-ablation.py` — the entry main added to the curated list while this PR was open — plus issue21, issue43, vision-exp, empty-encoder). Losing any of those references now fails CI explicitly instead of silently narrowing the derived set.
* Derived set at this head: 39 referenced hotfixes, 0 missing, required subset all present. `docker compose config -q` renders all three combos; the issue #43 simulator runs and PASSes.
* Per your sequencing note, this should land after the other `ci-validate.sh`-touching PRs so it absorbs their insertions in one rebase — happy to rebase again if anything else lands first.

---

Rebased onto current `main` and pushed — the head is now `a9900396`.

Ready for re-review.

plotarmordev (commented):
Source/CPU review at `512d22302bccd9e744b4b68f75207cf1dbcc8d67`; **MERGE HOLD — sequencing is not discharged**.

Normally integrated canonical main `e369ce6e506ab09199e0f2e29e7f04e62bb9efbe` without rewriting author ancestry. All incoming guards/registrations remain, the only accumulated-failure check runs after the final passthrough guard, and the original three compose-render combinations, issue #43 coverage and derived-hotfix scope are preserved (39 references, including all 21 formerly curated entries and the required boot subset).

The registered exit-code regression now has three behavioral cases, without diagnostic-wording assertions: pristine exits 0; first-late and final-guard failures exit 1. Main independently ran the exact-head gate (3 passed) and a private counterprobe moving the check back before the late guards (pristine passes, both late-failure cases fail as expected).

[Full configured hosted validate](https://github.com/MiaAI-Lab/DeepSeek-v4-Flash-DSpark-2x-DGX-Spark/actions/runs/34973137789) succeeded for this published head. Logs show all three compose render gates, issue #43 simulator, the 3-case fail-closed regression, derived-hotfix check and all twelve late guard results passing. Image/torch-dependent cases retain their reported skips; this is CPU-source qualification, not GPU/image-runtime qualification.

**Next-action owner: Main / maintainer.** PR #253 must remain LAST among CI-touching changes, including held #244/#250 and every other pending CI-touching dependency. Completed #242/#246/#254/#255 do not satisfy the remaining order, and unavailable maintenance does not waive it. Unblock only after the maintainer resolves all predecessor dispositions, integrates then-current canonical main and requalifies the exact cumulative head. No merge performed or authorized here; this review does not claim merge readiness.

plotarmordev (commented):
## Hold update

Public head checked: `512d22302bccd9e744b4b68f75207cf1dbcc8d67`.

**Why it is held:** Integration order, not an outstanding author defect. The published head already passed its CPU checks, including the fail-closed CI regression.

**What we checked:** We reviewed the current-head results and the combined CPU chain. That private combined chain is not the same as qualifying the final public merge candidate.

**Next action — me, @plotarmordev:** Finish the selected predecessors in order (#249, #244, then the corrected #250). If #152 is selected to land first, publish its integration fixes too. Then integrate current main here and run `bash scripts/ci-validate.sh` at that exact combined commit.

**Author action now:** None. Please do not rebase repeatedly while we are choosing or landing predecessors. Held #149, #220 and #256 are not author blockers for this PR. They can carry their own CI integration and qualification when they are ready. This PR can proceed after the selected preceding changes and its final CPU gate—not after every held PR in the repository.

PR #256, opened 2026-09-08
security(nfs): root_squash the export; fail closed on empty client list

shiaho777:
## Description and Motivation

The DSpark NFS exporter shares the head's HuggingFace cache root — which also holds the 0600 `huggingface-cli` **token file** — from a privileged, host-networked container on the ConnectX link. Three defaults weakened that:

**1. `no_root_squash` exposed the HF token.** The export options mapped client-side root to server root, so `$HF_CACHE/token` (an HF write-scope credential) was readable regardless of its 0600 mode. The default is now `root_squash`: remote root maps to nobody — world-readable cache blobs (0644) stay readable to the worker's root container, the token does not. (If an operator's cache has non-world-readable blobs, the worker read path needs those modes relaxed — noted in the PR because it is the one operational edge of squashing.)

**2. `insecure` allowed mounts from unprivileged client ports.** Dropped; the worker's kernel NFS client (docker volume) mounts from privileged ports anyway, so nothing legitimate changes.

**3. Fail-open wildcards wrote world exports.**
* `files/nfs-server/entrypoint.sh`: `CLIENTS="${NFS_CLIENTS:-*}"` — an empty variable produced `/export *(…)`. The entrypoint now refuses to start with a clear message.
* `files/nfs-share.sh` `nfs_clients()`: `echo "${worker_ip:-*}"` — when CIDR detection failed AND no worker IP was known, the client list became `*`. It now fails the launch with an explicit refusal instead.

`nfs_grant_subnet()` (additive grants on a running exporter) keeps its opts default in lockstep (`root_squash`, no `insecure`).

The export stays read-only; client-IP restriction on the RoCE /24 remains the access model (unchanged) — this PR removes the foot-guns around it, not the model.

## Related Issues

None filed.

---

## Testing

* New CPU suite `scripts/test-nfs-hardening.py` (wired into `ci-validate.sh`):
  * executes the shipped entrypoint guard block: empty `NFS_CLIENTS` → exit 1 with guidance; explicit list → pass;
  * runs the shipped `nfs_clients()` against a stub `ip` binary through all four branches (CIDR+worker IP → `ip,net`; CIDR only → `net`; worker IP only → `ip`; neither → exit 1, never `*`);
  * pins the opts defaults in both files (`root_squash` present, `no_root_squash`/`insecure` absent) and bans the wildcard forms (comments excluded).
* `bash -n files/nfs-share.sh files/nfs-server/entrypoint.sh` clean.

Not run: an actual NFS mount on the two-node lane (no hardware here). One live check worth doing before merge: worker boots with `DSPARK_WORKER_HF_NFS=1` against an operator cache whose blobs are 0644 (standard `hf download` modes) — the expected common case.

---

## Checklist:

- [x] My change is reproducible and verified (CI validate + focused suite).
- [x] New runtime behavior is opt-in and default-off, matching the recipe's existing knobs; defaults are unchanged unless the PR is explicitly about a default — this PR **is** explicitly about hardening defaults, and the export's functional behavior (read-only weights to the worker) is preserved.
- [x] I updated the README, `docs/ENVS.md`, `.env.dspark.example`, and `CHANGELOG.md` for any knob, default, or behavior I changed (CHANGELOG entry added; no knob semantics changed).

Note: this PR is one of a batch of small fixes; CHANGELOG entries sit at the same file position by convention, so the last ones to merge may need a trivial rebase — happy to do that on request.


plotarmordev:
**Maintainer status: HOLD — security intent endorsed, compatibility evidence required before adoption.**

Reviewed head: `7a9d38cd247eec345b94763e245cd53ba8ace8e2`

**What static review established.** The problem is real on current main: the NFS exporter defaults to `no_root_squash,insecure` with a wildcard client default (`NFS_CLIENTS:-*`, `${worker_ip:-*}` fallback), running privileged with host networking while exporting the HF cache root that holds the 0600 token. The patch direction — `root_squash`, drop `insecure`, fail closed on an empty client list, and lockstep opt-in via `nfs_grant_subnet` — is the right shape, and the feature stays opt-in (`DSPARK_WORKER_HF_NFS` defaults off).

**Why HOLD rather than merge.** This is a compatibility hold, not a claim that the patch is broken. With `root_squash`, cache blobs that are not world-readable become unreadable to worker root; standard download modes are currently 0644, but that needs to be demonstrated against a real, previously populated cache rather than assumed. Empty-client detection also now fails the launch instead of falling back to a world export — deliberate, but it changes startup behavior for anyone who enabled the feature without setting clients.

**Still needed.** (1) Evidence that a worker can read an existing, previously populated cache through the squashed export. (2) A controlled NFS/startup run in a non-production window showing the fail-closed path and the normal opt-in path.

**What would unblock.** Both pieces of evidence above, plus a CI run of the CPU gates at this head.

shiaho777:
Thanks for the compatibility hold — here is the evidence path, plus the one piece I cannot produce without the cluster.

**Shipped in this PR:**
* `scripts/check-nfs-export-compat.sh` — a live evidence run for the head node. It (1) proves an empty client list is refused fail-closed, (2) shows the live `/etc/exports` carry `root_squash`, (3) proves a **populated** cache stays readable from the worker through the squash (`docker run … test -r /hf/<model>/config.json`), and (4) proves the 0600 HF `token` is **not** readable from it. Output is designed to be pasted into this PR; `bash -n`-gated, never run in CPU CI.
* README + `.env.dspark.example` now document the resulting contract: cache files must be world-readable (standard `hf download` modes are 0644), the squash is what protects the token, and `NFS_OPTS` overrides the DSpark-owned exporter's options for a non-standard cache.

**What I cannot produce:** the live two-node run itself — I have no DGX Spark here. Requested action, whenever a maintenance window allows:

```bash
bash scripts/check-nfs-export-compat.sh      # on the head, worker reachable
```

Please attach its output here if any expectation fails. The script recreates the worker `dspark-hf` volume but never stops an exporter it did not start.

CI is green at this head.

---

Rebased onto current `main` and pushed — the head is now `16a43120`.

Ready for re-review.

plotarmordev:
**Maintainer-owned compatibility hold; current CPU CI requirement addressed.**

At `16a43120306d7944ceb4649b3352ed2974d583ea`, `validate` is green ([run](https://github.com/MiaAI-Lab/DeepSeek-v4-Flash-DSpark-2x-DGX-Spark/actions/runs/34557555516)). The supplied checker is an evidence path, not evidence that the live gate has passed.

**Next-action owner: maintainer (plotarmordev).** We must schedule a separate, explicitly approved populated-cache/root_squash compatibility window, with advance notification before local tests. This is not a read-only check: `scripts/check-nfs-export-compat.sh` calls `nfs_ensure_worker_volume recreate` (line 95), recreating the worker `dspark-hf` volume; it can build/start or reuse the exporter and launches worker probe containers. That mount/volume disruption must be planned around active consumers, not folded into an ordinary startup smoke.

**Unblock condition:** at the exact integrated head proposed for adoption, capture existing populated-cache readability through the squashed export, the empty-client fail-closed path, token protection where applicable (an absent token is not proof of protection), and the normal opt-in NFS/startup path, alongside exact-head CPU/integration checks. The checker alone is not a normal model startup. Hardware remains in use and the maintenance window is unavailable: do not run the checker now, and no author changes, rebase, or hardware tests are owed to clear this maintainer scheduling hold. No merge is being performed here.

plotarmordev (changes requested):
## Hold update

Public head checked: `16a43120306d7944ceb4649b3352ed2974d583ea`.

**Changes needed in `scripts/check-nfs-export-compat.sh`:**
1. Match the export option exactly. The current substring check also accepts `no_root_squash` as `root_squash`, so its PASS is unreliable.
2. Check the selected model revision through the standard `snapshots/<revision>/config.json` layout, not `config.json` at the model-cache root.
3. Remove advice to disable `root_squash` or broadly loosen cache permissions when that incorrect path check fails. Report the actual problem instead.

**What we checked:** These findings were checked in source, including this current public head. We did not run the live helper against an existing exporter or recreate its worker volume. No live security/compatibility pass is claimed.

**Next action — PR author:** Fix the checker and add behavioral cases for exact option handling and a representative populated cache layout. Use disposable data and exports for the tests.

**Before ready:** Run the CPU checks at the fixed commit. Then, in an agreed maintenance window, verify that the worker can read the intended populated model cache while a dummy private token remains unreadable. A missing token is not a passed token-isolation test. I (@plotarmordev) own approval and scheduling of any shared-machine test.

shiaho777:
## Author update — checker fixes landed

Public head: `52903a66ef9cfacc4dac45277a5a4e40310d5ffe`

All three findings addressed in `scripts/check-nfs-export-compat.sh`:

1. **Exact option match** — every `/etc/exports` `(opts)` group is split on commas and must contain a whole-token `root_squash` (`grep -qx`). `no_root_squash` no longer passes; a mixed exports file with any unsquashed group fails and reports the count.
2. **Snapshots layout** — the check resolves the selected revision the same way the serve does (`DSPARK_REVISION` pin → `refs/main` → first `snapshots/*/config.json`, the compose glob order) and probes `snapshots/<revision>/config.json`. A `config.json` at the model root no longer satisfies it.
3. **Actual-problem reports** — three distinct failures now: no `snapshots/*/config.json` on the head (unpopulated cache), the pinned revision absent from the head cache (fix `DSPARK_REVISION` or run prepare), and present-on-head-but-unreadable-through-the-export (other-readability requirement + mount verification). The `NFS_OPTS without root_squash` / `chmod a+r` advice is gone.

Also: `stat -c %a` || `stat -f %Lp` for the token mode so the check is not silently skipped on non-GNU `stat`.

**Behavioral tests** — `scripts/test-nfs-hardening.py` gained a `CheckNfsExportCompat` class that runs the shipped checker end-to-end in a temporary checkout (stub `docker`/`ssh`/`nfs-share.sh`, disposable `/etc/exports`, disposable populated HF cache): exact-option accept and reject, unsquashed-only and mixed exports, all three revision-resolution paths, root-only config, pinned-miss head gap, non-other-readable config, and the 0600/0644 token contract. The ssh stub models the squashed-nobody read as the file's other-read bit.

**Checks run at `52903a6`:**
- `python3 scripts/test-nfs-hardening.py -q` — 17 tests, OK
- `bash -n scripts/check-nfs-export-compat.sh` — OK
- `bash scripts/ci-validate.sh` — **passed** (all gates)

**Environment disclosure / incomplete runs:** local runs were on macOS, so ci-validate needed GNU shims (homebrew bash, `gstat`/`gfind`/`gwc`/`gsed` on PATH, an exported `stat` function for hardcoded-PATH tests, `python3.12` for `anext`). Under that environment every gate passed except one test in `test-build-rsync-guard.py` — `test_apostrophe_path_allowed` — which fails under `/bin/bash` 3.2's double-quoted `${var//pat/repl}` replacement semantics (verified the `'\''` idiom is correct under bash 5.x; unrelated to this PR). Earlier runs also hit pre-existing macOS issues now covered by the shims: `anext` on Python 3.9, BSD `stat`, BSD `find -printf`, BSD `wc` padding. The hosted `validate` run on this head is the authoritative gate.

**Not tested:** the live compatibility window (populated cache readable through the squash + a dummy private token staying unreadable) — still maintainer-owned and scheduled per your note. Branch shows CONFLICTING vs `main`; happy to integrate on your word, or leave sequencing to you.

Issue #55, opened 2026-08-14
Tool call truncated by max_tokens reports finish_reason=tool_calls and can return invalid JSON arguments (poisons the conversation with 400s)


plotarmordev:
## Summary

When generation is cut off by `max_tokens` during a tool call, the response reports `finish_reason: "tool_calls"` rather than `"length"`, and can hand the client a tool call whose `arguments` are **invalid JSON**. A client that stores that call and replays it in the next request then gets the whole conversation rejected with HTTP 400.

No fix offered — reporting the evidence.

## Evidence 1 — truncation is reported as `tool_calls`

One prompt that requires a large tool-call argument, varying only `max_tokens`:

| `max_tokens` | `completion_tokens` | `finish_reason` |
|---:|---:|---|
| 300 | **300** | `tool_calls` |
| 800 | **800** | `tool_calls` |
| 1500 | **1500** | `tool_calls` |

`completion_tokens` lands exactly on the cap every time, which is the signature of truncation, but the client is told the call completed normally. There is no way to distinguish "the model finished its tool call" from "the model was cut off mid-call".

## Evidence 2 — invalid JSON arguments

In multi-turn agent runs (streaming, tools declared each turn, `max_tokens=3000`) tool calls came back with `finish_reason: "tool_calls"` and unterminated JSON:

- 2,809-char `arguments`, `Unterminated string starting at: line 1 column 109 (char 108)`
- 4,915-char `arguments`, `Unterminated string starting at: line 1 column 110 (char 109)`

Both ended mid-string, e.g. `... \"launcher diag\", \"bus failure ->` with no closing quote or brace.

## Consequence

An agent harness appends the returned call to its transcript verbatim. Every subsequent request then contains invalid JSON, and the server rejects the **entire conversation**:

```
400 {"error":{"message":"Unterminated string starting at: line 1 column 91 (char 90)",
     "type":"BadRequestError","param":null,"code":400}}
```

That is unrecoverable from the client's point of view without editing history — in our runs every following turn failed identically until the bad message was dropped.

## Possible relationship to #52

A harness that hits this 400 will typically retry, and a retry can leave the transcript ending with an assistant message, which is the dead state in #52. That would make this the fuse and #52 the bomb.

[...]

PR #264, opened 2026-09-15
fix(issue55): preserve empty tool-call lists during truncation

### LINKED ISSUE #263: [Bug] Issue55 truncation hotfix assigns tool_calls=None and crashes vLLM serialization
The recipe-side #55 truncation hotfix can replace `tool_calls` with `None` when it filters out every call. In the pinned vLLM, both `DeltaMessage` and `ChatMessage` declare this field as a list and their serializers call `len(data.get("tool_calls", []))`. The result is a `PydanticSerializationError` instead of a valid terminal response with `finish_reason="length"`.

This is a follow-up defect in the mitigation for #55. I searched the existing issues and PRs for `NoneType`, `PydanticSerializationError`, serialization, and truncation and did not find a report of this specific failure. Both assignments are still present on current main (`f3d76450e075d5f89698aed2e865655f0b7eeb1c`).

## Environment

- 2× DGX Spark, TP=2 over RoCE.
- Runtime image: `ghcr.io/anemll/dspark-vllm-gx10`, image digest `sha256:a83948492cf13df455170fb42885f5ef4db54fefe0feff0f841ecbff464ac9d8`.
- vLLM: `0.25.2.dev0+g752a3a504.d20260714`.
- Pydantic: `2.13.4`; pydantic-core: `2.46.4`; Python 3.12.
- Observed during a streaming tool request with `max_tokens=8000`, thinking enabled and `reasoning_effort=max`.
- The failure is in response serialization; the minimal reproduction below does not require generating tokens.

## Steps to reproduce

In the pinned vLLM environment, this reproduces the value assigned by the streaming hotfix after filtering the last invalid tool call:

```python
from vllm.entrypoints.openai.engine.protocol import (
    DeltaMessage, DeltaToolCall, DeltaFunctionCall,
)

message = DeltaMessage(tool_calls=[
    DeltaToolCall(index=0, function=DeltaFunctionCall(arguments='{"unfinished":')),
])
kept = []
message.tool_calls = kept or None  # assignment inserted by the #55 hotfix
print(message.model_dump_json(exclude_unset=True))
```

**Actual:**

```text
PydanticSerializationError: Error serializing to JSON: PydanticSerializationError:
Error calling function `_serialize`: TypeError: object of type 'NoneType' has no len()
```

Changing that assignment to `message.tool_calls = kept` yields `{}` without error. I ran both variants inside the serving container.

**Expected:** the hotfix preserves a list, letting vLLM omit an empty `tool_calls` field and deliver the `length` finish reason so clients can discard the truncated response.

## Server log

The head container recorded this traceback at 2026-09-15 04:14:36 UTC:

```text
Error in chat completion stream generator.
Traceback (most recent call last):
  File ".../vllm/entrypoints/openai/chat_completion/serving.py", line 774, in chat_completion_stream_generator
    data = chunk.model_dump_json(exclude_unset=True)
  File ".../pydantic/main.py", line 542, in model_dump_json
    return self.__pydantic_serializer__.to_json(
pydantic_core._pydantic_core.PydanticSerializationError: Error serializing to JSON: PydanticSerializationError: Error calling function `_serialize`: TypeError: object of type 'NoneType' has no len()
```

## Affected code and proposed fix

`patches/hotfix-dsv4-issue55-tool-truncation.py` inserts both:

```python
delta_message.tool_calls = _kept or None
# and, in the non-streaming branch:
message.tool_calls = [...] or None
```

Preserve the empty lists in both branches. The non-streaming branch has the same contract violation, although the incident above was streaming.

Existing installations also need an upgrade path: merely changing the generated replacement string leaves the old already-patched bytes on disk, and the new streaming string is a prefix of the old one. The #191 source-identity pins must track the corrected #55 output as well, since #191 runs after #55 when enabled.

I have a patch covering these changes, regression cases for empty/mixed/natural-stop tool lists, and fresh/legacy/same-container-restart compatibility with #191. The append-only limitation described in #55 remains: fragments already sent before truncation cannot be recalled.

shaircast:
## Description and Motivation

When the #55 truncation hotfix filters every tool call, it assigns `None` to a list-valued field. The pinned vLLM `DeltaMessage` / `ChatMessage` serializers then call `len(None)` and raise `PydanticSerializationError`, preventing delivery of the intended `finish_reason="length"` response.

This change preserves empty lists in both streaming and non-streaming paths. It also upgrades the exact previous hotfix output (including mixed old/new installs), instead of skipping the old streaming line because the fixed line is its prefix.

The #191 fixture and source hashes are updated to match the corrected #55 output. Integration tests cover fresh images, previously applied #55, previously combined #55+#191, and repeated same-container startup. Existing #191 source/version/drift checks remain in place.

This changes terminal response handling only; no model, sampling, GPU kernel, context, memory, or concurrency settings change. The append-only limitation remains: already-streamed argument fragments cannot be recalled, so clients still need to discard truncated calls.

## Related Issues

Fixes #263. Follow-up to #55; preserves compatibility with #191.

## Testing

- `bash scripts/ci-validate.sh`: passed on Ubuntu 24.04 (complete CPU recipe gates).
- `python3 scripts/test-issue55-tool-truncation.py -q`: 7 tests passed, including streaming/non-streaming empty, mixed, valid-only and natural-stop cases.
- `python3 scripts/test-issue191-toolcall-failclosed.py -q`: 32 tests passed, including fresh/legacy/restart patch ordering.
- New behavioral regression test against unmodified main: fails in both response paths (`None != []`); passes with this change.
- Reproduced the serializer exception inside the deployed pinned vLLM container; assigning `[]` instead produces `{}` successfully.
- Applied the patch to a copy of the actual deployed serving source and verified the corrected source hash.
- Post-fix two-node deployment: both containers healthy, both loaded serving sources match corrected SHA-256 `4024243c259058c09f4d9340d42f410e68e160b96fcd5c09b1a68467b04f4f3a`; startup warmup passed 47/47 requests.
- Live API: 4/4 checks passed (streaming/non-streaming × length-truncated/natural completion). Forced `save_note` calls at `max_tokens=128` ended with `finish_reason="length"` without an error; streaming delivered its terminal chunk and `[DONE]`. Natural completions preserved the requested tool name and exact JSON arguments. The synthetic tool was never executed.
- The consuming application health returned to normal after restart.

## Checklist

- [x] Read the README and `docs/ENVS.md`; changes follow the existing patch ordering.
- [x] Concrete title, bug reproduction, and focused regression coverage.
- [x] Complete CPU recipe validation passed.
- [x] Post-fix two-node boot and live response verification complete.
- [x] No new knobs, flags, or default changes; this repairs the existing unconditional #55 mitigation.
- [x] No `.env` / Compose fallback changes are needed.


plotarmordev (approved):
Reviewed all five changed files at a1fed70176c871d384ee86b8e58b88ffe79ca6d4. Both truncation branches preserve list-valued tool_calls, and the previous-anchor upgrade handles the streaming prefix trap without disturbing already-upgraded installs. The issue191 stock/patched identities and composed upgrade/restart coverage match the corrected issue55 bytes.

Authorized and inspected hosted validate run 34931414933 attempt 2: completed success on ubuntu-24.04. Its tested merge-ref 68d745f35a7b6be7b7138cd9bb0195a4c15e70bc has the same tree as this exact head; issue55 (7 tests) and issue191 (32 tests, including composed upgrades/restarts) passed. Separately exercised the actual upstream message class definitions from vLLM 752a3a504485790a2e8491cacbb35c137339ad34 with Pydantic 2.13.4 and these injected branches: both remove invalid terminal calls, serialize without a tool_calls field, and retain finish_reason=length.

Scope and focused secret/deployment-content scans passed. No serving, GPU, SSH or container operations; live author evidence is supplementary. The pre-existing docs/PATCHES.md identity prose remains stale and non-blocking; no source edits requested in this review.

